PR #8, opened 2015-06-01
Current state of the ResolvingJsonParser (jsonres.js)

planger:


edgarmueller:
Closing since there hasn't been any activity since April and we are currently also not actively working in the area of resolving references between different instances. Please re-open, if you disagree.

Issue #10, opened 2015-06-11
Provide jsonforms directive to ease integration


edgarmueller:
Instead of configuring the JSONForms via routing and obtaining the schema/UI schema/instance data via HTTP requests we should instead provide a custom directive that lets users supply the data with sources of their choice.

planger:
Fixed with PR #20!

PR #20, opened 2015-06-11
Provide jsonforms directive to ease integration #10

edgarmueller:


planger:
The structure is much clearer now and it is way easier to get the picture of where we are heading with the framework!

Of course, we should fix the Travis CI, but aside from that it looks great to me!

Thanks a lot!

Issue #23, opened 2015-06-16
Scope used by controls should be modeled by using refs 


edgarmueller:
Instead of providing an own format for specifying the resolution path of a control, we should use JSON references (https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03)

PR #27, opened 2015-06-19
Scope used by controls should be modeled by using refs #23

edgarmueller:


sdirix:
I get an error: the json-refs module is missing. The same error occurs in Travis.

edgarmueller:
Yep, seems like I forgot to update bower.json. I've added the entry for json-refs.

Issue #33, opened 2015-06-26
Support label attribute for controls


edgarmueller:
It should already be supported but by the name of 'displayName'.

planger:
Merged with #34

planger:
Sry, wrong issue... this one is still open.

Issue #32, opened 2015-06-26
Support databinding within jsonforms directive


edgarmueller:
The jsonforms directive should be bindable in order to ease in-place editing of entities. Otherwise we would need a mechanism to submit changes back to the model.

planger:
Merged with #34

Issue #14, opened 2015-06-11
Implement JSONAPI compliant data provider


edgarmueller:
The table widget should be capable of paging its content.

edgarmueller:
We should try to be compliant with http://jsonapi.org/

PR #34, opened 2015-06-26
Support databinding within jsonforms directive #32

edgarmueller:
Support label attribute for controls #33


planger:
Looks great and works well (except for the known bugs such as #14)! Specifying the data via scope variables makes a clean impression. I also like the rename from parameter to scope in the ui schema. Thanks!!

PR #65, opened 2015-08-04
Fix enum properties not being rendered as select element

### LINKED ISSUE #64: Enum properties are rendered as regular input instead of select element
Hi, in the latest version, enum properties are not rendered as select elements, but as regular inputs instead. It still works in the live demo, however the code there seems outdated.

This is probably because in JSONSchema the type of an enum is (I think) always `string`, but never `enum`. However in the control template's `ng-switch-when`, the select input is only rendered when the `schemaType` equals `enum`.

I have prepared a pull request which should fix the problem. I will submit it later.

muenchdo:
This should fix #64. Please get back to me if anything is wrong, I haven't made many contributions yet...

Cheers
Dominik


sdirix:
Works like a charm! Thanks for adding a unit test too ;)

+1 for merge!

sdirix:
A short update for you:
We are currently revising the rules for external committers. This is why we did not yet merge your contribution ;) As soon as we are finished (presumably within the next two weeks) we will inform you regarding the next steps.

muenchdo:
Don't worry, it is not urgent ;)

sdirix:
We cherry-picked your unit test in #97 

Thanks for your contribution ;)

PR #67, opened 2015-08-19
Support for generating ad-hoc JSON schemas for instances

planger:


js/services.ts
```diff
@@ -139,12 +143,127 @@ module jsonforms.services {
 
     }
 
+    export class SchemaGenerator {
+
+        public generateDefaultSchema = (instance: Object) : any => {
+            var schema = this.schemaObject(instance);
+            console.log("generated default schema: " + JSON.stringify(schema));
```
edgarmueller: Remove log

js/services.ts
```diff
@@ -42,6 +42,10 @@ module jsonforms.services {
         resolveModelPath(instance:any, path:string): any
     }
 
+    export interface ISchemaGenerator {
```
edgarmueller: Would it make sense to unify the ISchemaGenerator and the IUISchemaGenerator such that having a single IGenerator interface? If more 'default' generators are added in the future it might get cumbersome if we need inject each one of them. What do you think?

js/services.ts
```diff
@@ -139,12 +143,127 @@ module jsonforms.services {
...
+
+        private isArray = (instance: any): boolean => {
+            return Object.prototype.toString.call(instance) === '[object Array]';
+        };
+
+        private isNotNull = (instance: any): boolean => {
+            return (typeof(instance) !== 'undefined') && (instance !== null);
+        };
+
+        private distinct = (array: Array<Object>, key: (item: Object) => any): Array<Object> => {
```
edgarmueller: I'd rename key to something else. It looks more like a transformation function that determines uniqueness. I makes sense when looking at the implementation of distinct, but I'd guess it doesn't matter for any callers.

js/services.ts
```diff
@@ -139,12 +143,127 @@ module jsonforms.services {
...
+
+    }
+
     export class UISchemaGenerator{
         generateDefaultUISchema = (jsonSchema:any):any =>{
             var uiSchemaElements = [];
             this.generateUISchema(jsonSchema, uiSchemaElements, "#", "");
 
-            console.log("generated schema: " + JSON.stringify(uiSchemaElements[0]))
+            console.log("generated ui schema: " + JSON.stringify(uiSchemaElements[0]))
```
edgarmueller: Remove log.

js/services.ts
```diff
@@ -139,12 +143,127 @@ module jsonforms.services {
...
+                if (known.hasOwnProperty(keyValue)) {
+                    return false;
+                } else {
+                    return (known[keyValue] = true);
+                }
+            });
+        };
+
+        protected requiredProperties = (properties: Object): Object => {
+            return properties; // all known properties are required by default
```
edgarmueller: The defaults for required and additionalProperties make sense, but what if I'd like to customize this behavior? Do I need to subclass the generator or can we add additional parameters to the generateDefaultSchema method?

edgarmueller:
Some minor nitpicks, otherwise this looks real good to me :+1:

planger:
Thanks a lot for the review! I incorporated all of your comments, except for joining ISchemaGenerator and IUISchemaGenerator. I think that there will probably be more implementations of IUISchemaGenerator in the future than for ISchemaGenerator and I assume that they don't necessarily should depend on each other; so I left them separated. However, if you still think they should be joined, I'm happy to do so.

Except for this comment, I incorporated all of your changes and added a test case that shows how the `allowAdditionalProperties` and `requiredProperties` can be configured now. In particular, these two can now be configured using functions that are passed as parameters. This makes the code in `SchemaGenerator` a bit more complex, because the functions need to be passed through a lot, but it eases the configurability of this class.

edgarmueller:
I didn't intend to have a single interface that is implemented by both generators, but an additional one that has references to the schema and the UI generator. Anyways, I don't think that's a important requirement. Thanks!

planger:
Thanks for the merge!

Issue #72, opened 2015-08-24
Refine IRenderer interface


edgarmueller:
isApplicable should only take the current UIElement, the according sub schema and the current schema path.

render should take the same arguments, except one more which is the data provider.

PR #73, opened 2015-08-24
Fix for issue #72: Refinded IRenderer interface

edgarmueller:


app/js/LocalController.js
```diff
@@ -79,26 +79,22 @@ angular.module('makeithappen').controller('LocalController', ['$scope', function
...
     };
     $scope.usersUiSchema = {
-        "elements": [
+        "type": "Control",
+        "scope": {
+            "$ref": "#"
+        },
+        "columns": [
             {
-                "type": "Control",
```
sdirix: The elements in the columns do not need "type":"Control" ? Maybe we should support any type of element there?

edgarmueller: No, columns currently don't have any type (just like it was before this PR). What I've actually changed here it to remove the restriction that the top level element has no type attached, but an elements attribute must be present. Such a top-level entry point does not exist anymore, any UI element can be used now and if multiple elements at once should be specified a Layout should be used.

js/renderers/VerticalLayout.ts
```diff
@@ -26,9 +25,19 @@ class VerticalLayout implements jsonforms.services.IRenderer {
         var renderedElements = renderElements(element.elements);
 
         return {
-            "type": "VerticalLayout",
+            "type": "Layout",
             "elements": renderedElements,
-            "size": 99
+            "size": 99,
+            "template":
+                `<fieldset class="col-sm-{{element.size}}">
```
sdirix: Please remove the class from the fieldset. The size is already set on the parent div

js/renderers/HorizontalLayout.ts
```diff
@@ -35,11 +34,23 @@ class HorizontalLayout implements jsonforms.services.IRenderer {
         }
 
         return {
-            "type": "HorizontalLayout",
+            "type": "Layout",
             "elements": renderedElements,
-            "size": maxSize
+            "size": maxSize,
+            "template":
+                `<fieldset class="col-sm-{{element.size}}">
```
sdirix: Please remove the class from fieldset

sdirix:
Looks good ;)

edgarmueller:
Thanks for the feedback, you're good to with merging :)

Issue #74, opened 2015-08-25
Fix namespace


edgarmueller:
jsonforms.services is currently used for everything

Issue #76, opened 2015-08-26
Validation should be based on tv4


edgarmueller:
Currently the validation for single controls is implemented by hand. Additinally it doesn't consider validation contstraints. Hence we should back our validation mechanism on an existing schema validator.

Issue #75, opened 2015-08-26
Create ResultControlFactory


edgarmueller:
Creating controls via native JS objects can get quite clumy, hence we should provide a factory that is responsible for prefilling most properties with sensible default values

PR #77, opened 2015-08-26
Closes issues #74, #75 and #76

edgarmueller:


js/services.ts
```diff
@@ -80,6 +81,49 @@ module jsonforms.services {
         size: number
     }
 
+    export interface IControlResult extends IResult {
```
sdirix: Please rename Result to something meaningful ;)

edgarmueller: Valid point ;) How about IRenderDescription?

sdirix:
Looks good to me ;)

PR #101, opened 2015-10-14
First iteration of reference control

edgarmueller:
This PR features a design for a references, which come in two flavors:

- as a dedicated 'ReferenceControl' which displays a href 
- as a column that displays a link as a cellTemplate

In both cases the reference can be specified in the UI schema via the `href` keyword. This keyword expects a value as a object which may have the following properties:

- `url` (mandatory): specifies the base view URL the reference points
- `scope` (optional): If a reference features a unique segment that differs from the currently bound property, the `scope` property on a `href` object allows to bind against another property: e.g. suppose we have a User Table containing different users and users are uniquely identified by integer IDs. Further, we want a column that displays a single user's name and that takes us to the respective user, once clicked. In this case, we would need to specify a `scope` property pointing to the `id` property of a user, since we can not use the name of a user to uniquely identify him or her.
- `label`: If the shown text for a reference should differ from the one bound by the control's scope, we can specify an other text via the `label` property.

This PR also extends the demo to display three additional kinds of entities. The demo contains quite a lot of boilerplate and is subject to later changes.


js/renderers/ReferenceControl.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    priority: number = 2;
+
+    constructor(private pathResolver: JSONForms.IPathResolver) {}
+
+    render(element:IUISchemaElement, schema:SchemaElement, schemaPath:string, dataProvider:JSONForms.IDataProvider) {
+        var control = new JSONForms.ControlRenderDescription(dataProvider.data, schemaPath, element.label);
+        var normalizedPath = this.pathResolver.toInstancePath(schemaPath);
+        var linkText = element['label'] ? element['label'] : "Go to " + control.label;
+        control['template'] = `<a href="${element['href']}/${dataProvider.data[normalizedPath]}">${linkText}</a>`;
```
sdirix: This should probably read ```control['template'] = `<a href="#${element['href']['url']}/${dataProvider.data[normalizedPath]}">${linkText}</a>`;```

app/js/PlaceholderController.js
```diff
@@ -0,0 +1,296 @@
...
+            {
+                "type": "Control",
+                "scope": { "$ref": "#/properties/body" }
+            },
+            {
+                "type": "ReferenceControl",
+                "scope": { "$ref": "#/properties/postId" },
+                "href": {
+                    "url": "/placeholder-users",
+                    "title": "Go to post"
```
sdirix: This should be called ```label``` instead of ```title``` to be consistent with other controls. Also it seems that this property is not read anywhere at the moment.

js/renderers/ReferenceControl.ts
```diff
@@ -0,0 +1,27 @@
...
+class ReferenceControl implements JSONForms.IRenderer {
+
+    priority: number = 2;
+
+    constructor(private pathResolver: JSONForms.IPathResolver) {}
+
+    render(element:IUISchemaElement, schema:SchemaElement, schemaPath:string, dataProvider:JSONForms.IDataProvider) {
+        var control = new JSONForms.ControlRenderDescription(dataProvider.data, schemaPath, element.label);
+        var normalizedPath = this.pathResolver.toInstancePath(schemaPath);
+        var linkText = element['label'] ? element['label'] : "Go to " + control.label;
```
sdirix: In regards to my other note this could be changed to ```var linkText = element['href']['label'] ? element['href']['label'] : "Go to " + control.label;```

sdirix: Additionally you could add non-linked text with the ```element['label']``` before the ```<a></a>```

edgarmueller: Good idea!

app/js/PlaceholderController.js
```diff
@@ -0,0 +1,299 @@
...
+                "type": "Control",
+                "scope": { "$ref": "#/properties/body" }
+            },
+            {
+                "type": "ReferenceControl",
+                "scope": { "$ref": "#/properties/postId" },
+                "label": "Navigate to ",
+                "href": {
+                    "url": "/placeholder-users",
+                    "label": "Post"
```
sdirix: Shouldn't this be "User" or "Poster" ?

app/js/PlaceholderController.js
```diff
@@ -0,0 +1,299 @@
...
+                }).$promise;
+            }
+        },
+        fetchPage: function(page, size) {
+            var start = (page - 1) * this.pageSize;
+            var end = start + this.pageSize;
+            return Comments.query({_start: start, _end: end}, function(response) {
+                $scope.CommentDataProvider.data = response;
+            }).$promise;
+        }
```
sdirix: ```setPageSize``` is missing, leading to errors

sdirix:
The main functionality is there and works fine! Great work!

There are however some problems left:
 * The controls in the tables showing total number of pages, currentPage and pageSize do not work properly. Mostly because the data providers are ignored when setting their values: https://github.com/edgarmueller/jsonforms/blob/cfc0d0bdf508a90ec72547669a3cb946f61ca269/js/renderers/ArrayControl.ts#L151
 * The functions/fields of data providers should be sanity checked if they exist before calling/using them to avoid exceptions
 * When the data provider fails at least an empty form and at best an error message should/could be displayed. Currently if for example ```jsonplaceholder``` is not executed only an error is thrown in console.

I am fine with merging this PR if these problems are fixed or created as new issues afterwards.

edgarmueller:
I created follow-up issues #113, #114 and #115. That said, the `IDataProvider` interface is still far from completion, since its lacks features like proper filter etc.

sdirix:
Awesome, thanks for the testcase and the fix! Will merge now ;)

There is one more issue: When for example "/placeholder-posts/29" is navigated in the example app, the validator on "Id" shows an alert: ```Invalid type: number (expected string)```. I will open it as a new issue.

PR #103, opened 2015-10-16
Remove recelement from dom

eneufeld:
Here is a possible suggestion on how to remove the recelement so that it also not shows up in the dom.
By doing so we can provide renderers which rely on a specific hierarchy of elements in order to apply correct css e.g. a renderer for bootstrap tabs


js/services.ts
```diff
@@ -72,7 +72,7 @@ module JSONForms {
 
     export interface IRenderer {
         /**
-         * When the RenderService�s render method is called it gets passed the UI Schema element (e.g. a Control)
+         * When the RenderService�s render method is called it gets passed the UI Schema element (e.g. a Control)
```
edgarmueller: Seems like an encoding issue on your side.

PR #110, opened 2015-10-26
Adjust JSONForms to recommended angular app structure

sdirix:


components/renderers/controls/autocomplete/autocomplete-renderer.ts
```diff
@@ -0,0 +1,23 @@
+///<reference path="../../../references.ts"/>
+
+class AutoCompleteControl implements JSONForms.IRenderer {
```
edgarmueller: Should be called AutoCompleteRenderer in order to conform with the new naming schema.

edgarmueller:
The new structure looks great, I have only found one nitpick

edgarmueller:
Thank you!

PR #111, opened 2015-10-26
Validation support for ArrayControl

edgarmueller:
The ArrayControl (Table) is now capable of supporting validation. This PR also features the necessary refactorings in order to do so.


sdirix:
Looks good.

The new array control is nice but still needs some care: for example the columns can be removed to be seen never again (at least until reload). The ServiceObject is interesing, we should discuss if we want to go further that route or use a different approach. At the moment it is not much different from the DataService before, so I will accept it.

PR #112, opened 2015-10-29
Fix nested path bindings

edgarmueller:
Also fixed potential undefined error in isApplicable of controls

### DIFF
--- a/js/directives.ts
+++ b/js/directives.ts
@@ -155,24 +155,34 @@ jsonFormsDirectives.directive('jsonforms', function ():ng.IDirective {
     }
 })
 .directive('dynamicWidget', ['$compile', '$templateRequest', function ($compile: ng.ICompileService, $templateRequest: ng.ITemplateRequestService) {
-    var replaceJSONFormsAttributeInTemplate = (template: string): string => {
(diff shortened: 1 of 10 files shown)


js/directives.ts
```diff
@@ -155,24 +155,34 @@ jsonFormsDirectives.directive('jsonforms', function ():ng.IDirective {
     }
 })
 .directive('dynamicWidget', ['$compile', '$templateRequest', function ($compile: ng.ICompileService, $templateRequest: ng.ITemplateRequestService) {
-    var replaceJSONFormsAttributeInTemplate = (template: string): string => {
-        return template
-            .replace("data-jsonforms-model",      "ng-model='element.instance[element.path]'")
-            .replace("data-jsonforms-validation", `ng-change='element.validate()'`);
+    var replaceJSONFormsAttributeInTemplate = (template, fragments) => {
```
sdirix: Is there a reason to remove the return type?

edgarmueller: Nope, I guess not, will fix.

sdirix:
Looks good to me. Thanks for adding a testcase!

PR #118, opened 2015-11-03
Added watch  to re-render forms on change of the UI schema

edgarmueller:
Also added devDependency for ace-ui and updated embed-index.html and the editor template and controller


sdirix:
Great work! It is really fun to change the view model on the fly!

I have two issues:
- After re-rendering the validator throws wrong validation errors. To reproduce: Enter a date in the form and then re-render
- The console is spammed with exceptions while typing in the editor. This is fine at the moment but at some time in the future we should probably validate the ui-schema before re-rendering.

PR #126, opened 2015-11-04
Fixed array renderer

edgarmueller:
Fixed column definitions creation method in array renderer and typo in placeholder-posts.html


sdirix:
Thanks for the quick fix! 

Sadly the nextPage/previousPage buttons still do not work :/ If you have the time could you maybe look at them? Thanks!

Issue #71, opened 2015-08-21
Support rules


edgarmueller:
For showing/hiding element in the UI. Should be specified within UI schema.

edgarmueller:
Initial support for rules is available, for all follow-up issues additional issues should be created (like for instance #132 )

PR #130, opened 2015-11-05
RuleService with control support

eneufeld:
This is a first implementation of a possible rule support as requested in #71 .
Rules cannot be applied to layouts yet.


eneufeld:
will create a clean new pr

PR #134, opened 2015-11-06
Added support for show/hide rules

eneufeld:
This is a first implementation of a possible rule support as requested in #71  .
Disable and enable rules are not supported yet.

### DIFF
--- a/components/renderers/controls/number/number-renderer.ts
+++ b/components/renderers/controls/number/number-renderer.ts
@@ -5,7 +5,7 @@ class NumberRenderer implements JSONForms.IRenderer {
     priority = 2;
 
     render(element: IUISchemaElement, subSchema: SchemaElement, schemaPath: string, services: JSONForms.Services) {
-        var control = JSONForms.RenderDescriptionFactory.createControlDescription(schemaPath, services, element.label);
+        var control = JSONForms.RenderDescriptionFactory.createControlDescription(schemaPath, services, element.label, element.rule);
         control['template'] = `<control><input type="number" step="0.01" id="${schemaPath}" class="form-control jsf-control jsf-control-number" data-jsonforms-validation data-jsonforms-model/></control>`;
         return control;
     }
--- a/components/renderers/jsonforms-renderers.ts
+++ b/components/renderers/jsonforms-renderers.ts
@@ -46,8 +46,8 @@ module JSONForms {
     }
 
     export class RenderDescriptionFactory implements IRendererDescriptionFactory {
-        static createControlDescription(schemaPath:string, services:JSONForms.Services, label?:string):IRenderDescription {
-            return new ControlRenderDescription(schemaPath, services, label);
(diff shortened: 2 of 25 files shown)


components/renderers/controls/number/number-renderer.ts
```diff
@@ -5,7 +5,7 @@ class NumberRenderer implements JSONForms.IRenderer {
     priority = 2;
 
     render(element: IUISchemaElement, subSchema: SchemaElement, schemaPath: string, services: JSONForms.Services) {
-        var control = JSONForms.RenderDescriptionFactory.createControlDescription(schemaPath, services, element.label);
+        var control = JSONForms.RenderDescriptionFactory.createControlDescription(schemaPath, services, element.label, element.rule);
```
edgarmueller: We should think about passing `element`

components/renderers/jsonforms-renderers.ts
```diff
@@ -125,6 +142,7 @@ module JSONForms {
         private setupModelChangedCallback():void {
             this.scope.$on('modelChanged', () => {
                 this.validate();
+                this.ruleService.revaluateRules(this, this.schemaPath);
```
edgarmueller: typo: should be `reevaluateRules`

components/renderers/jsonforms-renderers.ts
```diff
@@ -72,17 +85,21 @@ module JSONForms {
...
             this.schema = services.get<JSONForms.ISchemaProvider>(ServiceId.SchemaProvider).getSchema();
             this.validationService = services.get<JSONForms.IValidationService>(ServiceId.Validation);
             this.pathResolver = services.get<JSONForms.IPathResolverService>(ServiceId.PathResolver).getResolver();
+            this.ruleService = services.get<JSONForms.IRuleService>(ServiceId.RuleService);
             this.scope = services.get<JSONForms.IScopeProvider>(ServiceId.ScopeProvider).getScope();
 
             this.path = PathUtil.normalize(schemaPath);
             this.label = this.createLabel(schemaPath, label);
+            this.ruleService.addRuleTrack(this);
+            this.ruleService.revaluateRules(this, this.schemaPath);
```
edgarmueller: Does this handle the init case? If so, I think we can remove this line, since a broadcast `modelChanged` event is triggered after the first `render` call.

typings/schemas/uischema.d.ts
```diff
@@ -1,11 +1,30 @@
...
+    HIDE,SHOW,ENABLE,DISABLE
+}
+interface ICondition {
+    type:string; //nice to have
+}
+interface ILeafCondition extends ICondition {
+    scope: {
+        $ref: string;
+    }
+    value: any;
```
edgarmueller: This is the expected value for the condition to evaluate to true, right? Can we rename this to `expectedValue` then?

edgarmueller:
This looks great, I only have minor nitpicks. Can you address them, so I can merge this?

edgarmueller:
Thanks!

Issue #15, opened 2015-06-11
Configure and implement filter for table


edgarmueller:
The table widget should be capable filtering its content by queries.

PR #140, opened 2015-11-12
Fix for issue #15: Configure and implement filter for table 

edgarmueller:


edgarmueller:
Fixes filtering by directly wireing through ui-grid options to ui-grid. 
- if `useExternalFiltering` is enabled, the `DataProvider`'s filter method will be called
- if 'enableFiltering' is set, default internal filtering is done

Both options can be set to `false` to completly disable filtering.

components/renderers/controls/array/array-renderer.ts
```diff
@@ -2,80 +2,105 @@
...
-        defaultGridOptions.enableHorizontalScrollbar = 0;
-        defaultGridOptions.enableVerticalScrollbar = 0;
         var dataProvider = services.get<JSONForms.IDataProvider>(JSONForms.ServiceId.DataProvider);
         var validationService = services.get<JSONForms.IValidationService>(JSONForms.ServiceId.Validation);
 
-        if (dataProvider instanceof JSONForms.DefaultInternalDataProvider) {
-            defaultGridOptions.totalItems = dataProvider.getTotalItems();
-        } else {
+        let externalPaginationEnabled = ArrayRenderer.isExternalPagingAvailable(dataProvider, existingOptions);
+        let externalFilteringEnabled  = ArrayRenderer.isExternalFilterAvailable(dataProvider, existingOptions);
```
eneufeld: can the existingOptions parameter be undefined/null here? if so both static method should handle this correctly

components/renderers/controls/array/array-renderer.ts
```diff
@@ -2,80 +2,105 @@
...
 
-    private maxSize = 99;
-    priority = 2;
+    private static convertColumnValue(colDef: uiGrid.IColumnDef, value: string): any {
+        if (colDef.type) {
+            if (colDef.type == "number" || colDef.type == "integer") {
+               return Number(value);
+            } else if (colDef.type == "boolean") {
+                return Boolean(value);
+            }
```
eneufeld: what about date? http://www.w3schools.com/jsref/jsref_obj_date.asp

components/utils/dataproviders.d.ts
```diff
@@ -2,12 +2,65 @@
...
+
+        /**
+         * Returns the instance.
+         *
+         * @return {any} the instance
+         */
         getData(): any
-        getPageSize(): number
-        getPage(): number
+
```
eneufeld: this are table specific methods, couldn't we provide them via an optional object instead of making all those method optional?

examples/js/PlaceholderController.js
```diff
@@ -202,7 +208,7 @@ angular.module('makeithappen').controller('PlaceholderController', ['$scope', '$
     // Resources --
     //
 
-    var Posts = $resource('http://localhost:3000/posts/:id');
+    var Posts = $resource('http://localhost:3000/posts/:id?');
```
eneufeld: why add a ? here only

examples/js/PlaceholderController.js
```diff
@@ -235,7 +241,7 @@ angular.module('makeithappen').controller('PlaceholderController', ['$scope', '$
                 var start = (page - 1) * this.pageSize;
                 var end = start + this.pageSize;
                 var that = this;
-                return Comments.query({_start: start, _end: end}, function (response) {
+                return toQuery.query({_start: start, _end: end}, function (response) {
```
eneufeld: was there a bug here?

examples/js/PlaceholderController.js
```diff
@@ -254,6 +260,14 @@ angular.module('makeithappen').controller('PlaceholderController', ['$scope', '$
                         console.log("error occurred: " + JSON.stringify(error));
                     }).$promise;
                 }
+            },
+            filter: function(terms) {
```
eneufeld: wouldn't a fat arrow help, so you can remove that=this

eneufeld:
Thank you, looks good, just some questions

components/renderers/controls/array/array-renderer.ts
edgarmueller: It shouldn't be because `defaultGridOptions` is called with `element.options || {}`

components/renderers/controls/array/array-renderer.ts
edgarmueller: I think we don't have to handle `Date`s here, because JSON schema doesn't have a `Date` type, but a string with `format` being set to `date-time`.

examples/js/PlaceholderController.js
edgarmueller: Yes, actually It can be removed again. It's a leftover I have overseen.

components/utils/dataproviders.d.ts
edgarmueller: I think we could, of course, but I think paging/filtering should stay optional nevertheless. Anyways, I think we should discuss in a separate issue, what do you think?

examples/js/PlaceholderController.js
edgarmueller: Yes, the `createDataProvider` function should be generic and not specific to any resource.

examples/js/PlaceholderController.js
edgarmueller: Not in this case, since we are consuming JSON Forms via JS here.

edgarmueller:
Thanks for the review, I will push an update later today.

components/utils/dataproviders.d.ts
eneufeld: I totally agree

eneufeld:
Thanks for the clarification

eneufeld:
Looks good, thanks for the great work.
Just add a test for the table so the coverage check doesn't fail.

Issue #92, opened 2015-09-24
Apply styleguide


edgarmueller:
See https://github.com/johnpapa/angular-styleguide

PR #142, opened 2015-12-07
Fix for #92: Styleguide

edgarmueller:
- directives names are prefixed with jsonforms
 - usage of controllerAs syntax
 - restructured examples
 - also removed async variants of jsonforms attributes
 - some minor additional fixes


components/form/form-directive.ts
```diff
@@ -49,7 +50,7 @@ class FormController {
 
             if (uiSchemaMaybe === undefined || uiSchemaMaybe === null || uiSchemaMaybe === "") {
                 // resolve JSON schema, then generate ui Schema
-                JsonRefs.resolveRefs(schema, {}, (err, resolvedSchema, meta) => {
```
eneufeld: what is/was the meta variable?

components/form/form-directive.ts
```diff
@@ -68,108 +69,106 @@ class FormController {
...
+            if (FormController.isDataProvider(this.scope.data)) {
+                dataProvider = this.scope.data;
             } else {
                 dataProvider = new JSONForms.DefaultDataProvider(this.$q, data);
             }
+
             services.add(dataProvider);
 
-            this.$scope['elements'] = [this.RenderService.render(this.$scope, uiSchema, services)];
+console.log("the ui schema is " + JSON.stringify(uiSchema))
```
eneufeld: did you forget to remove this?

components/form/form-directive.ts
```diff
@@ -68,108 +69,106 @@ class FormController {
...
+    scope = {
+        schema: '=',
+        uiSchema: '=',
+        data: '=',
+    };
+    link = (scope, el, attrs, ctrl) => {
+        ctrl.element = el;
+        scope.$watchGroup(['data', 'uiSchema'], (newValue) => {
+            if (angular.isDefined(newValue)){
+                console.log("data changed");
```
eneufeld: is this the normal way to log?

components/form/form-directive.ts
```diff
@@ -68,108 +69,106 @@ class FormController {
...
 
-    asyncSchema: () => any;
-    asyncUiSchema: () => any;
-    asyncDataProvider: JSONForms.IDataProvider;
+class JsonFormsDirective implements ng.IDirective {
+    restrict = "E";
+    replace = true;
+    templateUrl = 'components/form/form.html';
+    controller = FormController;
+    controllerAs = 'vm';
```
eneufeld: what does this mean?

components/renderers/layouts/masterdetail/masterdetail-directives.ts
```diff
@@ -0,0 +1,53 @@
...
+                            <jsonforms-masterdetail-member element="element" member="value.items"></jsonforms-masterdetail-member>
+                        </accordion-group>
+                    </accordion>
+                </accordion-group>
+            </accordion>
+        </div>
+        `
+}
+
+// TODO (?) http://stackoverflow.com/questions/23535994/implementing-angularjs-directives-as-classes-in-typescript
```
eneufeld: is todo still valid?

components/renderers/renderers-directive.ts
```diff
@@ -21,24 +42,13 @@ angular.module('jsonforms.renderers').directive('widget', function ():ng.IDirect
...
-                var updatedTemplate = replaceJSONFormsAttributeInTemplate(scope.element.template, fragments);
-                var template = $compile(updatedTemplate)(scope);
-                element.replaceWith(template);
-            }
-        }
     }
-}]);
+}
+
+// TODO (?) http://stackoverflow.com/questions/23535994/implementing-angularjs-directives-as-classes-in-typescript
```
eneufeld: still valid?

examples/app/custom/custom.controller.js
```diff
@@ -1,6 +1,8 @@
-angular.module('makeithappen').controller('MyController', ['$scope', function($scope) {
+'use strict';
 
-    $scope.schema = {
+angular.module('makeithappen').controller('MyController', function() {
```
eneufeld: make controller name consistent with name in js file?

eneufeld:
thanks, looks good, just some minor questions

components/form/form-directive.ts
edgarmueller: What exactly do you mean :) There are actually two things happerning here:
1. I've removed the async-* attribute variants from the `jsonforms` directive since I think this is only confusing users. Instead one can either specify the name of respective Provider or the object literal and we will dispatch accordingly.
2. Usage of `controllerAs` syntax, which binds the controller to the current scope, and make it available via `vm` in the `form.html` template.

components/form/form-directive.ts
edgarmueller: The meta resolution data varaible of json-refs. It was unused and caused a warning, hence I decided to remove it.

components/form/form-directive.ts
edgarmueller: Yes, I did :)

components/form/form-directive.ts
edgarmueller: Same as above.

components/renderers/layouts/masterdetail/masterdetail-directives.ts
edgarmueller: Yes, because we could also use factories here, which has the advantage of only maintaining a single list of parameter. But this isnt' very high prio currently.

components/renderers/renderers-directive.ts
edgarmueller: Yes, same as above.

examples/app/custom/custom.controller.js
edgarmueller: Will do.

edgarmueller:
Thanks for your comments, I'll fix the nitpicks you mentioned.

components/form/form-directive.ts
eneufeld: thanks, i meant the controllerAs thing.

edgarmueller:
@eneufeld I've included your suggestions

PR #143, opened 2015-12-07
Fixes #139

### LINKED ISSUE #139: RenderDescriptionFactory#createControlDescription should be passed whole element
Currently, we only pass the `label`, but it's likely we need more parameters soon, hence we wouldn't need to constantly update all renderers.

edgarmueller:
`RenderDescriptionFactory#createControlDescription` is passed the whole UI schema element.


sdirix:
Looks good ;)

PR #154, opened 2015-12-14
Added polymer example

edgarmueller:
- Added polymer example
- Fixed validation of enum control


bower.json
```diff
@@ -29,6 +29,8 @@
   "devDependencies": {
     "angular-mocks": "1.3.16",
     "angular-resource": "1.3.16",
-    "angular-ui-ace": "0.2.3"
+    "angular-ui-ace": "0.2.3",
+    "polymer": "Polymer/polymer#^1.2.0",
+    "paper-input": "PolymerElements/paper-input#^1.0.0"
```
eneufeld: is this the name of the version?

examples/app/polymer/polymer.controller.js
```diff
@@ -0,0 +1,68 @@
+angular.module('makeithappen').controller('PolymerController', function() {
+    var vm = this;
+    vm.schema = {
```
eneufeld: can't we load the schema and example data from a data-schema.json file just like we do on the webpage? we would reduce the duplication by a lot

examples/app/polymer/polymer.html
```diff
@@ -0,0 +1,12 @@
+<div class="panel panel-default">
+    <div class="panel-heading">
+        <h3>Bound data</h3>
+    </div>
+    <div class="panel-body">
+        <pre>{{vm.formattedData()}}</pre>
+    </div>
+</div>
+
+<jsonforms data="vm.data" ui-schema="vm.uiSchema" schema="vm.schema"></jsonforms>
```
eneufeld: why not call the controller polymer instead of vm or is vm a common name?

examples/app/polymer/polymer.jsf.js
```diff
@@ -0,0 +1,39 @@
+var app = angular.module('makeithappen');
+app.run(['RenderService', 'PathResolver', '$rootScope', function(RenderService, PathResolver, $rootScope) {
+
+    function PolymerControl() {
+
+        var instance;
+        var path;
+
+        // set-up event binding
+        document.addEventListener('WebComponentsReady', function() {
```
eneufeld: i don't like this part at all ... but if the control doesn't support databinding there is no other way :-(

examples/app/polymer/polymer.jsf.js
```diff
@@ -0,0 +1,39 @@
...
+        return {
+            priority: 100,
+            render: function(element, schema, schemaPath, services) {
+                var control = JSONForms.RenderDescriptionFactory.createControlDescription(schemaPath, services, element);
+                var data = services.get(JSONForms.ServiceId.DataProvider).getData();
+                path = PathResolver.toInstancePath(schemaPath);
+                instance = data;
+                control['template'] =
+                 '<div class="panel panel-default">' +
+                    '<paper-input type="text" value="{{element.instance[\'' + path + '\']}}" data-jsonforms-model label="Last name goes here"></paper-input>' +
```
eneufeld: the paper-input has it own label? or why don't we use the jsonform-control directive?

eneufeld:
Looks good, thank you. Just some minor points

bower.json
edgarmueller: Yes, it is, see [here](https://www.polymer-project.org/1.0/docs/start/getting-the-code.html) for comparison.

examples/app/polymer/polymer.controller.js
edgarmueller: I don't know whether the schemas are identical, but we probably should unify all/most of them. I'd prefer to track that via a separate issue however.

examples/app/polymer/polymer.html
edgarmueller: Yes, the latter, vm is commonly used for ViewModel, but of course you are free to use whatever you like.

examples/app/polymer/polymer.jsf.js
edgarmueller: Me neither, but I couldn't think of another way. I think this is where Angular 2 would probably be more useful since we could bind against value and the respective event.

examples/app/polymer/polymer.jsf.js
edgarmueller: Yes it does, it looks different from the default one we provide (it's hovering over the actual input field, if no text has been entere).

edgarmueller:
@eneufeld See comments. Please note I also mistakenly pushed an additional change which has nothing in common with the Polymer example.

PR #163, opened 2016-01-21
updated angular version

eneufeld:
There is a new version of angular, we should make our version more flexible, done this by adding a ~


eneufeld:
@edgarmueller please review and create a new release if possible

edgarmueller:
Looks good, thanks!

Issue #162, opened 2016-01-20
Read only controls


hecforga:
I am working in the JSON Forms Editor. In the properties panel, which are rendered using jsonforms, there are some that the user should not modify (read only). This is the case of the "Type" property when there is only one type (for example Control or Categorization). Or the "Scope" property of Controls. Now there is no way of doing this using jsonforms.

hecforga:
I think this is already implemented

PR #165, opened 2016-01-27
Fix for issue #162: Implemented readOnly flag for controls 

edgarmueller:


sdirix:
Great work! This will be very useful. I also like all the new testcases!

PR #174, opened 2016-02-18
Initial implementation of Array control

hecforga:


components/app.ts
```diff
@@ -30,4 +30,6 @@ angular.module('jsonforms', [
     'jsonforms.renderers.layouts.categories.category',
     'jsonforms.renderers.extras.label',
     'jsonforms.renderers.layouts.masterdetail',
+    'jsonforms.renderers.widgets.widgeta',
+    'jsonforms.renderers.widgets.widgetb'
```
edgarmueller: We should probably give these widgets more meaningful names.

components/renderers/widgets/widgeta/widgeta-renderer.ts
```diff
@@ -0,0 +1,57 @@
+///<reference path="../../../references.ts"/>
+
+class WidgetARenderer implements JSONForms.IRenderer {
```
edgarmueller: What the difference between `WidgetARenderer` and `WidgetBRenderer`? Do we need this one at all?

components/renderers/widgets/widgetb/widgetb-renderer.ts
```diff
@@ -0,0 +1,64 @@
...
+                        "type": "Control",
+                        "label": JSONForms.PathUtil.beautify(prop),
+                        "scope": {
+                            "$ref": schemaPath + "/items/" + i + "/properties/" + prop
+                        }
+                    });
+                }
+            }
+            auxUISchemaElements.push(auxUISchemaElement);
+            i++;
```
edgarmueller: Replace `i` with `d` (which already tracks the index)

components/renderers/widgets/widgetb/widgetb-renderer.ts
```diff
@@ -0,0 +1,64 @@
...
+    constructor(private renderService: JSONForms.IRenderService, private pathResolver: JSONForms.IPathResolver) { }
+
+    render(element: IControlObject, subSchema: SchemaArray, schemaPath: string, services: JSONForms.Services): JSONForms.IRenderDescription {
+        var data = services.get<JSONForms.IDataProvider>(JSONForms.ServiceId.DataProvider).getData();
+
+        if (!Array.isArray(data)) {
+            data = this.pathResolver.resolveInstance(data, schemaPath);
+        }
+
+        var dataString = JSON.stringify(data);
```
edgarmueller: This isn't used anywhere, is it?

components/app.ts
hecforga: Any idea? What do you think of simplearraywidget and grouparraywidget?

edgarmueller:
Nice work, let's fix the issues together and I think we're good to go

components/renderers/widgets/widgeta/widgeta-renderer.ts
hecforga: Jonas told me to implement one basic widget (widgeta) and other showing controls instead of lines of text (widgetb). I'm not sure if widgeta is what Jonas wanted, so we can not include it by the moment.

components/renderers/widgets/widgetb/widgetb-renderer.ts
hecforga: WidgetBRenderer is the one showing controls in different groups. WidgetARenderer has lower priority so it's never applied. I think is better not including widgeta by the moment.

Issue #93, opened 2015-09-28
Fix missing glyphicons warnings


edgarmueller:
GET 
http://localhost:8080/fonts/glyphicons-halflings-regular.woff [HTTP/1.1 404 Not Found 5ms]
downloadable font: download failed (font-family: "Glyphicons Halflings" style:normal weight:normal stretch:normal src index:1): status=2147746065 source: http://localhost:8080/fonts/glyphicons-halflings-regular.woff qb-bootstrap.min.css:261:12
GET 
http://localhost:8080/fonts/glyphicons-halflings-regular.ttf [HTTP/1.1 404 Not Found 5ms]
downloadable font: download failed (font-family: "Glyphicons Halflings" style:normal weight:normal stretch:normal src index:2): status=2147746065 source: http://localhost:8080/fonts/glyphicons-halflings-regular.ttf qb-bootstrap.min.css:261:12
GET 
http://localhost:8080/bower_components/jsonforms/dist/fonts/glyphicons-halflings-regular.woff2 [HTTP/1.1 404 Not Found 5ms]
downloadable font: download failed (font-family: "Glyphicons Halflings" style:normal weight:normal stretch:normal src index:1): status=2147746065 source: http://localhost:8080/bower_components/jsonforms/dist/fonts/glyphicons-halflings-regular.woff2 jsonforms.css:257:12
GET 
http://localhost:8080/bower_components/jsonforms/dist/fonts/glyphicons-halflings-regular.woff [HTTP/1.1 404 Not Found 6ms]
downloadable font: download failed (font-family: "Glyphicons Halflings" style:normal weight:normal stretch:normal src index:2): status=2147746065 source: http://localhost:8080/bower_components/jsonforms/dist/fonts/glyphicons-halflings-regular.woff jsonforms.css:257:12
GET 
http://localhost:8080/bower_components/jsonforms/dist/fonts/glyphicons-halflings-regular.ttf [HTTP/1.1 404 Not Found 4ms]
downloadable font: download failed (font-family: "Glyphicons Halflings" style:normal weight:normal stretch:normal src index:3): status=2147746065 source: http://localhost:8080/bower_components/jsonforms/dist/fonts/glyphicons-halflings-regular.ttf

edgarmueller:
Seems like a bootstrap problem. In the our resulting CSS file the paths to glyphicons are wrong.

koegel:
[...]

PR #185, opened 2016-02-29
Fix missing glyphicons warnings #93

sdirix:
Adds the missing Bootstrap fonts to the distribution and the app
example (#93)


edgarmueller:
Thanks!

Issue #190, opened 2016-03-03
make array submit renderer default


eneufeld:
The array submit renderer should be the default, not the other way round

PR #188, opened 2016-02-29
UISchemaGenerator improvements

sdirix:
* Adjust UISchemaGenerator for Array controls (#182) 
 * Also generate VerticalLayout if JSONSchema consists of a single primitive type

PR #192, opened 2016-03-04
Beef up examples

edgarmueller:


edgarmueller:
This also closes #190

components/renderers/controls/array/array-renderer.ts
```diff
@@ -122,7 +123,7 @@ class ArrayRenderer implements JSONForms.IRenderer {
                 containerDescription['submitControls'] = this.createControlsForSubmit(items, schemaPath, submitElement, services);
                 if (data == undefined) {
                     containerDescription["instance"][containerDescription['path']] = [];
-                    data = [];
+                    data = containerDescription["instance"][containerDescription['path']];
```
sdirix: I liked the previous line better

edgarmueller: Yeah, but it doesn't work. `data` need to point to the actual reference and not to a anonymous array.

sdirix:
First things first: This looks GREAT!

Minor nitpicks:
 * ```#editor```: 

image
 * ```#custom```, ```#defaultui```, ```#defaultschema```:

image
 * ```#placeholder```: 

image

Good work! I like the look of the new examples!

edgarmueller:
Incoming, btw: does the editor example work for you?

sdirix:
No it does not. Also there are validation problems in the ```#local``` example. When the form is initially rendered the ```Age``` is not validated. Not before I change an unrelated property the ```Required``` finally shows up.
image
image
I assume these problems do not originate from this change? That is why I did not mention them.

components/renderers/controls/array/array-renderer.ts
sdirix: Ah good to know! :+1:

edgarmueller:
This is intended, although I admit, that this is might actually be a bad example. The age field and "that unrelated property" both are contained in a sub-object, which does not exists initially, hence validation does not complain. Only when you change the unrelated property the object will be created and the validation is triggered. 
Could you create another issue for fixing the editor example and continue with this PR?
Thank you!

sdirix:
Great work ;) :+1:

edgarmueller:
Fixed via #188

PR #195, opened 2016-03-21
Fix deploy script to create a proper annotation for the Git release tag

### LINKED ISSUE #189: Releases are tagged with 'Release version v%= pkg.version %'
All releases starting from version ```0.0.9``` are tagged with ```Release version v%= pkg.version %``` instead of the proper version information.

sdirix:
Fixes #189


edgarmueller:
Thanks!

PR #196, opened 2016-03-22
Introduce new 'show' option for control labels

### LINKED ISSUE #183: Control labels can not be hidden
Since we are trying to be super-smart, we derive labels from the schema path for controls even if one did not specify a label in the UI schema. Question then is: how can we actually hide a label?

sdirix:
Fixes #183

Introduces a new configuration option in ui schemas for labels of
controls. In particular a 'show' option is added which allows to hide labels.

Labels are now configured as blocks with a min-height of 20px (the current line-height) to avoid misalignments between inputs with and without labels. However when no label for any control is displayed there are seamingly meaningless 20px of whitespace present. This problem should probably be solved in a generic way in a future PR since labels are also used to display the required property.

### DIFF
--- a/components/renderers/jsonforms-renderers.spec.ts
+++ b/components/renderers/jsonforms-renderers.spec.ts
@@ -1,0 +1,242 @@
+/// <reference path="../references.ts"/>
+
+describe('Generic renderer', () => {
+
+    // load all necessary modules and templates
+    beforeEach(module('jsonforms.form'));
+    beforeEach(module('jsonforms.renderers.controls.string'));
+
+    beforeEach(module('components/form/form.html'));
+    beforeEach(module('components/renderers/controls/control.html'));
+
+    it("should render the provided label string", inject(($rootScope, $compile) => {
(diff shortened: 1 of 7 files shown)


components/renderers/jsonforms-renderers.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+        };
+        scope.data = { "name": "My Name" };
+        let el = $compile('<jsonforms schema="schema" ui-schema="uiSchema" data="data"/>')(scope);
+        scope.$digest();
+        expect(el.find("label").text()).toEqual("Name");
+    }));
+
+    it("should hide the label when text is defined but show is set to false", inject(($rootScope, $compile) => {
+        let scope = $rootScope.$new();
+        scope.schema = {
```
edgarmueller: The schema seems to be the same in all cases, maybe we could extract it and declare it as a global var within the test

edgarmueller:
This looks good! The one comment I added I'll probably take care of myself. Thanks!

PR #221, opened 2016-04-19
Webpack bundling

edgarmueller:
Removes the need for grunt and bower and instead uses webpack and npm modules to create a bundle for easy client consumption. Code structure also has been refined slightly: framework code is now under src/. Furthermore module syntax has been introduced and obsolete files have been removed. Some minor things still need to be fixed, among them re-enabling protractor tests and bootstrap workflow configuration.


edgarmueller:
@eneufeld Sorry, this is quite a monster..

src/components/renderers/controls/integer/integer-renderer.ts
```diff
@@ -0,0 +1,29 @@
...
+        return control;
+    }
+
+    isApplicable(uiElement: IUISchemaElement, subSchema: SchemaElement, schemaPath: string):boolean {
+        return uiElement.type == 'Control' && subSchema !== undefined && subSchema.type == 'integer';
+    }
+}
+
+export default angular
+    .module('jsonforms.renderers.controls.integer', [])
```
eneufeld: i think a `jsonforms.renderers.controls`dependency is missing

src/components/generators/schema/schema-service.ts
```diff
@@ -0,0 +1,6 @@
+
```
eneufeld: why don't register in jsonforms-schemagenerator just like you do it for renderer and thus remove the schema/uischema folders which would then contain exactly one class?

src/components/generators/uischema/uischema-service.ts
```diff
@@ -0,0 +1,5 @@
+import {UISchemaGenerator} from "./jsonforms-uischemagenerator";
```
eneufeld: why not register in jsonforms-uischemagenerator, also see jsonforms-schemagenerator

src/components/renderers/layouts/masterdetail/masterdetail-renderer.ts
```diff
@@ -38,6 +42,9 @@ class MasterDetailRenderer implements JSONForms.IRenderer {
     }
 }
 
-angular.module('jsonforms.renderers.layouts.masterdetail').run(['RenderService', (RenderService) => {
-    RenderService.register(new MasterDetailRenderer());
-}]);
+export default angular
```
eneufeld: missing dependency to `jsonforms.renderers.layouts` in module registration

eneufeld:
Looks good, thank you for the refactoring. 
I think there are some minor issues left

src/components/renderers/layouts/masterdetail/masterdetail-renderer.ts
edgarmueller: Registering the module happens in `masterdetail-renderer.ts`, so I think this ok.

edgarmueller:
@eneufeld  Thanks for the feedback, I addressed your changes in the lastest commit.

eneufeld:
Thank you for the huge refactoring

PR #227, opened 2016-04-26
changed jsonforms.d.ts to jsonforms.ts

### LINKED ISSUE #226: jsonforms.d.ts should be a ts
We define an enum in the d.ts file which we cannot use in expressions as no code is generated for it.
We should change the d.ts to be a ts and thus get code for the enum

eneufeld:
Fixes #226


eneufeld:
@edgarmueller 
I also changed the tsconfig to not compile on save and changed the out dir, I can remove it form the commit if you prefer this to be an extra commit

edgarmueller:
@eneufeld No, it's fine. FYI, I also create https://github.com/eclipsesource/jsonforms/issues/235 along the way.

PR #239, opened 2016-04-28
removed duplicated dependencies, fixed dev and non dev

### LINKED ISSUE #136: Dependencies of the example app should be Dev-Dependencies
E.g. Bootstrap

This is to avoid unnecessary version conflicts when other users only want to use our framework via bower

eneufeld:
Fixes #136

@edgarmueller I removed font-awesome as I could not find any user

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,19 +13,16 @@
   },
   "dependencies": {
     "angular": "1.5.0",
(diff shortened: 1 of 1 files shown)


edgarmueller:
Thanks!

PR #241, opened 2016-04-29
Clean-up of PathUtil 

edgarmueller:
Small refactoring of PathUtil + JSDoc, also fixed some minor warnings.
Usage of [start case](https://en.wikipedia.org/wiki/Letter_case#Stylistic_or_specialised_usage) when beautifying labels.
Fixed name of inits function, which actually should be named init (as e.g. [here](http://www.scala-lang.org/api/2.7.7/scala/List.html#init)).

### DIFF
--- a/src/components/services/pathutil.spec.ts
+++ b/src/components/services/pathutil.spec.ts
@@ -10,6 +10,10 @@ describe("PathUtil", () => {
       expect(() => PathUtil.toPropertyAccessString(null)).toThrowError();
    });
 
+    it("should return an empty array when converting a path to fragments", () => {
+        expect(PathUtil.toPropertyFragments(undefined)).toEqual([]);
+    });
+
     it("should return the last fragment", () => {
         expect(PathUtil.lastFragment("foo/bar")).toBe("bar")
     });
--- a/src/components/services/pathutil.spec.ts
+++ b/src/components/services/pathutil.spec.ts
@@ -3,12 +3,24 @@ import "angular-mocks"
 
 import {PathUtil} from "./pathutil";
 
-describe("Path-Util", () => {
+describe("PathUtil", () => {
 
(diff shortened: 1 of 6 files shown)


edgarmueller:
@eneufeld This is a refactoring I did in my spare time to explore lodash. No issue or whatsoever tracks this.

src/components/services/pathutil.ts
```diff
@@ -1,80 +1,110 @@
...
     }
 
+    /**
+     * Returns the last fragment of the given path in a beautified manner.
+     *
+     * @param schemaPath the path from which to extract the beautified fragment
+     * @returns {any} the last fragment in a beautified manner
+     */
+    static beautifiedLastFragment(schemaPath: string): string  {
+        return _.flow(PathUtil.lastFragment, _.upperFirst, PathUtil.beautify)(schemaPath);
```
eneufeld: the _.uperFirst call is not necessary as the beautify method will upper case the string

src/components/services/pathutil.spec.ts
```diff
@@ -3,12 +3,28 @@ import "angular-mocks"
...
+        expect(PathUtil.lastFragment("foo/bar")).toBe("bar")
+    });
+
+    it("should beautify the last fragment", () => {
+        expect(PathUtil.beautifiedLastFragment("foo/bar")).toBe("Bar")
+    });
+
+    it("should filter indices from a path", () => {
+        expect(PathUtil.filterIndexes("foo/items/1/bar")).toBe("foo/items/bar")
+    });
```
eneufeld: a test case for a beautify of a property like firstName to be "First Name" would be good

eneufeld:
Thank you, the code is much better readable. I added two small comments, but they should/could be fixed separately as they are improvements

PR #246, opened 2016-04-29
fix hide of empty layouts

eneufeld:


edgarmueller:
Thanks for the quick fix.

PR #247, opened 2016-04-29
Update of README and ui.bootstrap version

edgarmueller:
Update of README and ui.bootstrap version

PR #248, opened 2016-04-30
fixed broken directives due to angular ui bootstrap version bump

eneufeld:
Fixes broken directives due to #247


edgarmueller:
Thanks!

PR #253, opened 2016-05-03
added categorization expandbar renderer

eneufeld:
Added additional renderer for categories as expandbars instead of tabs

### DIFF
--- a/bootstrap/jsonforms-bootstrap.ts
+++ b/bootstrap/jsonforms-bootstrap.ts
@@ -5,6 +5,8 @@ import booleanControlDirective from './controls/boolean/bootstrap-boolean.templa
 import arrayControlDirective from './controls/array/bootstrap-array.template';
 import controlDirective from './controls/bootstrap-control.template';
 import categorizationDirective from './layouts/categories/bootstrap-categorization.template';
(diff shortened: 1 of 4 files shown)


edgarmueller:
This one probably needs to be updated once #268 has been merged.

edgarmueller:
@eneufeld Do we need to update this PR?

eneufeld:
It is not important and as far as I can recall not used anywhere.
I will update it when I have some time.

eneufeld:
@edgarmueller can you take a look at this?
- rebased
- moved renderer to bootstrap

edgarmueller:
The regular case (`dev`) works fine, but the bootstrap renderer renders an empty form on the right-hand side  for me:

pr253

Do you experience this as well?

The material renderer ignores the custom `expandbar` option altogether, but I suppose this is intended behaviour.

eneufeld:
Yes, 
the problem is, that this renderer expects a specific structure of the categorization.
The current example doesn't fit. 
I will update the code and provide a dedicated example for this renderer.

eneufeld:
@edgarmueller 
I created a dedicated example, this must be refactored when we singlesource the examples.

PR #263, opened 2016-05-09
Added generation of d.ts

eneufeld:
Updated tsconfig.json to generate d.ts files in the dist/js folder.

@edgarmueller please review

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,9 +4,10 @@
     "noImplicitAny": false,
     "removeComments": true,
     "sourceMap": true,
+    "declaration":true,
     "module": "commonjs",
     "experimentalDecorators": true,
-    "outDir": "dist/src"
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -4,9 +4,10 @@
     "noImplicitAny": false,
     "removeComments": true,
     "sourceMap": true,
+    "declaration":true,
     "module": "commonjs",
     "experimentalDecorators": true,
-    "outDir": "dist/src"
+    "outDir": ""
```
edgarmueller: Does this have to be empty or can we simply get rid of the `outDir` property?

edgarmueller:
@eneufeld Looks good, just a question regarding the `outDir` property.

edgarmueller:
Thank you!

PR #265, opened 2016-05-09
Replace 'replace' for A2 compatiblity

edgarmueller:
JSON Forms can't not be used with A2 migration mode, since we use `replace` in our code base.
I'm not very happy with this fix, since it introduces additional DOM elements, but I couldn't think of something else in order to keep the restriction that CSS layout styles should only be applied in case the respective layouts are visible. 

Furthermore the PR also adds missing style capabilities for categorizations and groups.

This commit also contains two minor fixes that bugged me as well: that the person example is still called 'local' and that we did not yet include any usage info about JSON Forms in the README (although, of course, it's still WIP). If necessary I can also push them as single commits, I was just too lazy to do so :)


src/components/renderers/layouts/categories/categorization-directive.ts
```diff
@@ -7,20 +7,22 @@ class CategorizationDirective implements ng.IDirective {
...
-                        ng-repeat="category in vm.uiSchema.elements"
-                        select="vm.changeSelectedCategory(category)">
-                        <fieldset ng-if="vm.selectedCategory===category">
-                            <jsonforms-inner ng-repeat="child in category.elements"
-                                             ui-schema="child" >
-                            </jsonforms-inner>
-                        </fieldset>
-                    </uib-tab>
-                </uib-tabset>
+        <div class="jsf-categorization">
```
eneufeld: why not add the jsf-categorization class to the existing div?
i think, that the new div might destroy the grid that is created by bootstrap.

edgarmueller: Placing the class attribute onto the existing elements didn't work with the vertical/horizontal layout, that's why also introduced the additional element here as well, but I didn't actually give it a try. Will do.

eneufeld:
Looks good, thank you.

PR #268, opened 2016-05-23
Extracted Bootstrap

eneufeld:
- Bootstrap templates are now optional. To use you have to run build-bootstrap.
- extracted css to single files, currently must be copied by hand to dist
- fixed bugs with treemaster detail

@edgarmueller sorry this is a bit of a mess


examples/app/app-bootstrap.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+angular.module('makeithappen', [
+    'ngRoute',
+    //'ngResource',
+    //'ui.ace',
+    'jsonforms-bootstrap'
```
edgarmueller: Instead of copying the whole app.js file we also could just include this line and make sure it is loaded after app.js in index-bootstrap.html

src/components/renderers/controls/datetime/datetime-directive.ts
```diff
@@ -5,54 +5,35 @@ import {IUISchemaElement} from '../../../../jsonforms';
...
-    public openDate() {
-        this.isOpen = true;
+        scope.$watch('vm.modelValue[vm.fragment]', (newValue) => {this.updateDateObject(); });
     }
     protected modelChanged() {
         if (this.dt != null) {
             // returns a string in the form 'yyyy-mm-dd'
             this.modelValue[this.fragment] = this.dt.toISOString().substr(0, 10);
-        } else {
+        }else {
```
edgarmueller: This should be covered by tslint, but it isn't :weary:

src/components/renderers/controls/datetime/datetime-renderer.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('DateTimeTest', () => {
             }
         };
 
-    it("should be rendered",
+    xit("should be rendered",
```
edgarmueller: I guess theses tests are bootstrap specfic? Does it make sense to replace them then? If so, do we track this as an issue?

src/index.ts
```diff
@@ -4,14 +4,14 @@ import form from './components/form/form';
 import capitalize from './components/ng-services/capitalize/capitalize.filter';
 import pathResolver from './components/ng-services/pathresolver/pathresolver-service';
 
-import '../jsf-bootstrap.css';
-
-require('angular-ui-bootstrap');
-require('angular-ui-bootstrap/dist/ui-bootstrap-csp.css');
+// import '../jsf-bootstrap.css';
```
edgarmueller: I guess the commented out lines can be removed.

webpack/webpack.build.js
```diff
@@ -4,7 +4,7 @@ var webpack = require('webpack');
 
 module.exports = {
     entry: [
-        "bootstrap-webpack!./bootstrap.config.js",
+        // "bootstrap-webpack!./bootstrap.config.js",
```
edgarmueller: Remove commented out line

edgarmueller:
This already looks great and I have commented on some minor issues, but I also have some questions, since I'm am not entirely sure whether everything is supposed to work already:
- Does the horizontal layout work (without bootstrap, I saw the flex is being used at some point)? 
- Does master/detail work (without bootstrap), clicking user/enemies in the person example has no effect?
- Would it make sense to extract the whole bootstrap part and release it as a standalone NPM package that can be used with the 'core', i.e. it would overwrite existing renderers when it has been included.

eneufeld:
Hi thank you for the review.
- horizontal layout should work
- master/detail should work -> you have to press the chevron
- ideally I would like to have a jsonforms-bootstrap.js which would contain only the bootstrap renderers and bootstrap dependencies and which could be loaded after the jsonforms.js, but I don't know how to configure the build accordingly.

edgarmueller:
Regarding the last point: Yes, that's also what I'd like to have. About the build: nevermind, we leave that up to @sdirix :grin:

edgarmueller:
Ah, I forgot to copy a CSS file, of course. You should be able to use the [this plugin](https://www.npmjs.com/package/copy-webpack-plugin) and update the webpack files accordingly.

```JavaScript
var CopyWebpackPlugin = require('copy-webpack-plugin');
```
and in the `plugins` section add this entry:
```JavaScript
new CopyWebpackPlugin([
    { from: 'src/jsonforms.css' }
]),
```
Likewise for the bootstrap webpack.

src/components/renderers/controls/datetime/datetime-renderer.spec.ts
eneufeld: I would love to keep them, but I don't know how to get them running in a job eg test-bootstrap

PR #269, opened 2016-05-24
added resolvement of schema using jsonrefs

### LINKED ISSUE #158: JSONRefs version bump
We should update the JSONRefs version to latest (1.3). 
The problem is, that currently we have a problem executing the unit tests as with the new version always a Promise is returned and is not resolved before the end of the test.

eneufeld:
Fixes #158 

### DIFF
--- a/src/components/form/form-directive.ts
+++ b/src/components/form/form-directive.ts
@@ -83,15 +83,14 @@ export class FormController {
             let schema = values[0];
             this.uiSchema = <IUISchemaElement> values[1];
             let data = values[2];
-            // this.render(schema, data);
             let unresolvedRefs = JsonRefs.findRefs(schema);
-            if (unresolvedRefs === undefined || Object.keys(unresolvedRefs).length === 0) {
+            if (_.size(unresolvedRefs) === 0) {
                 this.render(schema, data);
             } else {
                 JsonRefs.resolveRefs(schema).then(
                     res => {
                         this.render(res.resolved, data);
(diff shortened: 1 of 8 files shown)


src/components/form/form-directive.ts
```diff
@@ -82,33 +83,49 @@ export class FormController {
...
-            services.add(new ScopeProvider(this.scope));
-            services.add(new SchemaProvider(schema));
-            services.add(new ValidationService());
-            services.add(new RuleService(this.PathResolver));
-
-            if (FormController.isDataProvider(this.scope.data)) {
-                dataProvider = this.scope.data;
+            // this.render(schema, data);
+            let unresolvedRefs = JsonRefs.findRefs(schema);
+            if (unresolvedRefs === undefined || Object.keys(unresolvedRefs).length === 0) {
```
edgarmueller: Does `findRefs` ever return undefined? The length check could be replaced with `_.size`.

src/components/form/form-directive.ts
```diff
@@ -82,33 +83,49 @@ export class FormController {
...
+                this.render(schema, data);
             } else {
-                dataProvider = new DefaultDataProvider(this.$q, data);
+                JsonRefs.resolveRefs(schema).then(
+                    res => {
+                        this.render(res.resolved, data);
+                        // Eugen: don't understand why this is needed in the remote case
```
edgarmueller: Probably because Angular can't know when this callback will be completed.

Issue #223, opened 2016-04-21
Configure webpack.dev correctly


edgarmueller:
Html and BrowserSync plugins do not work as expected.

eneufeld:
fixed with pr #273

PR #273, opened 2016-06-06
Configure webpack-dev-server task

edgarmueller:
Fix for #223 

Can be executed via `npm run dev`.

Re-introduces bower to avoid dependencies that are only necessary for the examples (like it is the case with angular-route ATM).


eneufeld:
Can't we use npm dev dependencies instead of bower?
Otherwise the bower.json points to edgarmueller/jsonforms instead of eclipsesource/jsonforms as well as the email

edgarmueller:
The problem is that I don't know how to reference the node_modules folder with the dev server's contentBase property set to the examples folder (maybe there is an option to do that, but I couldn't find any). Referencing the parent folder via .. didn't work for me.

edgarmueller:
@eneufeld This should work with bootstrap now as well (see minor refactoring). Can be executed via `npm run dev-bootstrap`.

eneufeld:
On Windows I still get:

$ npm run dev

> jsonforms@0.0.21 dev C:\Users\Eugen\git\jsonforms\jsonforms_master_eugen
> webpack --config webpack/webpack.dev.js; webpack-dev-server --config webpack/webpack.dev.js --NODE_ENV=dev --inline

path.js:8
    throw new TypeError('Path must be a string. Received ' +
    ^

TypeError: Path must be a string. Received [ 'webpack/webpack.dev.js;', 'webpack/webpack.dev.js' ]

eneufeld:
Thank you, looks good now

PR #277, opened 2016-06-15
Fix date issues

eneufeld:
This fixes problems with the bootstrap date renderer and the validation service leading to stack overflows


sdirix:
Thanks for including the tests! Looks good to me!

PR #279, opened 2016-06-23
Return comment if no fitting renderer was found

### LINKED ISSUE #267: Complete Rendering fails when no renderer is found for one UI schema element
The rendering should be more graceful and log the error (e.g. into the DOM as comment) but continue rendering

eneufeld:
Fixes #267

### DIFF
--- a/src/components/renderers/renderer-service.ts
+++ b/src/components/renderers/renderer-service.ts
@@ -33,8 +33,10 @@ class RendererServiceImpl implements RendererService {
         this.renderer.push({directiveName: directiveName, tester: tester});
     }
     getBestComponent(element: IUISchemaElement, dataSchema: any, dataObject: any): string {
-        let bestRenderer = _.maxBy(this.renderer,
-            renderer => renderer.tester(element, dataSchema, dataObject, this.pathResolver));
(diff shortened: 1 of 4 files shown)


eneufeld:
@edgarmueller please review

src/components/renderers/renderer-service.ts
```diff
@@ -29,10 +29,14 @@ class RendererServiceImpl implements RendererService {
         this.renderer.push({directiveName: directiveName, tester: tester});
     }
     getBestComponent(element: IUISchemaElement, dataSchema: any, dataObject: any): string {
-        let bestRenderer = _.maxBy(this.renderer, renderer =>
-            renderer.tester(element, dataSchema, dataObject, this.pathResolver)
-        );
-
+        let bestRenderer = _.maxBy(this.renderer, renderer => {
+            let result = renderer.tester(element, dataSchema, dataObject, this.pathResolver);
+            return result === NOT_FITTING ? null : result;
```
edgarmueller: Why do we need to return `null`? Why don't we compare `bestRenderer` with `NOT_FITTING`?

edgarmueller:
Looks good, thanks!

PR #280, opened 2016-06-23
renamed ui-schema to uischema

### LINKED ISSUE #259: Change "ui-schema" to "uischema" to fix angular2 incompabilities
We can also think about renaming the attribute to something different.

eneufeld:
Fixes #259
I also replaced all " to ' in the spec files as I had to touch them anyway.
@edgarmueller please review


edgarmueller:
Looks good, thanks

PR #283, opened 2016-06-29
extended the uigenerator and adapted arraycontrol accordingly

### LINKED ISSUE #235: UI Schema generator should also be able to generate groups
The respective methods then could be used from components such as the array renderer.

eneufeld:
Fixes #235

@edgarmueller please review

### DIFF
--- a/src/components/generators/jsonforms-uischemagenerator.ts
+++ b/src/components/generators/jsonforms-uischemagenerator.ts
@@ -58,8 +58,9 @@ export class UISchemaGenerator implements IUISchemaGenerator {
     }
 
     /**
-     * Wraps the given {@code uiSchema} in a VerticalLayout if there is none already.
-     * @param uiSchema The ui schema to wrap in a vertical layout.
+     * Wraps the given {@code uiSchema} in a Layout if there is none already.
+     * @param uiSchema The ui schema to wrap in a layout.
+     * @param layoutType The type of the layout to create.
      * @returns the wrapped uiSchema.
      */
     private static wrapInLayoutIfNecessary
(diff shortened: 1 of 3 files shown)


src/components/generators/jsonforms-uischemagenerator.ts
```diff
@@ -61,44 +62,48 @@ export class UISchemaGenerator implements IUISchemaGenerator {
      * @param uiSchema The ui schema to wrap in a vertical layout.
```
edgarmueller: Doc is out-of-date.

src/components/generators/jsonforms-uischemagenerator.ts
```diff
@@ -61,44 +62,48 @@ export class UISchemaGenerator implements IUISchemaGenerator {
...
             verticalLayout.elements.push(uiSchema);
             return verticalLayout;
         }
         return <ILayout>uiSchema;
     }
 
-    generateDefaultUISchema(jsonSchema: any): IUISchemaElement {
-        let uiSchema = this.generateUISchema(jsonSchema, [], '#', '');
-        return UISchemaGenerator.wrapInLayoutIfNecessary(uiSchema);
+    generateDefaultUISchema(jsonSchema: any, layoutType?: string): IUISchemaElement {
```
edgarmueller: I think we could use default parameters here instead of an optional and checking for `undefined`, but the interface signature should be fine already (I'm not sure though).

eneufeld: you are right, I forgot about default parameters, that would be better

edgarmueller:
Thanks!

PR #284, opened 2016-06-29
Rework GH-Pages

sdirix:
See http://sdirix.github.io/jsonforms/#/


eneufeld:
- the white hover doesn't look good in the status bar, as the twitter icon is gone then, we need to replace it with the light blue icon on hover then
- the drop down menus (visible if width is < 800px) are broken (the main menu doesn't close on selection and the docs menu doesn't even open)
- The mail link info@jsonforms.org on http://sdirix.github.io/jsonforms/#/docs/emfforms directly before the last paragraph should be an actual link
- is it posible to fix width on the docs site if width<800 or <640 px so that there are no scroll bars? this would be helpfull for mobile users

The Jekyll integration looks great, the issues mentioned here are already broken on the current site maybe we should fix them separatly

sdirix:
Additionally updated the tutorial.md and fixed the issues mentioned by @eneufeld. 

However the site is still not ideal for mobile users. We should take care of that in a separate issue / PR.

edgarmueller:
@sdirix Nice effort and I don't mean to disappoint you, but I'd prefer a regular text at the top beneath the logo. Perhaps re-use the tone of red we already use on the site.

edgarmueller:
@eneufeld Damn, I was too slow while commenting ;)

PR #286, opened 2016-07-02
Add uischema registry support

eneufeld:
This PR adds support for a ui schema registry.
- We probably need to refactor the generator to not be a service anymore and move it to the uischemaregistry "package"
- We need an example

### DIFF
--- a/src/components/form/form-directive.ts
+++ b/src/components/form/form-directive.ts
@@ -2,7 +2,7 @@ import 'angular';
 let JsonRefs = require('json-refs');
 
 import {IPathResolver} from '../services/pathresolver/jsonforms-pathresolver';
-import {IUISchemaGenerator} from '../generators/generators';
+import {UiSchemaRegistry} from '../ng-services/uischemaregistry/uischemaregistry-service';
 import {ISchemaGenerator} from '../generators/generators';
 import {PathResolver} from '../services/pathresolver/jsonforms-pathresolver';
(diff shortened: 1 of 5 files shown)


src/components/form/form-directive.ts
```diff
@@ -2,7 +2,7 @@ import 'angular';
 let JsonRefs = require('json-refs');
 
 import {IPathResolver} from '../services/pathresolver/jsonforms-pathresolver';
-import {IUISchemaGenerator} from '../generators/generators';
+import {UiSchemaRegistry} from '../ng-services/uischemaregistry/uischemaregistry-service';
```
edgarmueller: We should perhaps agree on how we want to treat abbreviations (that is, `Ui` or `UI`), but that's another discussion topic.

src/components/form/form.ts
```diff
@@ -2,6 +2,7 @@
 import * as angular from 'angular';
 import rendererService from '../renderers/renderer-service';
 import pathResolver from '../ng-services/pathresolver/pathresolver-service';
+import uischemaRegistry from '../ng-services/uischemaregistry/uischemaregistry-service';
```
edgarmueller: Naming: S should be capital case to be consistent with `uiSchemaGenerator`.

edgarmueller:
Looks good, only minor nitpick regarding naming of imports

edgarmueller:
Thanks!

PR #289, opened 2016-07-07
Configure typings for external use

edgarmueller:
Export typings and put generated source into root folder (also update .gitignore).  This enables API consumers to do imports a-la `import { ... } from "jsonforms"`.

I'm not sure whether all required types/values are exported but I'd add those opportunistically.


edgarmueller:
@eneufeld I tested by compiling locally and consuming the result in another NPM project.

src/components/form/form-directive.ts
```diff
@@ -166,11 +166,19 @@ export interface JsonFormsDirectiveScope extends ng.IScope {
...
+<div>
+    <form role='form' class='jsf-form rounded'></form>
+</div>`;
 
-export class JsonFormsDirective implements ng.IDirective {
 
+export class JsonFormsDirective implements ng.IDirective {
+    static $inject = ['$templateCache'];
+    constructor($templateCache: ng.ITemplateCacheService) {
+        $templateCache.put('form.html', formTemplate);
```
eneufeld: why not do this in a run block, just like in all other cases?

eneufeld:
The jsonforms.ts is now a mixup of two schema definitions (ui-schema and json-schema) and our exports, is this a good way to do it?

eneufeld:
Looks good, 
did't you have typescript warnings for using " in the imports instead of ' ?

edgarmueller:
I probably do..will update :sweat_smile:

eneufeld:
looks good, thank you

PR #291, opened 2016-07-13
Fix custom control example

edgarmueller:
Fix custom control example (used old `uiSchema` name).


edgarmueller:
@sdirix @eneufeld Can any of you have a look?

sdirix:
This works, however the custom control has no label.

Add the following code to show a label.
```
    vm.showLabel = true;
    vm.label = $scope.uischema.label;
```

sdirix:
I will open a separate issue for this

PR #347, opened 2016-08-08
Improve Master Detail Renderer

eneufeld:
Improved the Master Detail renderer
- it looks like a tree now (the non bootstrap version)
- added support for array and object inputs
- added support for label and image provider using the options object in the ui schema
- added support for adding and removing tree elements

@edgarmueller could you review if you have some spare time


edgarmueller:
LGTM without having a very detailed look. I also did some changed in the CSS styling (see #349) and since you are on currently N/A fulltime I'll just merge your changes for now but it's likely I'll make some adaptions. I'll let you know when I'm finished so you can a look.

edgarmueller:
@eneufeld I noticed that the bootstrap version of the master detail renderer seems to be broken after i merged this change. I tried to fix it myself, but wasn't able to do so due to limited time, hence I reverted the merge. Could you check that the bootstrap renderer is working and re-open the PR?

edgarmueller:
@eneufeld Please note that I merged #349 meanwhile, which changes the styling of the examples. It's likely that you will run into some conflicts, which should be easy to solve.

PR #356, opened 2016-08-12
MasterDetail Renderer

eneufeld:
Refactored the MasterDetail renderer
- looks like a tree
- supports object and array input
- supports add and remove
- supports label and image provider


eneufeld:
i fixed the bootstrap master detail renderer as this was the reason for the revert. I would suggest to remove the bootstrap master detail renderer completely as it doesn't provide any real new features

src/components/renderers/layouts/masterdetail/masterdetail-directives.ts
```diff
@@ -31,13 +58,19 @@ class MasterDetailController extends AbstractControl {
 class MasterDetailCollectionController {
     static $inject = ['$scope'];
     public instance: any;
-    public properties: any;
+    public schema: any;
+    public labelprovider: LabelProvider;
```
edgarmueller: I'd prefer for camel case/kebap case (for HTML) naming: `labelProvider`, `imageProvider`, ...

src/components/renderers/layouts/masterdetail/masterdetail-directives.ts
```diff
@@ -47,128 +80,194 @@ class MasterDetailCollectionController {
...
+    }
+    public updateHasContents(scope): void {
+        if (scope.child === this.instance) {
+            scope.hasContents = true;
+            return;
+        }
+        let child = scope.child;
+        let schemaToCheck = scope.schema.items;
+        let keys = _.keys(this.getArraySubSchemas(schemaToCheck));
+        for (let key of keys) {
```
edgarmueller: I think this for-loop could be replaced with `scope.hasContents = _.some(keys, key => child[key] !== undefined ...`

PR #398, opened 2016-09-09
fixed validation icon not to be in the control

eneufeld:
Fixes relevant for #392


eneufeld:
@edgarmueller can you look at this

Issue #339, opened 2016-08-05
Add validation to email field


koegel:

eneufeld:
So the main problem is, that tv4 doesn't support the predefined formats, see: [https://github.com/geraintluff/tv4/issues/100](url) .
We need to add tv4-formats as a dependency and initialize it.

eneufeld:
closed via #400 and #401

PR #400, opened 2016-09-12
added support for formats for tv4

eneufeld:
Needed change for #339

### DIFF
--- a/src/components/services/services.ts
+++ b/src/components/services/services.ts
@@ -1,14 +1,14 @@
 import {SchemaElement} from '../../jsonschema';
 import {IUISchemaElement} from '../../uischema';
-import {PathUtil} from "./pathutil";
+import {PathUtil} from './pathutil';
 
 // TODO: replace
 let tv4 = require('tv4');
 let formats = require('tv4-formats');
 let validator;
-if(tv4!=null){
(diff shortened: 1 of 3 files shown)


src/components/services/services.ts
```diff
@@ -4,7 +4,14 @@ import {PathUtil} from "./pathutil";
 
 // TODO: replace
 let tv4 = require('tv4');
-
+let formats = require('tv4-formats');
+let validator;
+if(tv4!=null){
```
edgarmueller: TSLint should complain here (missing WS)

edgarmueller:
Just one word on this: according to https://github.com/ebdrup/json-schema-benchmark, tv4 is quite slow in comparison to ajv (which hinders rendering performance) so we might want to think about switching to ajv soon, as it supports formats out-of-the box.
Otherwise, looks good, except some TSLint warnings.

eneufeld:
I'll check the ts lint. 
I would suggest to fix it with the tv4 formats and then do a second refactoring to replace tv4 with ajv

edgarmueller:
I've [opened an issue](https://github.com/eclipsesource/jsonforms/issues/409) for replacing the validator. Otherwise looks good.

PR #422, opened 2016-10-03
Add Delete Button to ArrayControl

### LINKED ISSUE #346: Add option to array control to remove entries
Curretnly we only support creating entries

sdirix:
Fixes #346

Adds Delete Button to Plain, Bootstrap and Material ArrayControl Renderer.

Uses Material Icon Font in Material renderer which introduces a dependency to this font. See example index html.


src/components/renderers/controls/array/array.directive.ts
```diff
@@ -87,9 +108,21 @@ class ArrayController extends AbstractControl {
         this.resolvedData[this.fragment].push(_.clone(this.submitElement));
         this.submitElement = {};
     }
+    public deleteCallback(element: any) {
+        var index = this.resolvedData[this.fragment].indexOf(element);
+        if(index !== -1){
```
edgarmueller: Can you check for tslint warnings? Seems as if whitespace is missing here.

PR #431, opened 2016-10-07
changed content-box to border-box

### LINKED ISSUE #416: [Bootstrap] The enum control is higher than a normal control
This is probably a css issue.

eneufeld:
Fixes #416

### DIFF
--- a/src/jsonforms.css
+++ b/src/jsonforms.css
@@ -22,7 +22,7 @@
   padding: 1px 0px;
 }
 .jsf-control select.form-control {
-  box-sizing: content-box;
+  box-sizing: border-box;
   padding: 2px 0px;
 }
 .jsf-control-boolean {


edgarmueller:
@eneufeld The fix makes sense, but how I can reproduce the actual errorr? So far, I didn't notice any difference. 
What I noticed during testing is that with `border-box` being set, it seems as if the actual combo box is somewhat smaller than a neighboring text field. Is this intended?
Furthermore, does this fix only apply to to our vanilla CSS or do we have to make adoptions for material/bootstrap as well?

eneufeld:
You can see the error in the MakeItHappen example. The ComboBox there is a bit bigger then all the other controls there. This could also be an issue with wrong/missing/wrong order css.
I didn't notice any changes otherwise. But I will retest with all 3 versions.

eneufeld:
I close this and let us continue the discussion the issue on the bug.

PR #441, opened 2016-11-17
removed css inclusion

### LINKED ISSUE #416: [Bootstrap] The enum control is higher than a normal control

eneufeld:
Fixes #416


edgarmueller:
We shouldn't forget about document the CSS usage; I've created #442 in order to not forget about this.

PR #487, opened 2017-02-02
Material design fix for reading error data.

bosstjann:
Problems exists if you are using it with material design and angular version 1.5.8


edgarmueller:
Thank you!

PR #503, opened 2017-02-15
fixed double registration

eneufeld:

### DIFF
--- a/src/json-forms.ts
+++ b/src/json-forms.ts
@@ -42,15 +42,18 @@ export class JsonForms extends HTMLElement {
     if (this.lastChild !== null) {
       this.removeChild(this.lastChild);
     }
-
-    if (this.services.length === 0) {
-      this.createServices();
-      this.dataService.initialRootRun();
+    if (this.services.length !== 0) {
+      this.services.forEach(service => service.dispose());
+      this.services = [];
     }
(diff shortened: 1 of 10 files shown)


src/json-forms.ts
```diff
@@ -40,30 +42,39 @@ export class JsonForms extends HTMLElement {
     if (this.lastChild !== null) {
       this.removeChild(this.lastChild);
     }
-
-    if (this.services.length === 0) {
-      this.createServices();
-      this.dataService.initialRootRun();
+    if (this.services.length !== 0) {
```
edgarmueller: Is this guard necessary?

src/json-forms.ts
```diff
@@ -40,30 +42,39 @@ export class JsonForms extends HTMLElement {
...
-    this.render();
+    if (this.allowDynamicUpdate) {
+      this.render();
+    }
   }
 
   set dataSchema(dataschema: JsonSchema) {
     this.dataschema = dataschema;
-    this.render();
+    if (this.allowDynamicUpdate) {
```
edgarmueller: Since the `render` call is guarded by `this.allowDynamicUpdate` so often, wouldn't it makes sense to move the check into the `render` method directly?

PR #504, opened 2017-02-16
Add UI schema generator

edgarmueller:
Ported UI schema generator.

### DIFF
--- a/src/generators/ui-schema-gen.ts
+++ b/src/generators/ui-schema-gen.ts
@@ -13,13 +13,6 @@ const createLayout = (layoutType: string): Layout => ({
 });
 
 /**
- * Determines if the property should be ignored because it is a meta property
- */
-const isIgnoredProperty = (propertyKey: string, propertyValue: any): boolean =>
-    // could be a string (json-schema-id). Ignore in that case
-    propertyKey === 'id' && typeof propertyValue === 'string';
-
-/**
  * Derives the type of the jsonSchema element
  */
 const deriveType = (jsonSchema: any): string => {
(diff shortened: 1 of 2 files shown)


test/ui-schema-gen.test.ts
```diff
@@ -0,0 +1,344 @@
+import test from 'ava';
+import generateDefaultUISchema, { startCase } from '../src/generators/ui-schema-gen';
+import { JsonSchema } from '../src/models/jsonSchema';
+import { Layout, ControlElement } from '../src/models/uischema';
+
+test('startCase', t => {
```
eneufeld: needed?

src/generators/ui-schema-gen.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (uiSchema['elements'] === undefined) {
+        const verticalLayout: Layout = createLayout(layoutType);
+        verticalLayout.elements.push(uiSchema);
+        return verticalLayout;
+    }
+    return <Layout>uiSchema;
+};
+
+const generateUISchema =
+    (jsonSchema: any, schemaElements: UISchemaElement[],
```
eneufeld: should we use the JSONSchema Interface here?

test/ui-schema-gen.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        ]
+    };
+    t.deepEqual(generateDefaultUISchema(schema), uiSchema);
+});
+
+test('ignore json-schema id attributes', t => {
+    const schema: JsonSchema = {
+        type: 'object',
+        properties: {
+            id: 'ignore me',
```
edgarmueller: Schema is actually invalid, value must be an object. Remove test all together.

src/generators/ui-schema-gen.ts
edgarmueller: Yep, makes sense.

test/ui-schema-gen.test.ts
edgarmueller: Ironically, the test case failures I had after porting the generator were all related to labels ;)
So, I'd argue to just leave it as is.

PR #505, opened 2017-02-16
Add JSON schema generator

edgarmueller:
Port JSON schema generator with some modifications.

### DIFF
--- a/src/generators/schema-gen.ts
+++ b/src/generators/schema-gen.ts
@@ -6,14 +6,15 @@ const RequiredProperties   = "required";
 
 type Properties = {[property: string]: JsonSchema}
 
-const distinct = (array: Array<any>, discriminator: (item: any) => any): Array<any> => {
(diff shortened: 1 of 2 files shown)


src/generators/schema-gen.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const AdditionalProperties = "additionalProperties";
+const RequiredProperties   = "required";
+
+type Properties = {[property: string]: JsonSchema}
+
+const distinct = (array: Array<any>, discriminator: (item: any) => any): Array<any> => {
+    const known = {};
+    return array.filter(item => {
+        const discriminatorValue = discriminator(item);
```
eneufeld: discriminatorValue must be a string, in order to do the check in line 13
the discriminator can then be defined to be (item: any) => string
or am I mistaken?

edgarmueller: That's correct, will update.

PR #507, opened 2017-02-23
Basic Array control

edgarmueller:
Makes use of generators to implement a basic array control.

### DIFF
--- a/src/services/rule.service.ts
+++ b/src/services/rule.service.ts
@@ -68,8 +68,4 @@ class JsonFormsRuleService implements DataChangeListener, JsonFormService {
         uischema => this.evaluate(uischema, data))
       );
   }
-
-  schemaChanged(dataSchema: JsonSchema) {
-
-  }
 }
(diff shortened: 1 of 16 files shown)


src/services/rule.service.ts
```diff
@@ -68,4 +68,8 @@ class JsonFormsRuleService implements DataChangeListener, JsonFormService {
         uischema => this.evaluate(uischema, data))
       );
   }
+
+  schemaChanged(dataSchema: JsonSchema) {
+
```
eneufeld: left over?

src/services/validation.service.ts
```diff
@@ -14,27 +16,42 @@ class JsonFormsValidator implements DataChangeListener, JsonFormService {
...
     dataService.registerChangeListener(this);
-    const ajv = new AJV({allErrors: true, jsonPointers: true, errorDataPath: 'property'});
     this.validator = ajv.compile(dataSchema);
     this.parseUiSchema(uiSchema);
   }
 
   isRelevantKey = (_: ControlElement): boolean => true;
 
+  schemaChanged(dataSchema: JsonSchema) {
+    this.validator = ajv.compile(dataSchema);
```
eneufeld: left over?

PR #508, opened 2017-03-01
added textarea renderer

eneufeld:


edgarmueller:
Looks good, maybe add a test if you have too much time in a follow-up commit

PR #527, opened 2017-03-13
Fix tslint warnings

edgarmueller:
Update `tslint.json` and fix tslint warnings.

### DIFF
--- a/src/generators/schema-gen.ts
+++ b/src/generators/schema-gen.ts
@@ -1,10 +1,10 @@
 
-import {JsonSchema} from "../models/jsonSchema";
+import {JsonSchema} from '../models/jsonSchema';
 
-const AdditionalProperties = "additionalProperties";
-const RequiredProperties   = "required";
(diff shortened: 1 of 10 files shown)


src/generators/schema-gen.ts
```diff
@@ -19,9 +19,6 @@ const distinct = (array: Array<any>, discriminator: (item: any) => string): Arra
     });
 };
 
-export const generateJsonSchema = (instance: Object): JsonSchema =>
-    generateJsonSchemaWithOptions({})(instance);
-
 export const generateJsonSchemaWithOptions = (options: any) => (instance: Object): JsonSchema => {
```
eneufeld: can'T we use a default parameter and thus exclude one method?

edgarmueller: Yes, I guess that would make sense, but I'd do that in a follow-up commit.

eneufeld: yes, totally fine with me

PR #528, opened 2017-03-13
Remove obsolete callbacks from array control

edgarmueller:
Also included fix suggested by Eugen to make use of default parameter for options parameter during schema generation.

### DIFF
--- a/src/generators/schema-gen.ts
+++ b/src/generators/schema-gen.ts
@@ -19,7 +19,7 @@ const distinct = (array: Array<any>, discriminator: (item: any) => string): Arra
     });
 };
 
-export const generateJsonSchema = (options: any = {}) => (instance: Object): JsonSchema => {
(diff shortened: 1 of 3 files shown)


src/generators/schema-gen.ts
```diff
@@ -19,7 +19,7 @@ const distinct = (array: Array<any>, discriminator: (item: any) => string): Arra
     });
 };
 
-export const generateJsonSchemaWithOptions = (options: any) => (instance: Object): JsonSchema => {
+export const generateJsonSchemaWithOptions = (options: any = {}) => (instance: Object): JsonSchema => {
```
eneufeld: would it not make sense to rename this one to generateJsonSchema and remove the old generateJsonSchema?

edgarmueller: Yes, of course, I also realized this after pushing my changes.

PR #531, opened 2017-03-14
Add initial test for json-forms element based on wct

edgarmueller:
Adds a simple initial test for the actual json-forms element.


wct-test/json-forms.html
```diff
@@ -0,0 +1,31 @@
+<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <script src="../node_modules/webcomponents.js/webcomponents-lite.js"></script>
+    <script src="../node_modules/web-component-tester/browser.js"></script>
+    <script src="./native-shim.js"></script>
```
eneufeld: can't we use the native-shim from the example or move this to some common place, so that we do not copy it multiple times?

wct-test/json-forms.html
```diff
@@ -0,0 +1,31 @@
...
+    <script src="../dist/jsonforms.js"></script>
+</head>
+<body>
+<script>
+    suite('json-forms element', function() {
+        test('is awesomest', function() {
```
eneufeld: this will allow us to create integration tests only?

edgarmueller: Yes, although we should be able to get some coverage via [this](https://github.com/thedeeno/web-component-tester-istanbul) or something similar.

PR #537, opened 2017-03-28
Test coverage improvement

eneufeld:
- Moved classes out of the core.ts into separate files
- added test cases for core services


src/webcomponents.d.ts
```diff
@@ -10,11 +10,11 @@
 //   readonly customElements: CustomElementRegistry;
 // }
 
-interface CustomElementRegistry {
-  define(name: string, constructor: Function, options?: ElementDefinitionOptions): void;
-  get(name: string): any;
-  whenDefined(name: string): Promise<void>;
-}
+// interface CustomElementRegistry {
```
edgarmueller: Can we remove this?

PR #538, opened 2017-03-31
Improve css

eneufeld:
Improved the example css
added a table renderer
fixed the tree renderer


src/renderers/additional/tree-renderer.ts
```diff
@@ -87,78 +117,166 @@ class TreeRenderer extends Renderer implements DataChangeListener {
...
+    return JSON.stringify;
+  }
+  private addElement(key: string, data: Object, schema: JsonSchema, li: HTMLLIElement): void {
+    if (data[key] === undefined) {
+      data[key] = [];
+    }
+    const childArray = data[key];
+    const newData = {};
+    const length = childArray.push(newData);
+    let subChildren = li.getElementsByTagName('ul');
```
edgarmueller: Can be const

src/renderers/additional/tree-renderer.ts
```diff
@@ -87,78 +117,166 @@ class TreeRenderer extends Renderer implements DataChangeListener {
...
+      li.appendChild(childParent);
+    }
+    this.expandObject(newData, childParent, schema.properties[key].items,
+      toDelete => childArray.splice(length - 1, 1));
+  };
+  private expandObject(data: Object, parent: HTMLUListElement, schema: JsonSchema,
+      deleteFunction: (element: Object) => void): void {
     const li = document.createElement('li');
+    const div = document.createElement('div');
+    if (this.uischema.options['imageProvider']) {
```
edgarmueller: Is is possible that the uischema has not options?

src/renderers/additional/tree-renderer.ts
```diff
@@ -87,78 +117,166 @@ class TreeRenderer extends Renderer implements DataChangeListener {
...
     });
+
     parent.appendChild(li);
   }
-  private renderDetail(element: Object, label: HTMLSpanElement): void {
+  private renderChildren
+    (array: Array<Object>, schema: JsonSchema, li: HTMLLIElement, key: string): void {
+    let ul: HTMLUListElement;
+    const children = li.children;
+    for (let i = 0; i < children.length; i++) {
```
edgarmueller: Extract method and rename method to findLastUl or something similar.

src/renderers/label.util.ts
```diff
@@ -19,22 +19,22 @@ const deriveLabel = (controlElement: ControlElement): string => {
 
 const getLabelObject = (withLabel: ControlElement): ILabelObject => {
     const labelProperty = withLabel.label;
+    const derivedLabel = deriveLabel(withLabel);
```
edgarmueller: Why not export and use startCase?

src/path.util.ts
```diff
@@ -9,14 +9,16 @@ export const getValuePropertyPair = (instance: any, path: string):
       validPathSegments
           .slice(0, validPathSegments.length - 1)
           .reduce((curInstance, pathSegment) => {
-              if (!curInstance.hasOwnProperty(pathSegment)) {
-                  curInstance[pathSegment] = {};
+            const decodedSegment = decodeURIComponent(pathSegment);
```
eneufeld: sure, would that change anything on the performance?

src/path.util.ts
```diff
@@ -26,8 +28,10 @@ export const toDataPath = (path: string): string =>
 export const resolveSchema = (schema: JsonSchema, path: string): JsonSchema => {
   const validPathSegments = path.split('/');
   const resultSchema = validPathSegments.reduce((curSchema, pathSegment) =>
-      pathSegment === '#' ? curSchema : curSchema[pathSegment], schema);
-  if (resultSchema.$ref !== undefined) {
+      pathSegment === '#' || pathSegment === undefined || pathSegment === '' ?
```
eneufeld: what exactly do you want to extract?

src/renderers/additional/tree-renderer.ts
eneufeld: yes, you're right

src/path.util.ts
edgarmueller: I don't think so, but it seems clearer to me since we only want decoded segments in the end anyway.

src/path.util.ts
edgarmueller: The `pathSegment === '#' || pathSegment === undefined || pathSegment === ''` condition

Issue #543, opened 2017-04-04
Move maxBy to util 


eneufeld:
max by is defined in uischema.registry.ts it could be used in the rendererservice too.

we should also discuss the possibility to use underscore.js

PR #552, opened 2017-04-06
Fix issue #543 and errors in unregister methods

edgarmueller:
* Add lodash
* Replace occurrences of maxBy
* Update renderer service and ui schema registry
* Fix errors in unregistration of both
* Update tests

### DIFF
--- a/src/core/renderer.service.ts
+++ b/src/core/renderer.service.ts
@@ -1,3 +1,4 @@
+import * as _ from 'lodash';
 import {UISchemaElement} from '../models/uischema';
 import {JsonSchema} from '../models/jsonSchema';
 import {DataService} from './data.service';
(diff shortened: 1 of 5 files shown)


src/core/renderer.service.ts
```diff
@@ -9,35 +10,30 @@ export interface RendererTester {
 export class RendererService {
   private renderers: Array<{tester: RendererTester, renderer: string}> = [];
   registerRenderer(tester: RendererTester, renderer: string): void {
-    this.renderers.push({tester: tester, renderer: renderer});
+    this.renderers.push({tester, renderer});
   }
   unregisterRenderer(tester: RendererTester, renderer: string): void {
-    const index = this.renderers.indexOf({tester: tester, renderer: renderer});
```
edgarmueller: This `indexOf` call always returned -1 since we are only able to compare testers (which are essentially functions) via strict equality (i.e. references). `splice` 1st argument may also be negative, see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice).

Issue #548, opened 2017-04-06
Create alpha NPM release


edgarmueller:

eneufeld:
Is available using npm install jsonforms@next , because of a wrong push to npm the release on npm is already called alpha.3, maybe rename this sprint ;-)

PR #554, opened 2017-04-07
added label layout, added scripts for release

eneufeld:
Preparation for  #548

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,29 @@
 {
-  "name": "webcomponents-jsonforms",
+  "name": "jsonforms",
   "private": true,
-  "version": "0.0.1",
+  "version": "2.0.0-Alpha2",
   "license": "MIT",
   "directories": {
     "test": "test"
+  },
+  "main": "dist/ts-build/index.js",
+  "typings": "dist/ts-build/index.d.ts",
+  "scripts": {
+    "start": "npm run dev",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,11 +1,34 @@
 {
-  "name": "webcomponents-jsonforms",
-  "private": true,
-  "version": "0.0.1",
+  "name": "jsonforms",
+  "version": "2.0.0-alpha.3",
+  "description": "Generating forms for JSON schema",
```
edgarmueller: I'd argue it's rather from JSON schema ;)

package.json
```diff
@@ -1,11 +1,34 @@
 {
-  "name": "webcomponents-jsonforms",
-  "private": true,
-  "version": "0.0.1",
+  "name": "jsonforms",
```
edgarmueller: Should we already use jsonforms as the name? I think we we'd need to update the existing docs for version 1 first, as `npm install` will fetch the latest version per default.

package.json
eneufeld: right, i copied the old description.

package.json
eneufeld: not if you upload it with a tag, eg next

package.json
eneufeld: I think we must go through all descriptions in a separate commit and make it consisten

package.json
edgarmueller: Ah, ok, didn't know that.

package.json
edgarmueller: I am ok with that.

PR #555, opened 2017-04-08
Initial Support for Material Design

eneufeld:
This is an initial support for material design.
this also fixes some issues with the tsc configuration
must be rebased after #554 is merged


edgarmueller:
I think this has overlapping changes with #554, right? I'll look at this, once rebased.

eneufeld:
This is outdated

Issue #544, opened 2017-04-05
Reintroduce Testers


eneufeld:
Often we check whether a  JSONSchema describes an object, an array or something else. 
We should reintroduce the testers we had earlier

PR #556, opened 2017-04-09
Fix issue #544: Add initial support for testers

edgarmueller:
* Add testers API
* Update existing testers
* Fix lint warnings


src/renderers/controls/number.control.ts
```diff
@@ -1,14 +1,18 @@
-import { UISchemaElement, ControlElement } from '../../models/uischema';
-import { JsonSchema } from '../../models/jsonSchema';
-import { BaseControl } from './base.control';
-import { JsonFormsRenderer } from '../renderer.util';
-import { resolveSchema } from '../../path.util';
+import {BaseControl} from './base.control';
+import {JsonFormsRenderer} from '../renderer.util';
+import {
```
eneufeld: why break into multiple lines?

test/testers.test.ts
```diff
@@ -0,0 +1,220 @@
...
+});
+
+test('refEndsWith checks whether the ref of a control ends with a certain string', t => {
+    const uiSchema: ControlElement = {
+        type: 'Control',
+        scope: {
+            $ref: '#/properties/bar'
+        }
+    };
+    t.true(refEndsWith('bar')(uiSchema, undefined))
```
eneufeld: should this test not show a difference to the refEndIs method, by eg comparing more then the last segment? or less?

src/core/testers.ts
```diff
@@ -0,0 +1,155 @@
...
+            return false;
+        }
+        const schemaPath = uiSchema['scope']['$ref'];
+        return !_.isEmpty(schemaPath) && _.last(schemaPath.split('/')) === expected;
+    };
+
+/**
+ * A tester that will always return true.
+ */
+export const always: Tester = () => true;
```
eneufeld: this is not used anywhere, right? should we even add it then?

PR #558, opened 2017-04-11
improved setup

eneufeld:
- Added test tsconfig, thus the normal build doesn't compile the tests
- moved the index.html from example to dedicated folder so we can have multiple different examples, eg with material design
- removed unneeded files


example_library/bs-config.json
```diff
@@ -5,7 +5,7 @@
         "routes": {
             "/jsonforms.js": "dist/jsonforms.js",
             "/jsonforms.js.map": "dist/jsonforms.js.map",
-            "/native-shim.js": "example/native-shim.js",
```
edgarmueller: The example_library script isn't working for me (this might have been the case before). First, I think the `lite-server` devDependency is missing in `package.json`. Once installed, I get an error: 

```
Uncaught TypeError: Cannot read property 'prototype' of undefined
    at (index):25
``` 
which is this line:

```
Object.setPrototypeOf(CustomElement.prototype, JSONForms.Renderer.prototype);
```

eneufeld: we need to run npn run bundle before, I will fix both issues

PR #586, opened 2017-06-22
Schema Service

eneufeld:
Added a schema service which provides functionality needed to support the creation of a dialog in the tree


test/schema.service.test.ts
```diff
@@ -0,0 +1,844 @@
...
+        }
+      }
+    },
+    type: 'object',
+    properties: {
+      persons: {type: 'array', items: {$ref: '#/definitions/person'}}
+    }
+  } as JsonSchema;
+  const service: SchemaService = new SchemaServiceImpl(schema);
+  const properties = service.getContainmentProperties(schema);
```
edgarmueller: Two things I noted when calling this method:
* The method will modify the original schema that has been passed in. This should be mentioned in the JSDoc
* The generated ID field in this contains two hashes and looks like this `##/definitions/person`. I don't think that's on purpose

src/json-forms.ts
```diff
@@ -156,4 +162,14 @@ export class JsonFormsElement extends HTMLElement {
         this.services.push(new service(this.dataService, dataSchema, uiSchema))
     );
   }
+  private instantiateSchemaIfNeeded(schema: JsonSchema): void {
```
eneufeld: created issue #589

edgarmueller:
Looks good now, thanks.

PR #598, opened 2017-07-19
Jsonforms2 extended add

### LINKED ISSUE #597: Containment Property: Add elements relative to existing elements
The containment properties returned by the schema service only allow adding an element to the end of the containment.
The containment property should allow to add an element at a position relative (before or after) to an element which already exists in the containment.

lucas-koehler:
Fixes #597 Added extended adding functionality to containment properties

- Containment properties now support to add elements relative to an existing element.
- It can be specified whether the new element is added before or after the reference element.
- Added test cases to the schema service tests.
- Fixed an index bug in the delete function of containment properties


src/core/schema.service.ts
```diff
@@ -38,20 +38,45 @@ const findAllRefs = (schema: JsonSchema, result: ReferenceSchemaMap = {}): Refer
...
+  if (neighbourValue) {
+    const index = childArray.indexOf(neighbourValue);
+    if (insertAfter) {
+      if (index >= 0 && index < (childArray.length - 1)) {
+        childArray.splice(index + 1, 0, valueToAdd);
+
+        return;
+      }
+      // TODO proper logging
+      console.log('Warning: Could not add the new value after the given neighbour value.' +
```
edgarmueller: If this is a warning, use `console.warn`

src/core/schema.service.ts
```diff
@@ -38,20 +38,45 @@ const findAllRefs = (schema: JsonSchema, result: ReferenceSchemaMap = {}): Refer
...
+
 };
 const deleteFromArray = (key: string) => (data: object) => (valueToDelete: object) => {
   const childArray = data[key];
   if (!childArray) {
     return;
   }
   const indexToDelete = childArray.indexOf(valueToDelete);
-  childArray.splice(indexToDelete - 1, 1);
+  childArray.splice(indexToDelete, 1);
```
edgarmueller: Makes sense, but it seems as we don't have any test case that reflects the incorrect delete behaviour (the commit does not alter any tests)? If so, could you add one?

lucas-koehler: Adapted existing delete test case to also check that the correct data was deleted (So far it only checked that something was deleted).

edgarmueller:
Thanks!

PR #601, opened 2017-07-21
added an id to controls

### LINKED ISSUE #590: Controls should have id
The Controls should have the id property in order to reference specific controls by their id as in jsonforms1

eneufeld:
Fixes #590


edgarmueller:
While there are no issues with the implementation itself, I think we can improve upon the current solution (which is basically based on JSON Forms 1 as we know) while we're at it:

* We should make use of classes instead of `id` as one can have multiple controls pointing to the same sub schema (or to be more precise: controls with the same `$ref` value) and handle the issue of selecting one specific control differently (would be a follow-up issue then). One possibility would be to introduce an id field within our options that allows users to assign an ID that can be used as CSS selector during styling, but of course this is up for discussion.
* Users should not need to escape CSS selectors as this might get clumsy quite fast. Instead we should think about a mapping of special characters like `#` and `/` (see [this list](https://mathiasbynens.be/notes/css-escapes) for an overview over characters that need to be escaped), for instance, instead of using `#\#\/properties\/date` as a CSS selector we could map `#` to `doc` or `root` or something similar and `/` to `-` resulting in `.root-properties-date`. We'd then of course need to document that mapping somewhere.

What do you think?

eneufeld:
I agree that it makes more sense to use set the ref-path as a class and not as an id.
I also agree that we should open a follow up to allow the user to define his own id.
The escaping is ugly, that is true. I would implement the escaping as you suggested and add a wiki page describing the current behavior.

edgarmueller:
Thanks. I've created follow-up issues #608 and #609 as discussed.

PR #603, opened 2017-07-24
Changed renderer to use StyleRegistry

### LINKED ISSUE #577: Use style registry in all renderers and finish "none"/"vanilla" style
apply the style class names from the style registry in all controls to all (relevant?) elements and also at least finish the "none"/"vanilla" style

eneufeld:
Fixes #577 for the vanilla case
Fixed an issue with the example switch

### DIFF
--- a/test/renderers/number.control.test.ts
+++ b/test/renderers/number.control.test.ts
@@ -4,6 +4,7 @@ import '../helpers/setup';
 // tslint:disable:ordered-imports
 import { DataService } from '../../src/core/data.service';
 // tslint:enable:ordered-imports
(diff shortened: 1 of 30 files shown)


test/renderers/number.control.test.ts
```diff
@@ -46,7 +58,7 @@ test('NumberControlTester', t => {
   t.is(numberControlTester({type: 'Control'}, undefined), -1);
 });
 
-test('NumberControl tester with wrong schema type', t => {
+test('NumberCostyleNamentrol tester with wrong schema type', t => {
```
edgarmueller: Typo

PR #626, opened 2017-08-09
Refactor reference property, allow ref. property to handle multi references

lucas-koehler:
This PR refactors the ReferenceProperty implementation of the schema service to make the implementation locations more consistent and allow the reference property implementation to properly handle multi references using ids to identify referenced objects.

- Adapted reference properties get and add functions to properly handle
ids and multi refs.
- Added constructor parameter to ReferencePropertyImpl for find
reference target function
- Implemented function in schema.service.impl and provide it to created
reference properties
- Removed obsolete resolveReference function (same purpose as getData)
from ReferenceProperty
- Removed obsolete constructor parameters from ReferencePropertyImpl
- Added and adapted reference property test cases

### DIFF
--- a/src/core/schema.service.impl.ts
+++ b/src/core/schema.service.impl.ts
@@ -70,36 +70,57 @@ const deleteFromArray = (key: string) => (data: object) => (valueToDelete: objec
 const getArray = (key: string) => (data: object) => {
   return data[key];
 };
-const addReference = (schema: JsonSchema, variable: string, pathToContainment: string) =>
+const addReference = (schema: JsonSchema, identifyingProperty: string, variable: string) =>
   (root: Object, data: Object, toAdd: object) => {
-    const containment = pathToContainment
-      .split('/')
-      .reduce(
-        (elem, path) => {
-          if (path === '#' || path === '') {
-            return elem;
-          }
 
-          return elem[path];
-        },
-        root
-      );
-    const index = (containment as Object[]).indexOf(toAdd);
+    const refValue = toAdd[identifyingProperty];
     if (schema.properties[variable].type === 'array') {
       if (!data[variable]) {
         data[variable] = [];
       }
-      data[variable].push(index);
+      data[variable].push(refValue);
     } else {
-      data[variable] = index;
+      data[variable] = refValue;
     }
   };
-const getReference = (href: string, variable: string, variableWrapped: string) =>
-  (root: Object, data: Object) => {
-    const variableValue = data[variable];
-    const pathToObject = href.replace(variableWrapped, variableValue);
+const resolveRef = (schema: JsonSchema, findTargets: (rootData: Object) => Object[],
+                    identifyingProperty: string, variable: string) =>
+  (rootData: Object, data: Object) => {
+    if (_.isEmpty(data) || _.isEmpty(identifyingProperty)) {
+      return null;
+    }
+    // get all objects that could be referenced.
+    const candidates = findTargets(rootData);
 
-    return pathToObject
+    if (schema.properties[variable].type === 'array') {
+      const ids: object[] = data[variable];
+      const resultList = candidates.filter(value => {
+        return ids.indexOf(value[identifyingProperty]) > -1;
+      });
+
+      return resultList;
+    } else {
+      // use identifying property to identify the referenced object
+      const resultList = candidates.filter(value => {
+        return value[identifyingProperty] === data[variable];
+      });
+
+      if (_.isEmpty(resultList)) {
+        return null;
+      }
+      if (resultList.length > 1) {
+        throw Error('There was more than one possible reference target with value \'' +
+                    JSON.stringify(data[variable]) + '\' in its identifying property \'' +
+                    identifyingProperty + '\'.');
+      }
+
+      return _.first(resultList);
+    }
+  };
+
+const findReferenceTargets = (pathToContainment: string, schemaId: string) =>
+  (rootData: Object) => {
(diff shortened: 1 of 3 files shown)


test/schema.service.test.ts
```diff
@@ -618,222 +725,99 @@ test('containment properties delete when array defined', t => {
...
+  const data = {
+    classes: [{id: 'c1'}, {id: 'c2', association: 'c1'}],
+    elements: [{id: 'e1'}]
+  };
+  const targets = property.findReferenceTargets(data);
+  t.is(targets.length, 2);
+  t.is(targets[0], data.classes[0]);
+  t.is(targets[1], data.classes[1]);
+});
+test('reference property find reference targets - target container undefined', t => {
```
edgarmueller: Can we add another test where the expected reference targets are a subset of the available classes?

src/core/schema.service.impl.ts
```diff
@@ -164,18 +189,19 @@ export class SchemaServiceImpl implements SchemaService {
         }
         const href: string = link.href;
         const variableWrapped = href.match(/\{.*\}/)[0];
-        const pathToContainment = href.split(/\{.*\}/)[0];
+        const pathToContainment = href.split(/\/\{.*\}/)[0];
         const variable = variableWrapped.substring(1, variableWrapped.length - 1);
+        const findTargets = findReferenceTargets(pathToContainment, targetSchema.id);
```
edgarmueller: Rename to foundTargets

src/core/schema.service.impl.ts
```diff
@@ -70,36 +70,57 @@ const deleteFromArray = (key: string) => (data: object) => (valueToDelete: objec
...
+    if (_.isEmpty(data) || _.isEmpty(identifyingProperty)) {
+      return null;
+    }
+    // get all objects that could be referenced.
+    const candidates = findTargets(rootData);
+
+    if (schema.properties[variable].type === 'array') {
+      const ids: object[] = data[variable];
+      const resultList = candidates.filter(value => {
+        return ids.indexOf(value[identifyingProperty]) > -1;
```
edgarmueller: Remove curly braces and return keyword.

src/core/schema.service.impl.ts
```diff
@@ -70,36 +70,57 @@ const deleteFromArray = (key: string) => (data: object) => (valueToDelete: objec
...
+      });
 
-    return pathToObject
+      return resultList;
+    } else {
+      // use identifying property to identify the referenced object
+      const resultList = candidates.filter(value => {
+        return value[identifyingProperty] === data[variable];
```
edgarmueller: Same here.

src/core/schema.service.impl.ts
```diff
@@ -70,36 +70,57 @@ const deleteFromArray = (key: string) => (data: object) => (valueToDelete: objec
...
+const resolveRef = (schema: JsonSchema, findTargets: (rootData: Object) => Object[],
+                    identifyingProperty: string, variable: string) =>
+  (rootData: Object, data: Object) => {
+    if (_.isEmpty(data) || _.isEmpty(identifyingProperty)) {
+      return null;
+    }
+    // get all objects that could be referenced.
+    const candidates = findTargets(rootData);
+
+    if (schema.properties[variable].type === 'array') {
```
edgarmueller: type keyword is not mandatory; handle error case

src/core/schema.service.impl.ts
```diff
@@ -70,36 +70,57 @@ const deleteFromArray = (key: string) => (data: object) => (valueToDelete: objec
 const getArray = (key: string) => (data: object) => {
   return data[key];
 };
-const addReference = (schema: JsonSchema, variable: string, pathToContainment: string) =>
+const addReference = (schema: JsonSchema, identifyingProperty: string, variable: string) =>
```
edgarmueller: `variable` is a very generic name. Can we rename this, e.g. to `prop`, `propName` or simlar?

src/core/schema.service.impl.ts
```diff
@@ -70,36 +70,57 @@ const deleteFromArray = (key: string) => (data: object) => (valueToDelete: objec
...
+      // use identifying property to identify the referenced object
+      const resultList = candidates.filter(value => {
+        return value[identifyingProperty] === data[variable];
+      });
+
+      if (_.isEmpty(resultList)) {
+        return null;
+      }
+      if (resultList.length > 1) {
+        throw Error('There was more than one possible reference target with value \'' +
```
edgarmueller: Use template string to avoid escaping

src/core/schema.service.impl.ts
lucas-koehler: I think the `type` keyword has to stay: The test cases fail on removal because the whole schema for the variable is compared to `array` which will always fail.

lucas-koehler:
Added another test case and applied all remarks except for removing the `type` keyword because this lead to failing test cases.

src/core/schema.service.impl.ts
edgarmueller: Yes, I know, my comment was about that we should check that the type keyword is present and otherwise emit an error, otherwise this could lead to follow-up errors.

lucas-koehler: Ah okay, I'll add that.

src/core/schema.service.impl.ts
lucas-koehler: `findTargets` is actually a function that takes the root data as a parameter. So I think this name makes more sense.

PR #627, opened 2017-08-09
Make tree master detail dialog styleable by adding css classes

lucas-koehler:
- Added style classes to dialog and its child elements
- Adapted tree renderer test cases to changes

### DIFF
--- a/src/renderers/additional/tree-renderer.ts
+++ b/src/renderers/additional/tree-renderer.ts
@@ -139,15 +139,20 @@ export class TreeMasterDetailRenderer extends Renderer implements DataChangeList
 
     this.appendChild(div);
     this.dialog = document.createElement('dialog');
+    this.dialog.classList.add('jsf-treeMasterDetail-dialog');
     const title = document.createElement('label');
     title.innerText = 'Select the Item to create:';
+    title.classList.add('jsf-treeMasterDetail-dialog-title');
     this.dialog.appendChild(title);
     const dialogContent = document.createElement('div');
     dialogContent.classList.add('content');
+    dialogContent.classList.add('jsf-treeMasterDetail-dialog-content');
     this.dialog.appendChild(dialogContent);
     const dialogClose = document.createElement('button');
     dialogClose.innerText = 'Close';
     dialogClose.onclick = () => this.dialog.close();
+    dialogClose.classList.add('btn');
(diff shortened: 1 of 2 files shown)


src/renderers/additional/tree-renderer.ts
```diff
@@ -139,15 +139,20 @@ export class TreeMasterDetailRenderer extends Renderer implements DataChangeList
...
+    title.classList.add('jsf-treeMasterDetail-dialog-title');
     this.dialog.appendChild(title);
     const dialogContent = document.createElement('div');
     dialogContent.classList.add('content');
+    dialogContent.classList.add('jsf-treeMasterDetail-dialog-content');
     this.dialog.appendChild(dialogContent);
     const dialogClose = document.createElement('button');
     dialogClose.innerText = 'Close';
     dialogClose.onclick = () => this.dialog.close();
+    dialogClose.classList.add('btn');
```
edgarmueller: Why `btn`? Shouldn't this be `button`?

lucas-koehler: This makes it compatible to materialize. However, that should be too specific for the general tree renderer. Should I change leave it, change it to `button` or remove it completely? I think having a general button class makes sense.

edgarmueller: No, it's fine. It's just that we have to aggre on a style naming convention. I'd go with `button` in this case as and our tests already make use this.

src/renderers/additional/tree-renderer.ts
```diff
@@ -139,15 +139,20 @@ export class TreeMasterDetailRenderer extends Renderer implements DataChangeList
...
+    title.classList.add('jsf-treeMasterDetail-dialog-title');
     this.dialog.appendChild(title);
     const dialogContent = document.createElement('div');
     dialogContent.classList.add('content');
+    dialogContent.classList.add('jsf-treeMasterDetail-dialog-content');
     this.dialog.appendChild(dialogContent);
     const dialogClose = document.createElement('button');
     dialogClose.innerText = 'Close';
     dialogClose.onclick = () => this.dialog.close();
+    dialogClose.classList.add('button');
```
edgarmueller: Sorry, I've missed that the styling registry isn't utilized here. Use `JsonForms.stylingRegistry.addStyle` to apply the actual styles for the relevant HTML elements.

edgarmueller:
Thank you!

PR #631, opened 2017-08-16
SchemaService - improve resolve reference function and some naming

lucas-koehler:
- In the resolveRef function: check that the linking property's schema has a type
- Added test case for this
- Renamed findTargetReferences to getFindReferenceTargetsFunction to make clear that it returns a function


edgarmueller:
Thanks

PR #633, opened 2017-08-17
Abstract Reference Renderer

lucas-koehler:
Implemented an abstract reference renderer that can be implemented by clients to renderer references to other objects in the same root data object.

- implemented abstract renderer
- the abstract renderer provides various overwritable methods to customize its behavior, e.g. like defining the label of the default option
- added global rootData variable to the JsonForms class
- implemented test cases for the abstract renderer

### DIFF
--- a/src/renderers/controls/reference.control.ts
+++ b/src/renderers/controls/reference.control.ts
@@ -33,11 +33,15 @@ export abstract class ReferenceControl extends BaseControl<HTMLSelectElement> {
   }
 
   /**
-   * Overwrite to provide the root data object which contains the possible reference targets.
+   * Returns the root data object that contains the possible reference targets.
+   * Might be overwritten to provide the root data object in another way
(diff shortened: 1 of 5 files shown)


src/renderers/controls/reference.control.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  protected getDefaultOptionLabel(): string {
+    return 'Choose Reference Target...';
+  }
+
+  /**
+   * Adds all possible reference targets as options to the control's combo box.
+   */
+  protected addOptions(input) {
+    const eReferenceSchema = this.dataSchema;
```
edgarmueller: Names are ecore specific

edgarmueller: I know :) I've approved and merge it nevertheless since it's not critical IMHO.

PR #636, opened 2017-09-05
Only generate an id if the identifying property is empty

lucas-koehler:
ID was always regenerated on add. This regenerated the id after drag and drop even if it was already present.


eneufeld:
Is there a test case for this newly added case?

lucas-koehler:
Added a test case for the new case

PR #674, opened 2017-11-16
Add time widget as regular and materialized control

1903Daniel:
Add a time widget

### DIFF
--- a/example/templates/arrays.ts
+++ b/example/templates/arrays.ts
@@ -33,6 +33,10 @@ const schema = {
               }
             }
           }
+        },
+        'time': {
+          'type': 'string',
+          'pattern': '^([0-1][0-9]|2[0-3]):[0-5][0-9]$'
         }
       },
       'required': ['occupation', 'nationality']
@@ -48,6 +52,12 @@ const uischema = {
       'options': {
         'submit': true
       }
+    },
+    {
+      'type': 'Control',
+      'scope': {
+        '$ref': '#/properties/time'
+      },
     }
   ]
 };
@@ -71,7 +81,8 @@ const data = {
       'date': new Date().toISOString().substr(0, 10),
       'message': 'Get ready for booohay'
     }
-  ]
+  ],
+  'time': '13:37'
 };
 registerExamples([
   {name: 'array', label: 'Array', data: data, schema: schema, uiSchema: uischema},
--- a/example/templates/arrays.ts
+++ b/example/templates/arrays.ts
@@ -33,10 +33,6 @@ const schema = {
               }
             }
           }
-        },
-        'time': {
-          'type': 'string',
-          'format': 'time'
         }
       },
       'required': ['occupation', 'nationality']
@@ -52,12 +48,6 @@ const uischema = {
       'options': {
         'submit': true
       }
-    },
-    {
-      'type': 'Control',
-      'scope': {
-        '$ref': '#/properties/time'
-      },
     }
   ]
 };
@@ -81,8 +71,7 @@ const data = {
       'date': new Date().toISOString().substr(0, 10),
       'message': 'Get ready for booohay'
     }
-  ],
-  'time': '13:37'
+  ]
 };
 registerExamples([
   {name: 'array', label: 'Array', data: data, schema: schema, uiSchema: uischema},
--- a/example/templates/arrays.ts
+++ b/example/templates/arrays.ts
@@ -36,7 +36,7 @@ const schema = {
         },
         'time': {
           'type': 'string',
-          'pattern': '^([0-1][0-9]|2[0-3]):[0-5][0-9]$'
+          'format': 'time'
         }
       },
       'required': ['occupation', 'nationality']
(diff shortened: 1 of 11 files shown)


example/templates/arrays.ts
```diff
@@ -33,6 +33,10 @@ const schema = {
               }
             }
           }
+        },
+        'time': {
+          'type': 'string',
+          'pattern': '^([0-1][0-9]|2[0-3]):[0-5][0-9]$'
```
edgarmueller: I'd prefer to use a format `time` instead of a pattern, even if it's not in the standard, since we use `date` currently anyway (actually there's only `date-time` available). Using a pattern can become tedious.

src/core/testers.ts
```diff
@@ -99,6 +99,21 @@ export const formatIs = (expectedFormat: string): Tester => schemaMatches(schema
...
 
+/**
+ * Only applicable for Controls.
+ *
+ * This function checks whether the given UI schema is of type Control
+ * and if so, resolves the sub-schema referenced by the control and checks
+ * whether the pattern of the sub-schema matches the expected one.
+ *
+ * @param {string} expectedPattern the expected pattern of the resolved sub-schema
+ */
```
edgarmueller: Also I wouldn't use a pattern in the tester, we can leave this in here as a tester.

src/renderers/controls/time.control.tsx
```diff
@@ -0,0 +1,57 @@
+import { JSX } from '../JSX';
+import * as _ from 'lodash';
+import { and, patternIs, RankedTester, rankWith, uiTypeIs } from '../../core/testers';
+import { Control, ControlProps, ControlState } from './Control';
+import {
+    formatErrorMessage,
+    mapStateToControlProps,
+    registerStartupRenderer
+} from '../renderer.util';
+import { connect, Event } from '../../common/binding';
```
edgarmueller: Cool that you've already rebased this corretly

src/renderers/controls/time.control.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Default tester for time controls.
+ * @type {RankedTester}
+ */
+export const timeControlTester: RankedTester = rankWith(2, and(
+    uiTypeIs('Control'),
+    patternIs('^([0-1][0-9]|2[0-3]):[0-5][0-9]$')
```
edgarmueller: See first comment about using patterns in the tester

edgarmueller:
Looks good so far, we'll merge this once we have taken care of the CLA.

1903Daniel:
@eneufeld implemented the requested changes, please check

src/renderers/controls/time.control.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    render() {
+        const { classNames, id, visible, enabled, errors, label, uischema } = this.props;
+
+        const isValid = errors.length === 0;
+        const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+
+        return (
+            <div className={classNames.wrapper}>
+                <label htmlFor={id} className={classNames.label} data-error={errors}>
```
eneufeld: Sorry, I just now saw this. As far as I know the data-error is attribute is not defined for a html label, and is a materialize only attribute,

src/renderers/materialized/materialized.time.control.tsx
```diff
@@ -0,0 +1,50 @@
...
+      canceltext: 'Cancel', // Text for cancel-button
+      autoclose: false, // automatic close timepicker
+      ampmclickable: true // make AM PM clickable
+    });
+  }
+
+  render() {
+      const { classNames, id, visible, enabled, errors, label, uischema } = this.props;
+      classNames.input += ' timepicker';
+      const fixedId = '\'' + id + '\'';   // because label[for=#/properties/time] is erroneous
```
eneufeld: I suggest to fix the id globally then, as it will be broken/invalid for all renderers then.

src/renderers/controls/time.control.tsx
1903Daniel: you may be right, but all the other non-material controls have it as well. Shall I remove it from every non-material control, then?

eneufeld: no, just from the ones you touch

eneufeld: @edgarmueller or should we do a dedicated issue out of this?

src/renderers/materialized/materialized.time.control.tsx
edgarmueller: I agree, but maybe not in this PR, but as a follow-up? We have an [issue](https://github.com/eclipsesource/jsonforms/issues/691) to track this already.

src/renderers/controls/time.control.tsx
edgarmueller: Yes, I've done so, see [here](https://github.com/eclipsesource/jsonforms/issues/707).

DanielKutasi: @edgarmueller @eneufeld Ok, thanks for #707. So I won't touch it here to avoid conflicts with the solution of #707

PR #675, opened 2017-11-17
Use field length

1903Daniel:
Fixes #


1903Daniel:
@edgarmueller done from my point of view. Please review and request changes or approve

src/renderers/renderer.util.tsx
```diff
@@ -155,5 +158,6 @@ export const mapStateToControlProps = (state, ownProps) => {
     enabled,
     id,
     path,
+    maxLength,
```
edgarmueller: This is only relevant to input controls (in contrast to say, `enabled`), so I'd try to avoid having an additional property here. Instead, the controls which are able to handle a `maxLength` property should read it within their respective `render` method.

edgarmueller:
Looks good, my only concern is having the `maxLength` property being set in `mapStateToControlProps`, see comment.

Side note: I've also added @eneufeld as a Reviewer who also works on JSON Forms and whom you'll get to know the next meeting.

src/renderers/materialized/materialized.text.control.tsx
```diff
@@ -21,6 +21,8 @@ export class MaterializedTextControl extends Control<ControlProps, ControlState>
                hidden={!visible}
                disabled={!enabled}
                autoFocus={uischema.options && uischema.options.focus}
+               maxlength={maxLength}
+               style={'width :' + maxLength * 15 + 'px;'}
```
eneufeld: Use a calculation for the font size like shown here: [JavaScript Font Size](https://stackoverflow.com/questions/15195209/how-to-get-font-size-in-html)

eneufeld: What will happen if you add something like `body { font-size: 10.0vh; }` in the `example.css`.
Will the constant width of 15px still work?

1903Daniel: I get it now, thanks 😄

1903Daniel:
@eneufeld implemented requested changes, please check

src/renderers/controls/text.control.tsx
```diff
@@ -36,6 +40,9 @@ export class TextControl extends Control<ControlProps, ControlState> {
                hidden={!visible}
                disabled={!enabled}
                autoFocus={uischema.options && uischema.options.focus}
+               maxlength={uischema.options && uischema.options.restrict ? maxLength : undefined}
+               size={uischema.options && uischema.options.trim ? maxLength : undefined}
+               style={uischema.options && uischema.options.trim ? 'width: initial' : ''}
```
eneufeld: why do we need a style attribute in the case of a normal control?

src/renderers/materialized/materialized.text.control.tsx
```diff
@@ -1,17 +1,34 @@
...
+declare let $;
 export class MaterializedTextControl extends Control<ControlProps, ControlState> {
 
+  componentDidMount() {
+    const { id, uischema, schema } = this.props;
+    const controlElement = uischema as ControlElement;
+    const maxLength = resolveSchema(schema, controlElement.scope.$ref).maxLength;
+    if (uischema.options && uischema.options.trim && maxLength !== undefined) {
+      const fontSize = parseFloat($('[id="' + id + '"]').css('font-size'));
+      $('[id="' + id + '"]').css('width', (maxLength * 15) * (fontSize / 14.5) + 'px');
```
eneufeld: Can you explain the constants?

sdirix:
@1903Daniel there is a problem with the CLA check for this pull request. You seem to have mixed multiple email addresses and Github accounts.
Our CLA tool checks all email addresses involved in the commits. All Github accounts corresponding to these email addresses must have signed the CLA. So you should either refactor the commits to use the email address of the Github account @1903Daniel which already has a signed CLA or you sign the CLA also with your @DanielKutasi account. For this you need to make sure to sign out of Github (and the CLA assistant) and login with your other account to sign again.

1903Daniel:
@sdirix thank you for notifying me 😄  I opted to also signing the CLA with @DanielKutasi

src/renderers/controls/text.control.tsx
1903Daniel: You are right, I confused the two. When testing, I temporarily removed the withIncreasedRank(...) from materialize to display the normal control. But then I forgot to select Style "None" from Dropdown menu.

I will fix that, but my fix needs more than removing just the style. In the example, the input field stretches over the entire width of the screen (or its div container, respectively), so somehow the size attribute gets overwritten.

src/renderers/materialized/materialized.text.control.tsx
1903Daniel: First part: W is the widest letter of the latin alphabet. It takes up 15px, so we multiply by the maximum number of characters (maxLength).

Second part: The initial font size is 14.5px, so increasing the (calculated) font-size property means we have to multiply the first part by the ratio of which the font size has increased compared to the initial 14.5px

src/renderers/controls/text.control.tsx
```diff
@@ -23,7 +23,8 @@ export class TextControl extends Control<ControlProps, ControlState> {
     const isValid = errors.length === 0;
     const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
     const controlElement = uischema as ControlElement;
-    const maxLength = resolveSchema(schema, controlElement.scope.$ref).maxLength;
+    const resolvedSchema = resolveSchema(schema, controlElement.scope.$ref);
+    const maxLength = resolvedSchema ? resolvedSchema.maxLength : undefined;
```
1903Daniel: @eneufeld @edgarmueller please don't merge yet because I found a bug in here I have to fix first.
For nested properties like the one below, the code doesn't work, because `resolveSchema` descends down to the first `properties` object, but not the nested ones within that.

`"address": {`
&nbsp;&nbsp;`"type": "object",`
&nbsp;&nbsp;`"properties": {`
&nbsp;&nbsp;&nbsp;&nbsp;`"streetnumber": {`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`"type": "string",`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`"maxLength": 4`
&nbsp;&nbsp;&nbsp;&nbsp;`}`
&nbsp;&nbsp;`}`
`},`

1903Daniel:
@eneufeld @edgarmueller I implemented the changes. Please check again

src/renderers/materialized/materialized.text.control.tsx
```diff
@@ -1,17 +1,46 @@
...
+         displayed with a font-size of 14.5px.
+         For the calculation of the input's width, the maximum number of allowed characters
+         (maxLength) has to be multiplied by 15px, as an input text consisting of W only
+         can be assumed to be the widest input text possible.
+         Furthermore, the result of this has to be multiplied by the ratio with which the
+         font-size has increased (decreased) compared to the base font-size of 14.5px
+         that applies to the mentioned width of 15px for a W, in order to enlarge (shrink)
+         the input width according to the changed font-size.
+       */
+      input.css('width', `${(maxLength * 15) * (fontSize / 14.5)}px`);
```
eneufeld: didn't we agree to move the constants to two constant fields and add a jsdoc to each constant?

1903Daniel: @eneufeld sorry, I forgot about that. Now it's changed

PR #676, opened 2017-11-19
Inferno issues

### LINKED ISSUE #671: Inferno migration issues
Open issues:
- [x] DnD support missing
- [x] Remove remaining global state, e.g. Styling registry or JsonForms global object, should to be removed
- [x] Schema service must be ported (does not work with redux due to mutable operations)
- [x] Tree/Categories styling for Material does not exist (hence appears broken)
- [x] Raise test coverage once again
- [x] Check for React compatibility

eneufeld:
Partly fixes #671 

This contributes multiple small fixes for the inferno migration.

### DIFF
--- a/src/renderers/materialized/materialized.date.control.tsx
+++ b/src/renderers/materialized/materialized.date.control.tsx
@@ -4,22 +4,32 @@ import { Control, ControlProps, ControlState } from '../controls/Control';
 import { mapStateToControlProps, registerStartupRenderer } from '../renderer.util';
 import { dateControlTester } from '../controls/date.control';
 import { connect, Event } from '../../common/binding';
-
+declare let $;
 export class DateControl extends Control<ControlProps, ControlState> {
+
+  componentDidMount() {
+    $('.datepicker').pickadate({
+      format: 'yyyy-mm-dd', // we should actually use some localized format
+      selectMonths: true, // Creates a dropdown to control month
+      selectYears: 15, // Creates a dropdown of 15 years to control year,
+      today: 'Today',
+      clear: 'Clear',
+      close: 'Ok',
+      closeOnSelect: false // Close upon selecting a date,
+    });
+  }
 
   render() {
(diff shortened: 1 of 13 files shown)


src/renderers/materialized/materialized.date.control.tsx
```diff
@@ -6,19 +6,20 @@ import { dateControlTester } from '../controls/date.control';
...
           {label}
         </label>
-        <input type='date'
-               value={this.state.value}
-               onChange={(ev: Event<HTMLInputElement>) =>
-                 this.handleChange(new Date(ev.currentTarget.value).toISOString().substr(0, 10))
+        <input type='text'
+               value={this.formatter.format(new Date(this.state.value))}
+               onChange={ev =>
+                 this.handleChange(ev.target.previousSibling.previousSibling.value)
```
edgarmueller: The date control doesn't seem to work for me (do I need to change any of the files):
* It looks like a regular text field, no date-specific controls
* Can you explain what `ev.target.previousSibling.previousSibling.value` does? Also adding a comment would be good
* `onChange` throws an error in any case
```
Uncaught TypeError: Cannot read property 'value' of null
    at onChange (materialized.date.control.tsx:22)
    at HTMLInputElement.wrappedOnChange (index.js:349)

eneufeld: Interesting, that I don't get hit by this issue.
Can you try to increase the timeout in line 50 to `1000` ms?
The `ev.target.previousSibling.previousSibling.value` assumes the dom structure which materialize/piackadate create when providing the date control see [Materialize DatePicker](http://materializecss.com/forms.html#date-picker).

On the long run we need a possibility to get notified when the jsonforms element finishes rendering so that we can run additional code, like the pickadate stuff.

eneufeld: as I learned there is one: `componentDidMount()` just overwrite this

PR #677, opened 2017-11-19
 Add datetime widget

1903Daniel:
Fixes #

### DIFF
--- a/example/index.ts
+++ b/example/index.ts
@@ -18,6 +18,7 @@ import './templates/masterdetail';
 import './templates/resolve';
 import './templates/uischema-registry';
 import './templates/ecore';
+import './templates/dates';
 
 window.onload = ev => {
   const selectExampleElement = createExampleSelection();
--- a/src/renderers/controls/datetime.control.tsx
+++ b/src/renderers/controls/datetime.control.tsx
@@ -1,0 +1,83 @@
+import { JSX } from '../JSX';
+import * as _ from 'lodash';
+import { and, formatIs, RankedTester, rankWith, uiTypeIs } from '../../core/testers';
+import { Control, ControlProps, ControlState } from './Control';
+import {
+  assembleDateTime,
+  formatErrorMessage,
+  mapStateToControlProps,
+  registerStartupRenderer
+} from '../renderer.util';
+import { connect, Event } from '../../common/binding';
+
+/**
+ * Default tester for datetime controls.
+ * @type {RankedTester}
+ */
+export const datetimeControlTester: RankedTester = rankWith(2, and(
+  uiTypeIs('Control'),
+  formatIs('date-time')
+));
+
+export class DateTimeControl extends Control<ControlProps, ControlState> {
+
+  render() {
+    const { classNames, id, visible, enabled, errors, label, uischema } = this.props;
+    const isValid = errors.length === 0;
+    const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+    const value = this.state.value.toString();
+    let date = value.substr(0, 10);
+    let time = value.substr(11, 5);
+
+    return (
+      <div className={classNames.wrapper} style={'display: block;'}>
+        <label htmlFor={id} className={classNames.label} data-error={errors}>
+          {label}
+        </label>
+        <input type='date'
+               value={date}
+               onChange={(ev: Event<HTMLInputElement>) => {
+                 date = ev.currentTarget.value;
+                 if (date && time) {
+                   this.handleChange(
+                     assembleDateTime(date, time)
+                   );
+                 }
+                }
+               }
+               className={classNames.input}
+               id={id + '-date'}
+               hidden={!visible}
+               disabled={!enabled}
+               autoFocus={uischema.options && uischema.options.focus}
+               style={'width: 49%;'}
+        />
+        <input type='time'
+               value={time}
+               onChange={(ev: Event<HTMLInputElement>) => {
+                 time = ev.currentTarget.value;
+                 if (date && time) {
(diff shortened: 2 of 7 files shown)


1903Daniel:
@edgarmueller this should work now 😄 Please review & request changes or approve

example/index.ts
```diff
@@ -5,7 +5,6 @@ import './templates/arrays';
 import './templates/day2';
 import './templates/day4';
 import './templates/day5';
-import './templates/day6';
```
eneufeld: day6 should already be fixed, so please revert this removal.

example/templates/arrays.ts
```diff
@@ -33,6 +33,10 @@ const schema = {
               }
             }
           }
+        },
```
eneufeld: please create a dedicated example for all the date/date-time/time renderer.

src/renderers/controls/datetime.control.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export class DateTimeControl extends Control<ControlProps, ControlState> {
+
+    render() {
+        const { classNames, id, visible, enabled, errors, label } = this.props;
+
+        const isValid = errors.length === 0;
+        const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+
+        const localISOTime = (datestring: string): string => {
```
eneufeld: we should make the `localISOTime` method a util, as you need it in the examples and the two renderers.

src/renderers/materialized/materialized.datetime.control.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+            return (new Date(date.valueOf() - tzoffset)).toISOString();
+        };
+
+        return (
+            <div className={classNames.wrapper}>
+                <label htmlFor={id} className={classNames.label} data-error={errors}>
+                    {label}
+                </label>
+                <input type='datetime-local'
```
eneufeld: Materialize provides its own implementation of a date and a time widget, check out their docs for [date](http://materializecss.com/forms.html#date-picker) and [time](http://materializecss.com/forms.html#time-picker).

1903Daniel: @eneufeld, yes, but date and time are separate widgets in Materialize. There is a way [1] to include both in one input field, but it seems quite complex. So we have two options here, and I'm asking @edgarmueller & @JonasHelming to please decide which one to choose:
1.) we'll implement this, even though it might take a bit and in the end, we may accomplish a bit fewer features in total, as we will have a bit less time for them
2.) we drop the two datetime controls altogether, and we will just have the date control which is already present, and the time control I am working on at the other PR.

[1]
https://github.com/Dogfalo/materialize/issues/610
the last comment says:
That's tricky.. 😃

Material design pickers seem to launch one after the other if date/time is used at the same time. I guess I would start the datepicker, wait for it to finish - there's an onClose callback which is not mentioned in Materialize docs but in pickadate docs. And start the timepicker from there.

The timepicker is actually a modified clockpicker which has a show method.. Then combine the results.

eneufeld: Yes this is something @edgarmueller  and especially @JonasHelming  have to decide.
I see a third option though. Just place both control next to each other, so that you can set date and time separately.
The programmatic integration has a better usability of course. I could imagine, that the result would be even better than what the browsers support (see description in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) .

edgarmueller: @1903Daniel I'd go with the 3rd option that @eneufeld suggested.

1903Daniel: @edgarmueller will do, thanks

src/renderers/controls/datetime.control.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    render() {
+        const { classNames, id, visible, enabled, errors, label, uischema } = this.props;
+
+        const isValid = errors.length === 0;
+        const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+
+        return (
+            <div className={classNames.wrapper}>
+                <label htmlFor={id} className={classNames.label} data-error={errors}>
```
eneufeld: `data-error` is a materialize attribute as far as I know

1903Daniel:
@eneufeld updated. It is still buggy, though, so work in progress. Especially the css formatting (red/green validation bar) is not working yet.

src/renderers/materialized/materialized.datetime.control.tsx
```diff
@@ -0,0 +1,95 @@
...
+import {
+  assembleDateTime, mapStateToControlProps, registerStartupRenderer
+} from '../renderer.util';
+import { datetimeControlTester } from '../controls/datetime.control';
+import { connect } from '../../common/binding';
+declare let $;
+export class DateTimeControl extends Control<ControlProps, ControlState> {
+
+  componentDidMount() {
+    $('.datepicker').pickadate({
```
eneufeld: We discussed the problems you encounter with date, time and datetime and we would like to try out the use of a react based materialize library like: https://material-ui-next.com/ (http://www.material-ui.com/) .
Could you implement the datetime using this library in version 1.0.0-alpha and probably one of the mentioned picker libraries in the docu and then create a new PR?

edgarmueller: To avoid any confusion, the current version of material-ui at the time of writing is v1.0.0-beta.22.

PR #681, opened 2017-11-22
Required control

AhmetTanakol:
- Required field control move to `mapStateToControlProps` function
- Required field warning is set in control classes

### DIFF
--- a/src/renderers/controls/date.control.tsx
+++ b/src/renderers/controls/date.control.tsx
@@ -21,14 +21,14 @@ export const dateControlTester: RankedTester = rankWith(2, and(
 export class DateControl extends Control<ControlProps, ControlState> {
 
   render() {
-    const { classNames, id, visible, enabled, errors, label, uischema } = this.props;
+    const { classNames, id, visible, enabled, errors, label, uischema, required } = this.props;
     const isValid = errors.length === 0;
     const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
 
     return (
       <div className={classNames.wrapper}>
         <label htmlFor={id} className={classNames.label} data-error={errors}>
-          {label}
+          {required && !this.state.value ? label + '*' : label}
         </label>
         <input type='date'
                value={this.state.value}
--- a/src/renderers/controls/date.control.tsx
+++ b/src/renderers/controls/date.control.tsx
@@ -28,7 +28,7 @@ export class DateControl extends Control<ControlProps, ControlState> {
     return (
       <div className={classNames.wrapper}>
         <label htmlFor={id} className={classNames.label} data-error={errors}>
-          {required && !this.state.value ? label + '*' : label}
+          {required ? label + '*' : label}
         </label>
         <input type='date'
                value={this.state.value}
--- a/src/renderers/controls/date.control.tsx
+++ b/src/renderers/controls/date.control.tsx
@@ -5,7 +5,8 @@ import { Control, ControlProps, ControlState } from './Control';
 import {
   formatErrorMessage,
   mapStateToControlProps,
-  registerStartupRenderer
+  registerStartupRenderer,
+  setLabelField
 } from '../renderer.util';
 import { connect, Event } from '../../common/binding';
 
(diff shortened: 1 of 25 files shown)


edgarmueller:
@eneufeld Ping me, if you have any questions about the requirements.

src/renderers/controls/date.control.tsx
```diff
@@ -21,14 +21,14 @@ export const dateControlTester: RankedTester = rankWith(2, and(
...
-    const { classNames, id, visible, enabled, errors, label, uischema } = this.props;
+    const { classNames, id, visible, enabled, errors, label, uischema, required } = this.props;
     const isValid = errors.length === 0;
     const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
 
     return (
       <div className={classNames.wrapper}>
         <label htmlFor={id} className={classNames.label} data-error={errors}>
-          {label}
+          {required ? label + '*' : label}
```
eneufeld: please extract this to a util to remove code duplication

src/renderers/label.util.ts
```diff
@@ -67,10 +55,6 @@ const isRequired = (schema: JsonSchema, schemaPath: string): boolean => {
  */
 export const getElementLabelObject =
```
eneufeld: This method is now obsolete, please remove it and export getLabelObject instead. You will need to update all the imports of course.

src/renderers/controls/date.control.tsx
eneufeld: a good place would be the renderer.util

AhmetTanakol: should I remove `required` field from `props`

AhmetTanakol: So you want me to create a function in renderer.util, so controllers will call this function with the `required` parameter. Am I correct?

edgarmueller: @AhmetTanakol Yes, e.g. think about what would be necessary if you change the asterisk to a different character: You'd need to change the line in each renderer, which is cumbersome.

AhmetTanakol:
I first pushed from wrong local branch but then I checkout the correct one and I used cherry-pick and updated the correct local branch and pushed again. I hope it is fixed now

src/renderers/label.util.ts
```diff
@@ -26,7 +25,14 @@ const deriveLabel = (controlElement: ControlElement): string => {
...
 
-const getLabelObject = (withLabel: ControlElement): ILabelObject => {
+/**
+ * Return a label object based on the given JSON schema and control element.
+ * @param {JsonSchema} schema the JSON schema that the given control element is referring to
+ * @param {ControlElement} controlElement the UI schema to obtain a label object for
+ * @returns {ILabelObject}
+ */
+
+export const getLabelObject = (withLabel: ControlElement): ILabelObject => {
```
eneufeld: The JavaDoc doesn't fit the method

src/renderers/renderer.util.tsx
```diff
@@ -121,16 +122,34 @@ export const registerStartupRenderer = (tester: RankedTester, renderer: any) =>
...
+    const nextHigherSchemaSegments = pathSegments.slice(0, pathSegments.length - 2);
+    const nextHigherSchemaPath = nextHigherSchemaSegments.join('/');
+    const nextHigherSchema = resolveSchema(schema, nextHigherSchemaPath);
+
+    return nextHigherSchema !== undefined
+        && nextHigherSchema.required !== undefined
+        && nextHigherSchema.required.indexOf(lastSegment) !== -1;
+};
+
+export const setLabelField = (label: string, required: boolean): string => {
```
eneufeld: what does this method set? Doesn't it only return something?

AhmetTanakol: Yes, at this point I am not sure where to define label setting. Or is it ok to use a method that returns the edited version of label in the control? Maybe it has to be set in the control's constructor. Do you have any advice?

eneufeld: I would simply rename the method, to describe what it does. Something like `computeLabel`.

PR #705, opened 2017-11-24
Update error data path setting

AhmetTanakol:
For nested objects validation, data error path is reconfigured

### DIFF
--- a/example/templates/person.ts
+++ b/example/templates/person.ts
@@ -118,7 +118,8 @@ const uischema = {
 const data = {
   name: 'John Doe',
   vegetarian: false,
-  birthDate: '1985-06-02'
+  birthDate: '1985-06-02',
+  personalData: {}
 };
 registerExamples([
   {name: 'person', label: 'Person', data: data, schema: schema, uiSchema: uischema}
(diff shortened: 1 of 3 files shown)


edgarmueller:
@AhmetTanakol Thanks for the PR! Would you mind also adding a test case? You may just use the person example with the `personalData` set to empty an object like in the example and then check whether the validation errors pop up.

AhmetTanakol:
test case added

example/templates/person.ts
```diff
@@ -118,7 +118,8 @@ const uischema = {
 const data = {
   name: 'John Doe',
   vegetarian: false,
-  birthDate: '1985-06-02'
+  birthDate: '1985-06-02',
+  personalData: {}
```
eneufeld: @edgarmueller shouldn't nested objects be created on the fly when needed?

edgarmueller: @eneufeld Depends on how you define 'needed': The object is already created once you touch any of its properties, but one might also argue that we should already create an empty object even in case it's not actually required: if that's what you want, I'd propose to move that discussion to a different issue.

edgarmueller:
Looks good. If @eneufeld has no objects we can merge this.

example/templates/person.ts
eneufeld: I opened an issue to discuss this #708

PR #710, opened 2017-11-28
Fix issue #704: Remove warnings during test execution

### LINKED ISSUE #704: Remove warnings during test execution
There are warnings while the tests are executing. We should check why these pop up and fix them

edgarmueller:
Fixes #704: Remove warnings during tests

* Add keys where necessary
* Add default values as undefined is not legal
* Replace deleteFunction with update action in tree renderer
* Replace categorization render methods with component

### DIFF
--- a/src/json-forms.tsx
+++ b/src/json-forms.tsx
@@ -146,9 +146,10 @@ export class JsonFormsElement extends HTMLElement {
     const uischema = this.uiSchema;
 
     this.store = initJsonFormsStore(this.dataObject, schema, uischema);
+    const storeId = new Date().toISOString();
 
     render(
-      <Provider store={this.store}>
+      <Provider store={this.store} key={`${storeId}-store`}>
         <DispatchRenderer uischema={uischema} schema={schema} />
       </Provider>,
       this
--- a/src/reducers/common.ts
+++ b/src/reducers/common.ts
@@ -38,9 +38,9 @@ export const commonStateReducer = (
           schema: state.schema
         };
       } else {
-        const currData = get(state.data, action.path);
-        const data = action.updater(currData);
-        const newState = set(state.data, action.path, data);
+        const oldData = get(state.data, action.path);
+        const newData = action.updater(oldData);
+        const newState = set(state.data, action.path, newData);
 
         return {
           data: newState,
--- a/src/reducers/common.ts
+++ b/src/reducers/common.ts
@@ -1,3 +1,4 @@
+import * as _ from 'lodash';
 import { get, set } from 'dot-prop-immutable';
 import { INIT, UPDATE_DATA } from '../actions';
 
@@ -38,7 +39,8 @@ export const commonStateReducer = (
         };
       } else {
         const currData = get(state.data, action.path);
-        const newState = set(state.data, action.path, action.updater(currData));
+        const data = action.updater(currData);
+        const newState = set(state.data, action.path, data);
 
         return {
           data: newState,
--- a/src/renderers/additional/categorization-renderer.tsx
+++ b/src/renderers/additional/categorization-renderer.tsx
@@ -1,12 +1,89 @@
 import { JSX } from '../JSX';
 import * as _ from 'lodash';
 import { JsonForms } from '../../core';
-import { Renderer, RendererProps } from '../../core/renderer';
+import { RendererProps } from '../../core/renderer';
 import { and, RankedTester, rankWith, uiTypeIs } from '../../core/testers';
 import { Categorization, Category } from '../../models/uischema';
 import { mapStateToLayoutProps, registerStartupRenderer } from '../renderer.util';
 import DispatchRenderer from '../dispatch-renderer';
 import { Component, connect } from '../../common/binding';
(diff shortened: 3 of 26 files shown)


src/json-forms.tsx
```diff
@@ -146,9 +146,10 @@ export class JsonFormsElement extends HTMLElement {
     const uischema = this.uiSchema;
 
     this.store = initJsonFormsStore(this.dataObject, schema, uischema);
+    const storeId = new Date().toISOString();
 
     render(
-      <Provider store={this.store}>
+      <Provider store={this.store} key={`${storeId}-store`}>
```
eneufeld: Is this kind of a UUID?

src/reducers/common.ts
```diff
@@ -38,7 +39,8 @@ export const commonStateReducer = (
         };
       } else {
         const currData = get(state.data, action.path);
-        const newState = set(state.data, action.path, action.updater(currData));
+        const data = action.updater(currData);
```
eneufeld: shouldn't the variables be named:
`currData` -> `oldData`
`data` -> `newData`
?

src/renderers/additional/categorization-renderer.tsx
```diff
@@ -1,12 +1,89 @@
...
+export interface SingleCategoryProps {
+  category: Category;
+  schema: JsonSchema;
+  path: string;
+}
+
+export const SingleCategory = ({ category, schema, path }: SingleCategoryProps) => {
+  // TODO: add selected style
+  if (category.elements === undefined) {
+    return (<div id='categorization.detail'/>);
```
eneufeld: do we need two cases? isn't the second case handling the undefined case?

src/renderers/additional/categorization-renderer.tsx
```diff
@@ -1,12 +1,89 @@
...
 import { Categorization, Category } from '../../models/uischema';
 import { mapStateToLayoutProps, registerStartupRenderer } from '../renderer.util';
 import DispatchRenderer from '../dispatch-renderer';
 import { Component, connect } from '../../common/binding';
+import {JsonSchema} from "../../models/jsonSchema";
+
+const getCategoryClassName = (category: Category, selectedCategory: Category): string =>
+  selectedCategory === category ? 'selected' : '';
+
+export interface CategorizationProps {
```
eneufeld: why do we need to export all those props and inner components?

src/renderers/controls/boolean.control.tsx
```diff
@@ -30,7 +30,7 @@ export class BooleanControl extends Control<ControlProps, ControlState> {
           {label}
         </label>
         <input type='checkbox'
-               checked={this.state.value}
+               checked={this.state.value || ''}
```
eneufeld: does checked accept a string? should it be `false` ?

src/renderers/controls/date.control.tsx
```diff
@@ -24,14 +24,15 @@ export class DateControl extends Control<ControlProps, ControlState> {
...
+    const now = new Date().toISOString().substr(0, 10);
 
     return (
       <div className={classNames.wrapper}>
         <label htmlFor={id} className={classNames.label} data-error={errors}>
           {label}
         </label>
         <input type='date'
-               value={this.state.value}
+               value={this.state.value || ''}
```
eneufeld: this should be `now` not an empty string, doesn't it?

src/renderers/controls/enum.control.tsx
```diff
@@ -42,27 +52,26 @@ export class EnumControl extends Control<ControlProps, ControlState> {
           className={classNames.input}
           hidden={!visible}
           disabled={!enabled}
-          value={this.state.value}
+          value={this.state.value || ''}
```
eneufeld: doesn't the value of a `select` control correspond to the key? so shouldn't the alternative be `'empty'` rather than an empty string?

src/renderers/controls/integer.control.tsx
```diff
@@ -32,7 +32,7 @@ export class IntegerControl extends Control<ControlProps, ControlState> {
         </label>
         <input type='number'
                step='1'
-               value={this.state.value}
+               value={this.state.value || ''}
```
eneufeld: does an input of type number handle an empty string correctly?

test/renderers/textarea.control.test.tsx
```diff
@@ -274,7 +274,7 @@ test('update with undefined value', t => {
   const textArea = findRenderedDOMElementWithTag(tree, 'textarea') as HTMLTextAreaElement;
   store.dispatch(update('name', () => undefined));
   // keep value
```
eneufeld: this comment is wrong now, as we don't keep the value

src/json-forms.tsx
edgarmueller: Yes, kinda.

src/renderers/additional/categorization-renderer.tsx
edgarmueller: As discussed, we'll keep these in case anyone would create custom renderers for categories.

PR #715, opened 2017-11-30
added slider renderer

AhmetTanakol:


example/templates/person.ts
```diff
@@ -31,6 +31,11 @@ const schema = {
     },
     'occupation': {
       'type': 'string'
+    },
+    'slider': {
```
eneufeld: can't we name this property more fitting to a person, or simply use the age property?

AhmetTanakol: I noticed that when I use range-slider for height for instance, tooltip (icon) doesn't move when you change the data. The reason is there is no left padding in the style of the tooltip. I didn't have this problem when I use it as a separate field in a separate layout in the same example. What might be the problem here?

src/renderers/controls/slider.control.tsx
```diff
@@ -0,0 +1,61 @@
...
+import {
+    formatErrorMessage,
+    mapStateToControlProps,
+    registerStartupRenderer
+} from '../renderer.util';
+import { connect, Event } from '../../common/binding';
+
+export const sliderControlTester: RankedTester = rankWith(4, and(
+    uiTypeIs('Control'),
+    schemaTypeIs('number'),
```
edgarmueller: This might also be applicable for integers, doesn't it? Also, one might argue, but I'd only use a slider if there is a `minimum` and `maximum` configured (and you already do in the example) since sensible default rarely can be provided.

src/renderers/controls/slider.control.tsx
```diff
@@ -0,0 +1,61 @@
...
+        const isValid = errors.length === 0;
+        const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+
+        return (
+            <div className={classNames.wrapper}>
+                <label htmlFor={id} className={classNames.label} data-error={errors}>
+                    {label}
+                </label>
+                <input type='range'
+                       min='0'
```
edgarmueller: Why 0 and 100? Why not using `minimum` and `maximum`? 0 and 100 seem like arbitrary defaults.

src/renderers/controls/slider.control.tsx
eneufeld: do we have an `or`? I agree that we should also support integers.

src/renderers/controls/slider.control.tsx
eneufeld: yes, please use the minimum and maximum values from the schema

src/renderers/controls/slider.control.tsx
AhmetTanakol: We don't have an `or` right now. I think we need a different tester to check this

example/templates/person.ts
AhmetTanakol: Ok, solved this issue

src/renderers/controls/slider.control.tsx
AhmetTanakol: I created another function called `or` which is basically similar to `and`. I changed the logical expression `testers.reduce((acc, tester) => acc || tester(uischema, schema), true);` . Is it ok?

edgarmueller: Yes should be fine if you place it within `testers.ts`.

eneufeld: should the initial value be `false`?

edgarmueller: Uhm, yes, missed that.

AhmetTanakol: I see, the reason is if the initial value start with true, even tester returns false (neither number nor integer), the result of `or` will be always true which is not correct. Am I right?

AhmetTanakol:
updated

src/renderers/controls/slider.control.tsx
```diff
@@ -0,0 +1,66 @@
...
+    mapStateToControlProps,
+    registerStartupRenderer
+} from '../renderer.util';
+import { connect, Event } from '../../common/binding';
+import { resolveSchema } from '../../path.util';
+import { ControlElement } from '../../models/uischema';
+
+export const sliderControlTester: RankedTester = rankWith(4, and(
+    uiTypeIs('Control'),
+    or(schemaTypeIs('number'), schemaTypeIs('integer')),
```
eneufeld: do we really need an uischema option `slider`?
Can't/Shouldn't we simply check for a `minimum` and `maximum` validation on the schema?

AhmetTanakol: I am not sure how we can differentiate integer/number input field from a range slider if we dont say “I want this to be a slider”

edgarmueller: Yes, I agree, checking for `minimum` and `maximum` suffices.

edgarmueller: @AhmetTanakol The presence of `minimum` and `maximum` is enough to trigger the renderer.

AhmetTanakol: updated slider control tester and added new test cases for maximum and minimum validation

test/renderers/slider.control.test.tsx
```diff
@@ -104,6 +104,78 @@ test('tester with wrong schema type, but sibling has correct one', t => {
   );
 });
 
+test('tester with wrong schema type, missing maximum and minimum fields', t => {
```
eneufeld: the schema type is `correct` not `wrong`

test/renderers/slider.control.test.tsx
```diff
@@ -104,6 +104,78 @@ test('tester with wrong schema type, but sibling has correct one', t => {
...
+            type: 'number'
+          }
+        }
+      }
+    ),
+    -1
+  );
+});
+
+test('tester with wrong schema type, missing maximum', t => {
```
eneufeld: same as before

test/renderers/slider.control.test.tsx
```diff
@@ -104,6 +104,78 @@ test('tester with wrong schema type, but sibling has correct one', t => {
...
+            minimum: 2
+          }
+        }
+      }
+    ),
+    -1
+  );
+});
+
+test('tester with wrong schema type, missing minimum', t => {
```
eneufeld: same as before

example/templates/person.ts
```diff
@@ -10,10 +10,14 @@ const schema = {
       'type': 'object',
       'properties': {
           'age': {
-              'type': 'integer'
+              'type': 'integer',
+              'maximum': 25,
```
eneufeld: I'm not a huge fan of adding the minimum and maximum to the `age` and `height` properties. I would prefer to have either
- a dedicated example for numbers
or 
- add additional properties for the slider control to the person schema

but this is not a blocker in my opinion

AhmetTanakol: yes actually I want to add another property called`drivingSkill`

edgarmueller:
Closing in favor of #732

Issue #688, opened 2017-11-22
Handle multiple focus annotation


edgarmueller:
We need to handle multiple focus options within the UI schema in a consistent way as the behavior across different framework tends to be implementation-specific; we should clarify the exact semantics

PR #716, opened 2017-12-03
Extract Fields from base controls

### LINKED ISSUE #714: Extract input-based controls to common component
All input-based controls actually are more or less the same. We could extract these and re-use them in the renderers, e.g. when we don't need a label.

eneufeld:
Fixes #714

This defines a new Component `Field` which describes the input itself.
It introduces a `FieldDispatcher` which is now used by
- `input.control` The control which replaces the old base controls
- `table-array.control` The control which produces a table for object arrays
Also test are updated.
The Rating Control from the examples was fixed to not produce warnings in the console.
Finally a new example `index.html` was created to allow to test the default renderer set. This also includes a new npm run config `dev-deafault`.


src/actions.ts
```diff
@@ -13,6 +13,8 @@ export const ENABLE = 'ENABLE';
 export const DISABLE = 'DISABLE';
 export const ADD_RENDERER = 'ADD_RENDERER';
 export const REMOVE_RENDERER = 'REMOVE_RENDERER';
+export const ADD_INPUT = 'ADD_INPUT';
```
edgarmueller: Since we agreed on using the term 'Field', should we rename these occurrences (applies to reducers etc. as well)?

src/renderers/fields/dispatch.field.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { UnknownRenderer } from '../unknown.renderer';
+import { RankedTester } from '../../core/testers';
+import { connect } from '../../common/binding';
+import { FieldProps } from './field';
+
+export interface DispatchFieldProps extends FieldProps {
+  inputs?: { tester: RankedTester, input: any }[];
+}
+
+export const DispatchField = (dispatchInputProps: DispatchFieldProps) => {
```
edgarmueller: Naming

test/renderers/boolean.field.test.tsx
```diff
@@ -47,8 +47,8 @@ test.beforeEach(t => {
     }
   };
 });
-
-test('autofocus on first element', t => {
+// TODO: I don't understand this test
```
edgarmueller: See issue #688

test/renderers/boolean.field.test.tsx
```diff
@@ -420,12 +372,12 @@ test('update with undefined ref', t => {
   const input = findRenderedDOMElementWithTag(tree, 'input') as HTMLInputElement;
   t.is(input.checked, true);
 });
-
+// TODO: does it make sense to hide the input?
```
edgarmueller: Probably not, if the wrapping input control already takes care of this. I can't image a use case where we'd want to show the label, but not the field. This would allow us to move the test and remove the hidden prop from the field renderers

test/renderers/input.control.test.tsx
```diff
@@ -0,0 +1,348 @@
...
+  t.is(label.textContent, '');
+
+  const input = findRenderedDOMElementWithTag(tree, 'input') as HTMLInputElement;
+
+  const validation = findRenderedDOMElementWithClass(tree, 'validation');
+  t.is(validation.tagName, 'DIV');
+  t.is((validation as HTMLDivElement).children.length, 0);
+});
+
+test.skip('hide', t => {
```
edgarmueller: Why do we skip this one?

test/renderers/input.control.test.tsx
```diff
@@ -0,0 +1,348 @@
...
+  uischema={t.context.uischema}
+  visible={false}
+    />
+    </Provider>
+);
+  const control = findRenderedDOMElementWithClass(tree, 'control') as HTMLElement;
+  t.true(control.hidden);
+});
+
+test.skip('show by default', t => {
```
edgarmueller: Why skip?

test/renderers/input.control.test.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+test('empty errors by default', t => {
+  const store = initJsonFormsStore(
+    t.context.data,
+    t.context.schema,
+    t.context.uischema
+  );
+  const tree = renderIntoDocument(
+    <Provider store={store}>
+    <InputControl schema={t.context.schema}
```
edgarmueller: Indentation messed up (occurs multiple times; actually I don't care that much, but as we do not yet have set-up a formatter, maybe fix this)

src/renderers/fields/enum.field.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const options = resolveSchema(schema, (uischema as ControlElement).scope.$ref).enum;
+
+    return <select
+        className={className}
+        id={id}
+        hidden={!visible}
+        disabled={!enabled}
+        autoFocus={uischema.options && uischema.options.focus}
+        value={data || ''}
+        onChange={(ev: Event<HTMLSelectElement>) => {
```
edgarmueller: Indentation; you can get rid of the inner curly braces to fix it

src/renderers/fields/field.ts
```diff
@@ -0,0 +1,54 @@
...
+  dispatch: any;
+}
+export interface FieldState {
+  value: any;
+}
+export class Field<P extends FieldProps, S extends FieldState> extends Component<P, S> {
+
+  constructor(props: P) {
+    super(props);
+    this.state = {
```
edgarmueller: I think we can get rid of `state` again as I don't see where we are using it anyway, since we read the `data` property on each render now instead from `this.state.value`. We hence should be able to remove `FieldState` entirely or keep it empty.

src/renderers/fields/field.util.ts
```diff
@@ -0,0 +1,43 @@
...
+import { JsonForms } from '../../core';
+import { isEnabled, isVisible } from '../../core/renderer';
+import { composeWithUi, resolveData } from '../../path.util';
+import { getData, getValidation } from '../../reducers/index';
+import { Component } from '../../common/binding';
+import { errorAt } from '../../reducers/validation';
+import { Field, FieldProps } from './field';
+
+export interface JsonFormsInputConstructable {
+  // TODO: any state?
```
edgarmueller: I don't think so

PR #718, opened 2017-12-04
Remove Array Control and update table control

### LINKED ISSUE #695: Review array-control(s)
The array control(s) don't seem to make use of the `mapStateToControlProps` method. Is there any reasons why this is the case?
Also re-check how the renderer behaves if it is hidden or disabled.

eneufeld:
Fixes #695

This removes the array control and updates the table control with test cases and css.
This is based on #716.


test/renderers/table-array.control.test.tsx
```diff
@@ -106,260 +112,429 @@ test('render two children', t => {
...
+
+   t.is(tableArrayTester(uischema, schema), 10);
+ });
+
+test('hide', t => {
+   const store = initJsonFormsStore(t.context.data, t.context.schema, t.context.uischema);
+   const tree = renderIntoDocument(
+     <Provider store={store}>
+     <TableArrayControl schema={t.context.schema}
+   uischema={t.context.uischema}
```
edgarmueller: Formatting in entire file

test/renderers/table-array.control.test.tsx
```diff
@@ -106,260 +112,429 @@ test('render two children', t => {
...
+   const validation = findRenderedDOMElementWithClass(tree, 'validation');
+   store.dispatch(update('test', () => 3));
+   store.dispatch(validate());
+   t.is(validation.textContent, 'should be array');
+   store.dispatch(update('test', () => []));
+   store.dispatch(validate());
+   t.is(validation.textContent, '');
+ });
+
+/*test('validation of nested schema', t => {
```
edgarmueller: Why is this commented? Maybe rather mark it as `failing` or `skip` or remove it completely?

Issue #701, opened 2017-11-23
New examples structure


edgarmueller:
We should split the current `index.html` of the examples into multiple files which are specific to the CSS/UI framework in use, e.g. plain, materialized.
We then also might want to publish these examples as a NPM module.

edgarmueller:
I'm re-opening this because of #719 which broke the examples

PR #719, opened 2017-12-05
Convert repo structure to monorepo

edgarmueller:
Convert repository structure to monorepo with lerna.
This PR contains a couple of issues, e.g. `package.json` files for packages need to be cleaned up, publishing is untested yet etc. We'll need to create follow-up tickets for those issues but we should merge this in order to set-up the new structure


edgarmueller:
As discussed this change breaks the examples as well as the examples and probably also a couple of scripts; follow-up issues #720 and #721have been created as well as #701 re-opened.

PR #725, opened 2017-12-07
Add support for defining a details view on a table control

eneufeld:
Adds support for providing an option `detail-uischema` which takes a
UISchema and uses this as the detail view instead of a generated one.

### DIFF
--- a/src/components/renderers/controls/array/array.directive.ts
+++ b/src/components/renderers/controls/array/array.directive.ts
@@ -1,11 +1,11 @@
 import 'lodash';
-import {PathUtil} from '../../../services/pathutil';
-import {IUISchemaGenerator} from '../../../generators/generators';
-import {AbstractControl} from '../abstract-control';
-import {IGroup} from '../../../../uischema';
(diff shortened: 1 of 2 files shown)


src/components/renderers/controls/array/array.directive.ts
```diff
@@ -83,81 +83,83 @@ const arrayTemplate = `
...
+  private arraySchema: any;
+  private arrayUiSchema: IGroup;
+  constructor(scope: ArrayControllerScope, uiGenerator: IUISchemaGenerator) {
+    super(scope);
+    let resolvedSubSchema = PathResolver.resolveSchema(
+      this.schema, this.schemaPath) as SchemaArray;
+    let items = resolvedSubSchema.items;
+    this.arraySchema = items;
+    this.properties = _.keys(items['properties']);
+    const detailUiSchema = this.uiSchema['options']['detail-uischema'];
```
edgarmueller: This fails if we have no `options`

edgarmueller:
Thanks!

PR #731, opened 2017-12-07
Port of Fixes #674, #681 and #675

eneufeld:
This ports multiple fixes from the inferno branch to the react branch and the new mono repo structure
Fixes #674 
Fixes #681
Fixes #675


packages/default/src/controls/input.control.tsx
```diff
@@ -12,14 +13,15 @@ import {
...
 export const InputControl =
-  ({ classNames, id, errors, label, uischema, schema, visible }: ControlProps) => {
+  ({ classNames, id, errors, label, uischema, schema, visible, required }: ControlProps) => {
   const isValid = errors.length === 0;
   const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+  const trim = uischema.options && uischema.options.trim;
 
   return (
-    <div className={classNames.wrapper} hidden={!visible}>
+    <div className={classNames.wrapper} hidden={!visible} data-trim={trim}>
```
edgarmueller: Can we re-use the styling registry here instead of setting a custom `data-trim` attribute?
On a different note: Shouldn't we move the initialization of the styling registry into the default set as well, just like with the material set?

packages/default/test/renderers/text.field.test.tsx
```diff
@@ -343,3 +343,239 @@ test('enabled by default', t => {
...
+    }
+  };
+  const store = initJsonFormsStore(t.context.data, schema, uischema);
+  const tree = renderIntoDocument(
+    <Provider store={store}>
+      <TextField schema={schema} uischema={uischema}/>
+    </Provider>
+  );
+  const input = findRenderedDOMElementWithTag(tree, 'input') as HTMLInputElement;
+  t.is(input.maxLength, 524288);
```
edgarmueller: While we're at it I think we should extract 524288 to a constant and give it a proper name.

packages/default/test/renderers/input.control.test.tsx
```diff
@@ -321,3 +321,57 @@ test('validation of nested schema', t => {
   t.is(validation[1].textContent, 'is a required property');
   t.is(validation[2].textContent, 'is a required property');
 });
+test('required field with warning', t => {
```
edgarmueller: What's the warning? Is is the required marker (i.e. the asterisk)?

eneufeld: no idea what this means, I just ported the stuff ;-) . We should rename this.

packages/default/src/controls/input.control.tsx
eneufeld: Yes, we could use the styling registry.
Yes we should move the init of the styling registry into default

packages/default/test/renderers/text.field.test.tsx
eneufeld: yes would make sense

PR #732, opened 2017-12-07
Port of Slider Control

eneufeld:
The slider control was contributed with PR #715 . This is a port to the
new repo structure and API using fields

### DIFF
--- a/packages/core/test/testers.test.ts
+++ b/packages/core/test/testers.test.ts
@@ -1,213 +1,233 @@
 import test from 'ava';
 import {
-    and,
-    formatIs,
-    optionIs,
-    refEndIs,
-    refEndsWith,
-    schemaMatches,
-    schemaTypeIs,
-    uiTypeIs
+  and,
+  formatIs,
+  optionIs,
+  or,
+  refEndIs,
+  refEndsWith,
+  schemaMatches,
+  schemaTypeIs,
+  uiTypeIs
 } from '../src/core/testers';
 import { JsonSchema } from '../src/models/jsonSchema';
 import { ControlElement, LabelElement } from '../src/models/uischema';
 
 test('schemaTypeIs should check type sub-schema of control', t => {
-    const schema: JsonSchema = {
-        type: 'object',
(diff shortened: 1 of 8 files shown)


packages/core/test/testers.test.ts
```diff
@@ -1,213 +1,233 @@
...
+    }
+  };
+  const uischema: ControlElement = {
+    type: 'Control',
+    scope: {
+      $ref: '#/properties/foo'
+    }
+  };
+  t.true(and(
+    schemaTypeIs('string'),
```
edgarmueller: Offtopic, but I think we should rename `schemaTypeIs` and `schemaMatches` as we actually mean the schema referenced by the control. Do you agree? If so, we should rename these in a follow-up PR.

packages/default/test/renderers/slider.field.test.tsx
```diff
@@ -0,0 +1,476 @@
...
+    <Provider store={store}>
+      <SliderField schema={t.context.schema} uischema={t.context.uischema}/>
+    </Provider>
+  );
+  const input = findRenderedDOMElementWithTag(tree, 'input') as HTMLInputElement;
+  store.dispatch(update('foo', () => null));
+  t.is(input.value, '');
+});
+// FIXME this moves the slider and changes the value
+test('update with wrong ref', t => {
```
edgarmueller: Is the comment accurate? Should this fail?

packages/core/test/testers.test.ts
eneufeld: Yes I agree, will you open an Issue for this?

packages/default/test/renderers/slider.field.test.tsx
eneufeld: I placed the FIXME at the wrong place, will fix it with the rebase

packages/core/test/testers.test.ts
edgarmueller: Yes, done, see #734

PR #733, opened 2017-12-08
Improved dev workflow

eneufeld:
Changed the webpack config such that it pulls in the renderer set and
the examples as source. This allows to do changes on the examples and
the renderer set without running an explicit building step.


packages/materialized/package.json
```diff
@@ -12,8 +12,9 @@
   "main": "dist/ts-build-materialized/index.js",
   "typings": "dist/ts-build-materialized/index.d.ts",
   "scripts": {
-    "bundle": "../../node_modules/.bin/webpack --config webpack/webpack.build.js --env=production --display-error-details",
+    "bundle": "../../node_modules/.bin/webpack --config ../../webpack.build.js --env=production --display-error-details",
```
edgarmueller: devDeps for this package.json seem to be missing
* copy-webpack-plugin
* source-map-loader
* ts-loader
* webpack-dev-server

packages/materialized/example/index.ts
```diff
@@ -1,5 +1,5 @@
 import 'jsonforms-materialized';
```
edgarmueller: Remove import as this will fail

PR #737, opened 2017-12-08
Dropdown vs radiobuttons

1903Daniel:
Fixes #

### DIFF
--- a/src/core/testers.ts
+++ b/src/core/testers.ts
@@ -19,6 +19,40 @@ export const isControl = (uischema: any): uischema is ControlElement =>
 
 
 /**
+ * Checks whether the given UI schema has an option with an enum array
+ * and whether the length of the enum is at least the given minLength.
+ * If no enum array is present, returns false.
+ *
+ * @param {number} minLength the lower bound for the length of the enum array
+ */
+export const enumLengthAtLeast = (minLength: number): Tester =>
+    (uischema: UISchemaElement, schema: JsonSchema): boolean => {
+        const options = resolveSchema(
+            schema,
+            (uischema as ControlElement).scope.$ref
+        ).enum;
+
+        return !_.isEmpty(options) && options.length >= minLength;
+    };
+
+/**
+ * Checks whether the given UI schema has an option with an enum array
+ * and whether the length of the enum is at most the given maxLength.
+ * If no enum array is present, returns false.
+ *
+ * @param {number} maxLength the upper bound for the length of the enum array
+ */
+export const enumLengthAtMost = (maxLength: number): Tester =>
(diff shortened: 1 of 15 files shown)


src/core/testers.ts
```diff
@@ -18,6 +18,40 @@ export const isControl = (uischema: any): uischema is ControlElement =>
...
 
 
+/**
+ * Checks whether the given UI schema has an option with an enum array
+ * and whether the length of the enum is at least the given minLength.
+ * If no enum array is present, returns false.
+ *
+ * @param {number} minLength the lower bound for the length of the enum array
+ */
+export const enumLengthAtLeast = (minLength: number): Tester =>
```
eneufeld: This needs tests in the core package.

src/renderers/materialized/materialized.radiobutton.control.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        return (
+                            <p>
+                                <input
+                                    type='radio'
+                                    name={label}
+                                    id={optionValue}
+                                    value={optionValue}
+                                />
+                                <label for={optionValue}
+                                       onClick={e => this.handleChange(e.target.value)}>
```
edgarmueller: Data binding doesn't work here, I think `onChange` on input should work though (like in the default radiobutton renderer)

src/core/testers.ts
```diff
@@ -18,6 +18,40 @@ export const isControl = (uischema: any): uischema is ControlElement =>
...
+    };
+
+/**
+ * Checks whether the given UI schema has an option with an enum array
+ * and whether the length of the enum is at most the given maxLength.
+ * If no enum array is present, returns false.
+ *
+ * @param {number} maxLength the upper bound for the length of the enum array
+ */
+export const enumLengthAtMost = (maxLength: number): Tester =>
```
eneufeld: @edgarmueller do you prefer two different methods, or should we rather extend the `isEnumControl` tester and add two parameters, where the first is the `minLength` and the second is `maxLength`?

test/renderers/radiobutton.control.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+                    }
+                }
+            }
+        ),
+        2
+    );
+});
+
+test('tester with matching numeric type', t => {
+    // TODO should this be true?
```
edgarmueller: Comment can be removed

src/core/testers.ts
edgarmueller: I like the latter more

1903Daniel:
currently under re-implementation for material UI

PR #744, opened 2017-12-13
added description info for input fields

AhmetTanakol:
Added description for both materialize and default
Added test cases for default version

### DIFF
--- a/packages/vanilla/src/controls/input.control.tsx
+++ b/packages/vanilla/src/controls/input.control.tsx
@@ -1,34 +1,52 @@
 import * as React from 'react';
 import {
   computeLabel,
+  Control,
+  ControlElement,
   ControlProps,
+  ControlState,
   DispatchField,
   formatErrorMessage,
   isControl,
+  isDescriptionHidden,
   mapStateToControlProps,
   RankedTester,
   rankWith,
-  registerStartupRenderer
-} from '@jsonforms/core';
+  registerStartupRenderer,
+  resolveSchema
+} from 'jsonforms-core';
 import { connect } from 'react-redux';
 
-export const InputControl =
-  ({ classNames, id, errors, label, uischema, schema, visible, required }: ControlProps) => {
-  const isValid = errors.length === 0;
-  const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+export class InputControl extends Control<ControlProps, ControlState> {
+  render() {
+    const { classNames, id, errors, label, uischema, schema, visible, required } = this.props;
+    const isValid = errors.length === 0;
+    const divClassNames = 'validation' + (isValid ? '' : ' validation_error');
+    const controlElement = uischema as ControlElement;
+    const resolvedSchema = resolveSchema(schema, controlElement.scope.$ref);
+    const inputDescriptionClassName = 'input-description';
+    const description = resolvedSchema.description === undefined ? '' : resolvedSchema.description;
 
-  return (
-    <div className={classNames.wrapper} hidden={!visible}>
-      <label htmlFor={id} className={classNames.label}>
-        {computeLabel(label, required)}
-      </label>
-      <DispatchField uischema={uischema} schema={schema}/>
-      <div className={divClassNames}>
-        {!isValid ? formatErrorMessage(errors) : ''}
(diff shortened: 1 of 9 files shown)


AhmetTanakol:
I am not sure why `lerna run build` failed because when I try to run it on my local machine, it doesn't give any error

packages/default/src/controls/input.control.tsx
```diff
@@ -1,34 +1,52 @@
...
+        onBlur={() => this.onBlur()}
+      >
+        <label htmlFor={id} className={classNames.label}>
+          {computeLabel(label, required)}
+        </label>
+        <DispatchField uischema={uischema} schema={schema}/>
+        <div className={divClassNames}>
+          {!isValid ? formatErrorMessage(errors) : ''}
+        </div>
+        {isDescriptionHidden(visible, description, this.state.isFocused) ? null : <p className={inputDescriptionClassName}>{description}</p>}
```
eneufeld: do not use classnames directly, use the style registry

eneufeld: I would hide it using the 'hidden' attribute, and why use a paragraph and not a div?

packages/materialized/src/controls/material-input.control.tsx
```diff
@@ -1,32 +1,49 @@
...
+        fullWidth={!trim}
+        onFocus={() => this.onFocus()}
+        onBlur={() => this.onBlur()}
+      >
+        <InputLabel htmlFor={id} className={classNames.label}>
+          {computeLabel(label, required)}
+        </InputLabel>
+        <DispatchField uischema={uischema} schema={schema}/>
+        <FormHelperText error={!isValid}>{errors}</FormHelperText>
+        {isDescriptionHidden(visible, description, this.state.isFocused) ? null : <FormHelperText error={false}>{description}</FormHelperText>}
```
eneufeld: can't we mere the two FormHelperText into one or is it more common to have two FormHelperText elements?

packages/default/src/controls/input.control.tsx
AhmetTanakol: I didn't use `hidden` attribute, because it will always apply the styling on focus even there is no defined description

AhmetTanakol: Sorry my mistake, It works if I use `hidden`

packages/materialized/src/controls/material-input.control.tsx
AhmetTanakol: I read helper and error text patterns on this article `https://material.io/guidelines/patterns/errors.html#errors-incompatible-state-errors`, so they just use one FormHelperText. I will make the changes on the existing FormHelperText instead of creating a new one

edgarmueller:
This one needs a rebase due to recent changes regarding the API.

AhmetTanakol:
done

packages/vanilla/src/controls/input.control.tsx
```diff
@@ -1,34 +1,59 @@
...
+        className={classNames.wrapper}
+        hidden={!visible}
+        onFocus={() => this.onFocus()}
+        onBlur={() => this.onBlur()}
+      >
+        <label htmlFor={id} className={classNames.label}>
+          {computeLabel(label, required)}
+        </label>
+        <DispatchField uischema={uischema} schema={schema}/>
+        <div className={divClassNames}>
```
eneufeld: can we unite the two divs as you did in the material renderer?
Maybe make the styling also less intrusive?

AhmetTanakol: updated

AhmetTanakol:
updated

eneufeld:
final request, could you make the description/error div only hide its content and keep its height so that the layout doesn't jump, @edgarmueller agrees with this

AhmetTanakol:
Should I display an empty string to make it look like hidden?

eneufeld:
Take a look at how material is doing this, I don't think they use an empty string, maybe a minimal height would be enough

AhmetTanakol:
@edgarmueller as we talked, I set `align-items` to `flex-start` to prevent jumpings.

PR #748, opened 2017-12-18
Define API and split up vanilla renderers

edgarmueller:
This PR mostly involves moving around types and function declarations as well as providing a common API interface to JSONForms. 
It also features some refactorings of the more complex vanillla renderers that is a result of trying to make use of the API, although it perhaps should have been made a PR itself.
This is WIP, as the materila renderers are probably broken and no final API yet has been defined.


packages/core/src/core.ts
```diff
@@ -120,14 +120,4 @@ export class JsonForms {
...
- * Annotation for registering a class as JSONForms service.
- * @param config
- * @constructor
- */
-// Disable rule because it is used as an decorator
-// tslint:disable:variable-name
-export const JsonFormsServiceElement = () => (cls: JsonFormsServiceConstructable) => {
-  JsonForms.jsonFormsServices.push(cls);
-};
 // tslint:enable:variable-name
```
eneufeld: This should be removed, as the disablement was removed

packages/core/src/generators/index.ts
```diff
@@ -1,2 +1,16 @@
...
+
 export * from './schema-gen';
 export * from './ui-schema-gen';
+
+import { generateJsonSchema } from './schema-gen';
+import { generateDefaultUISchema } from './ui-schema-gen';
+
+const Generate: {
+  jsonSchema(instance: Object, options?: any): JsonSchema;
+  uiSchema(jsonSchema: JsonSchema, layoutType?: string, prefix?: string): UISchemaElement } = {
```
eneufeld: weird formatting

packages/core/src/generators/index.ts
```diff
@@ -1,2 +1,16 @@
...
+
+  jsonSchema: generateJsonSchema,
+  uiSchema: generateDefaultUISchema,
+};
+export { Generate };
```
eneufeld: shouldn't we export this as the default export, too?

packages/default/src/additional/array/index.ts
```diff
@@ -0,0 +1,10 @@
+import {
+  registerStartupRenderer,
+} from '@jsonforms/core';
+import ArrayControlRenderer from './ArrayControlRenderer';
+import { arrayTester } from './tester';
```
eneufeld: shouldn't we also export the renderer to allow to extend it?

packages/default/src/additional/categorization/index.ts
```diff
@@ -0,0 +1,8 @@
+import { registerStartupRenderer } from '@jsonforms/core';
+import CategorizationRenderer from './CategorizationRenderer';
+import { categorizationTester } from './tester';
+
```
eneufeld: Shouldn't we export all the components and the renderer to allow an easy extension?

edgarmueller:
@eneufeld I've updated the PR

PR #754, opened 2017-12-25
Improved Material Table

### LINKED ISSUE #729: Improve Table Array Renderer
The Table Array Renderer does not show errors of children.
The material version should probably use a `Fab` button and material-ui-icons instead of a normal button

eneufeld:
Fixes #729


eneufeld:
This will conflict with #748 und #749

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+
+        return;
+      }
+      this.setState({ selected: [] });
+    };
+
+    const handleClick = (_event, child) => {
+      const selected = this.state.selected.filter(s => data.indexOf(s) !== -1);
+      const selectedIndex = selected.indexOf(child);
+      let newSelected = [];
```
edgarmueller: I'd remove the initialization and instead add an else case in the following if-else-clause. This seems clearer to since we perform the `concat` with an empty array anyway in all the cases so far, which seems unnecessary.

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+      childErrors,
+      label: this.props.label,
+      path,
+      uischema,
+      numSelected,
+      handleOpenConfirmDelete,
+      dispatch: this.props.dispatch
+    };
+    const removeItems = () => {
+      this.props.dispatch(
```
edgarmueller: Extract to `mapDispatchToProps` (analogously to default export of `ArrayControlRenderer.tsx`). I think it would be beneficial to follow the same structure as with the vanilla renderer, meaning that we do the split between the presentational component and the container component.

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+        </Dialog>
       </div>
     );
   }
 }
+export interface TableProps extends ControlProps {
+  childErrors: ErrorObject[];
+}
+// FIXME copied from validation.ts
+const errorAt = instancePath => (state): ErrorObject[] => {
```
edgarmueller: Doesn't seem to be an exact copy. Why the difference? If we need this, then we should provide this extractor alongside the regular `errorAt` extractor?

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
 
+  return _.filter(state.errors, (error: ErrorObject) => error.dataPath.startsWith(path));
+};
+const mapStateToTableControlProps = (state, ownProps) => {
+  const {data, errors, classNames, label, visible, enabled, id, path, inputs, required} =
+    mapStateToControlProps(state, ownProps);
+
+  const childErrors = errorAt(path)(getValidation(state));
+
+  return {
```
edgarmueller: I think this can be shorted with de-structuring and spread syntax:
```js
const {path, ...props } = mapStateToControlProps(state, ownProps);
const childErrors = errorAt(path)(getValidation(state));
return {
   ...props,
  path,
  childErrors
};
```

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+const TableToolbar = ({errors, childErrors, label, path, uischema,
+  numSelected, handleOpenConfirmDelete, dispatch}) => {
+  const controlElement = uischema as ControlElement;
+  const labelClass = JsonForms.stylingRegistry.getAsClassName('array-table.label');
+  const buttonClass = JsonForms.stylingRegistry.getAsClassName('array-table.button');
+  const labelObject = Helpers.createLabelDescriptionFrom(controlElement);
+  const allErrors = [].concat(errors).concat(childErrors.map(e => e.message));
+
+  // TODO duplicate code
+  const addItem = () => {
```
edgarmueller: If we split up this into presentational and container component (as mentioned earlier) we might also provide some table-related common functionality that is used by both the vanilla and the material renderer in order to avoid code duplication.

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+      <Grid container alignItems='center' justify='space-between'>
+        <Grid item>
+          <Typography type='title' className={labelClass}>{label}</Typography>
+        </Grid>
+        <Grid item hidden={{smUp: allErrors.length === 0}}>
+          <ValidationIcon id='tooltip-validation' errorMessages={allErrors}/>
+        </Grid>
+        <Grid item>
+          <Tooltip id='tooltip-add' title={`Add to ${labelObject.text}`} placement='bottom'>
+            <Button
```
edgarmueller: Add some horizontal spacing between both buttons.

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
-  connect(mapStateToControlProps)(TableArrayControl)
+  connect(mapStateToTableControlProps)(TableArrayControl)
 );
+const TableToolbar = ({errors, childErrors, label, path, uischema,
+  numSelected, handleOpenConfirmDelete, dispatch}) => {
+  const controlElement = uischema as ControlElement;
+  const labelClass = JsonForms.stylingRegistry.getAsClassName('array-table.label');
+  const buttonClass = JsonForms.stylingRegistry.getAsClassName('array-table.button');
+  const labelObject = Helpers.createLabelDescriptionFrom(controlElement);
+  const allErrors = [].concat(errors).concat(childErrors.map(e => e.message));
```
edgarmueller: No need to concat on empty array. Maybe it would be helpful to include the current index in the message, which would helps users to identify at which entries contain errors? What do you think?

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+
+      this.setState({ selected: newSelected });
+    };
+
+    const isSelected = child => this.state.selected.indexOf(child) !== -1;
+    const handleOpenConfirmDelete = () => {
+      this.setState({ openConfirmDelete: true });
+    };
+
+    const handleClose = () => {
```
edgarmueller: Make helper methods part of the accompanying class instead of the `render` method.

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+    handleClick, isSelected} = tableProps;
+
+  return data.map((child, index) => {
+    const childPath = Paths.compose(path, index + '');
+
+    return (
+      <TableRow
+        key={childPath}
+        hover
+        onClick={event => handleClick(event, child)}
```
edgarmueller: I'd prefer to only select a row if the selection checkbox has been clicked explicitely. Currently, the row is marked as being selected if I only a field in order to edit it.

packages/material/src/complex/table-array.control.tsx
```diff
@@ -41,117 +55,325 @@ export const tableArrayTester: RankedTester = rankWith(10, and(
...
+      >
+        <TableCell padding='checkbox'>
+          <Checkbox checked={isSelected(child)} />
+        </TableCell>
+        {generateCells(TableContentCell, resolvedSchema, childPath, childErrors)}
+      </TableRow>
+    );
+  });
+};
+const TableContentCell = ({rowPath, cellProperty, cellPath, errors, resolvedSchema}) => {
```
edgarmueller: It might be worthwhile to extract these helper components to a file of their own.

eneufeld:
Thanks for the review.
I will remove the implicit row selection as it produces bugs anyway and keep only the checkbox.
About the split:
So I agree to split this into some Material Components (no renderer logic there) and a renderer Component which uses those Material Components (basically what is already there but split it into multiple files).
In order to reuse code in the vanilla renderer and the material renderer we would need additional logic in the core or define one renderer set as the master from which to extend.

edgarmueller:
Thanks for the update. The only issue left is that the checkbox that is used to toggle the selection of all rows does not correctly work:
1. It does not toggle the selection state of all involved rows
2. It does not update itself.

PR #755, opened 2017-12-25
Fixed material maxLength support

eneufeld:
Fixes #675


eneufeld:
I would prefer to merge this and add an initial set of tests for material as an explicit PR.
What is your take on this?

eneufeld:
created #759

PR #756, opened 2017-12-25
improved error visibility on material controls and fields

eneufeld:
Improves Error visibility in material controls and fields


eneufeld:
created #759

edgarmueller:
@eneufeld this needs a rebase

PR #758, opened 2017-12-27
fixed runtime evalEnablement and evalVisibility

eneufeld:
evalEnablement and evalVisibility called resolveSchema instead of resolveData

### DIFF
--- a/packages/core/test/runtime.test.ts
+++ b/packages/core/test/runtime.test.ts
@@ -1,0 +1,202 @@
+import test from 'ava';
+import { ControlElement, LeafCondition, RuleEffect } from '../src/models/uischema';
+
+import { evalEnablement, evalVisibility } from '../src/helpers/runtime';
+
+test('evalVisibility show valid case', t => {
+    const leafCondition: LeafCondition = {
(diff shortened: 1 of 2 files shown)


packages/core/test/runtime.test.ts
```diff
@@ -0,0 +1,208 @@
...
+            'condition': leafCondition
+        }
+    };
+    const data = {
+        value: 'foo',
+        ruleValue: 'foobar'
+    };
+    t.deepEqual(evalEnablement(uischema, data), false);
+});
+test('evalEnablement diable valid case', t => {
```
edgarmueller: typo

packages/core/test/runtime.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        'rule': {
+            'effect': RuleEffect.SHOW,
+            'condition': leafCondition
+        }
+    };
+    const data = {
+        value: 'foo',
+        ruleValue: 'bar'
+    };
+    t.deepEqual(evalVisibility(uischema, data), true);
```
edgarmueller: `t.is` would be sufficient.

PR #760, opened 2017-12-31
Fix materiallayout

eneufeld:
The Material Layout renderer does not use the Grid element. This is fixed by this commit.
This also renames the material layout renderer.


packages/material/src/layouts/layout.util.tsx
```diff
@@ -0,0 +1,44 @@
...
+export interface MaterialLayoutRendererProps {
+    elements: UISchemaElement[];
+    schema: JsonSchema;
+    path: string;
+    visible: boolean;
+    direction: 'row'|'column';
+}
+export const MaterialLayoutRenderer = (
+    {visible, elements, schema, path, direction}: MaterialLayoutRendererProps) => {
+        if (!_.isEmpty(elements)) {
```
edgarmueller: This will cause an error if `elements` is empty. Return `null` if you don't want to render anything instead

Issue #750, opened 2017-12-19
Bundle builds broken


sdirix:
Currently (react-branch, commit 0426062) the bundle builds for all packages are broken. These should be fixed and probably added to the "preparePublish" tasks in each package.

PR #770, opened 2018-01-10
updated webpack to fix bundle script

eneufeld:


sdirix:
This PR is related to #750

PR #773, opened 2018-01-10
Re-organize example code a bit and improve usage o dispatch renderer in entry point

edgarmueller:
Minor updates to the structure of the examples:
* Avoid code duplication, instead reference previous days
* Export day 1 and 3 of make it happen example
* Export custom rating control

Furthermore this PR contains commits that improve the usage of `DispatchRenderer` when used as an entry point to rendering. So far we had to specify the `schema` and the `uischema` properties, which actually is not necessary since these are the same as stored in the state container. 

Finally, this PR provides a default value for the uischema within the `initJsonFormsStore` function.

### DIFF
--- a/packages/core/src/store.ts
+++ b/packages/core/src/store.ts
@@ -1,11 +1,12 @@
 import thunk from 'redux-thunk';
-import { appReducer } from './reducers/index';
+import { appReducer } from './reducers';
 import { applyMiddleware, createStore } from 'redux';
 import { JsonFormsStore } from './json-forms';
 import { JsonForms } from './core';
 import { INIT, VALIDATE } from './actions';
 import { JsonSchema } from './models/jsonSchema';
 import { UISchemaElement } from './models/uischema';
+import { generateDefaultUISchema } from './generators';
 
 export const createJsonFormsStore = (initialState): JsonFormsStore => {
   // TODO: typing
@@ -18,8 +19,11 @@ export const createJsonFormsStore = (initialState): JsonFormsStore => {
   return store as JsonFormsStore;
 };
 
-export const initJsonFormsStore =
-  (data: any, schema: JsonSchema, uischema: UISchemaElement): JsonFormsStore => {
+export const initJsonFormsStore = (
+  data: any,
+  schema: JsonSchema,
+  uischema: UISchemaElement = generateDefaultUISchema(schema)
+): JsonFormsStore => {
 
     const store = createJsonFormsStore({
       common: {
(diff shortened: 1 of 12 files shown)


packages/core/src/store.ts
```diff
@@ -18,8 +19,11 @@ export const createJsonFormsStore = (initialState): JsonFormsStore => {
   return store as JsonFormsStore;
 };
 
-export const initJsonFormsStore =
-  (data: any, schema: JsonSchema, uischema: UISchemaElement): JsonFormsStore => {
+export const initJsonFormsStore = (
+  data: any,
+  schema: JsonSchema,
```
eneufeld: shouldn't the schema also be generated if it is missing?

eneufeld:
Shouldn't we also remove the handling of null uischematas in the JSONForms Element now?

packages/core/src/json-forms.tsx
```diff
@@ -137,20 +109,13 @@ export class JsonFormsElement extends HTMLElement {
       return;
     }
 
-    if (_.isEmpty(this.dataObject)) {
```
eneufeld: why don't we check for the dataobject existence anymore?

packages/vanilla/test/json-forms.test.ts
```diff
@@ -101,9 +101,10 @@ test.cb('render with data schema and UI schema set', t => {
   setTimeout(
     () => {
       jsonForms.connectedCallback();
-      t.is(jsonForms.children.length, 0);
-      t.deepEqual(jsonForms.dataSchema, t.context.schema);
-      t.is(jsonForms.uiSchema, t.context.uischema);
+      // label is rendered
+      t.is(jsonForms.children.length, 1);
```
eneufeld: why is the label now rendered?

edgarmueller:
Because we render now even if the data is empty, which was not the case before, although we have a schema and an UI schema available. I think the condition for not rendering (which I've removed for now) should be refined to checking whether the `dataObject` is either undefined or null and only in case the `dataschema` is empty we cancel rendering. I can add that if you think it's necessary.

eneufeld:
If the null/undefined data case is handled correctly now, it's fine by me. But I assume, that you could now also render with no data, no schema and no uischema?

packages/core/src/json-forms.tsx
```diff
@@ -116,41 +102,25 @@ export class JsonFormsElement extends HTMLElement {
...
   private render(): void {
     if (!this.allowDynamicUpdate) {
       return;
     }
     if (this.schemaPromise !== null) {
       return;
     }
-
-    if (_.isEmpty(this.dataObject)) {
+    if ((this.dataObject === undefined || this.dataObject === undefined)
```
eneufeld: one of those dataObject checks should be null, I assume

PR #778, opened 2018-01-11
Cleaned example package.json and day6-example

### LINKED ISSUE #772: Cleanup examples package.json
`dependencies` as well as `ava` block might be obsolete.

eneufeld:
Fixes #772


packages/examples/src/day6.ts
```diff
@@ -0,0 +1,141 @@
+import { registerExamples } from './register';
```
edgarmueller: Do we have day6 twice now? If yes, why so?

eneufeld: something went wrong, i will rebase this and clean up

PR #779, opened 2018-01-12
Add i18n to store with a translation object and locale value

AhmetTanakol:
This commit includes just a way to add a translation object and locale value on `initJsonFormsStore`. Currently, there is no translating functionality. The next step is to translate labels and descriptions from `schema.json`. As we discussed during the last meeting, what I understood is that we need to translate these information before initializing jsonforms store (or replace values after initializing? ) so that we can use the translated version of `schema.json`. If you can share your feedback, it would be great.

### DIFF
--- a/packages/core/src/reducers/i18n.ts
+++ b/packages/core/src/reducers/i18n.ts
@@ -3,7 +3,7 @@ import { SET_LOCALE, SET_TRANSLATIONS } from '../actions';
 export const i18nReducer = (
   state = {
     translations: {},
-    locale: navigator.language
+    locale: navigator.languages[0]
   },
   action) => {
   switch (action.type) {
--- a/packages/core/src/reducers/i18n.ts
+++ b/packages/core/src/reducers/i18n.ts
@@ -1,10 +1,7 @@
 import { SET_LOCALE, SET_TRANSLATIONS } from '../actions';
 
 export const i18nReducer = (
-  state = {
-    translations: {},
-    locale: navigator.languages[0]
-  },
+  state = {},
   action) => {
   switch (action.type) {
     case SET_TRANSLATIONS:
@@ -20,4 +17,8 @@ export const i18nReducer = (
     default:
       return state;
   }
+};
+
+export const fetchTranslation = state => {
+  return state.translations ? state.translations[state.locale] : undefined;
 };
--- a/packages/core/src/reducers/i18n.ts
+++ b/packages/core/src/reducers/i18n.ts
@@ -1,13 +1,13 @@
-import { LOAD_TRANSLATION, SET_LOCALE } from '../actions';
+import { SET_LOCALE, SET_TRANSLATIONS } from '../actions';
 
-export const translationReducer = (
+export const i18nReducer = (
   state = {
     translations: {},
-    locale: ''
+    locale: navigator.language
   },
   action) => {
   switch (action.type) {
-    case LOAD_TRANSLATION:
+    case SET_TRANSLATIONS:
       return {
         ...state,
         translations: action.translations
--- a/packages/core/src/reducers/i18n.ts
+++ b/packages/core/src/reducers/i18n.ts
@@ -1,7 +1,9 @@
 import { SET_LOCALE, SET_TRANSLATIONS } from '../actions';
 
 export const i18nReducer = (
-  state = {},
+  state = {
+    locale: navigator.languages[0]
+  },
   action) => {
   switch (action.type) {
     case SET_TRANSLATIONS:
@@ -12,7 +14,7 @@ export const i18nReducer = (
     case SET_LOCALE:
       return {
         ...state,
-        locale: action.locale
+        locale: action.locale === undefined ? navigator.languages[0] : action.locale
       };
     default:
       return state;
--- a/packages/core/src/reducers/index.ts
+++ b/packages/core/src/reducers/index.ts
@@ -35,4 +35,4 @@ export const getErrorAt = instancePath => state => {
 };
 export const getSubErrorsAt = instancePath => state => subErrorsAt(instancePath)(state.jsonforms.validation);
 
-export const getTranslations = state => fetchTranslation(state.i18n);
+export const getTranslations = state => fetchTranslation(state.jsonforms.i18n);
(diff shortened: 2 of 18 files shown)


packages/core/src/store.ts
```diff
@@ -41,5 +42,15 @@ export const initJsonFormsStore =
       data
     });
 
+    store.dispatch({
```
edgarmueller: No need to dispatch here since we can use the initial state for doing so. Note that with #780 the signature of this function will change: it will expect the overall initial state instead of multiple parameters.

packages/core/src/actions.ts
```diff
@@ -15,6 +15,8 @@ export const ADD_RENDERER = 'ADD_RENDERER';
 export const REMOVE_RENDERER = 'REMOVE_RENDERER';
 export const ADD_FIELD = 'ADD_FIELD';
 export const REMOVE_FIELD = 'REMOVE_FIELD';
+export const LOAD_TRANSLATION = 'LOAD_TRANSLATION';
```
edgarmueller: Probably rename to `SET_TRANSLATIONS'.

packages/core/src/reducers/index.ts
```diff
@@ -3,13 +3,15 @@ import { validationReducer } from './validation';
...
+import { translationReducer } from './translation';
 
 export const appReducer: Reducer<any> = combineReducers(
   {
     'common': commonStateReducer,
     'validation': validationReducer,
     'renderers': rendererReducer,
-    'fields': fieldReducer
+    'fields': fieldReducer,
+    'translation': translationReducer
```
edgarmueller: I'd give an other name to the reducer, maybe i18n? We'll probably also store other i18n related bits there.

packages/core/src/reducers/translation.ts
```diff
@@ -0,0 +1,23 @@
+import { LOAD_TRANSLATION, SET_LOCALE } from '../actions';
+
+export const translationReducer = (
+  state = {
+    translations: {},
+    locale: ''
```
edgarmueller: We should use a default locale.

AhmetTanakol:
I tried to update regarding to your comments. Also added unit tests to see if initializing store with translation and locale is working or not. If locale is not set on store initialization, I assign locale to `navigator.language`

AhmetTanakol:
Added `translate` function to set label regarding to defined i18n properties

packages/core/src/reducers/index.ts
```diff
@@ -19,7 +19,12 @@ export const getData = state => {
   return extractData(state.common);
 };
 
+const fetchTranslation = state => {
```
eneufeld: the fetchTranslation method does not fit the fetchTranslation call
I would inline this if possible

packages/examples/src/example.ts
```diff
@@ -6,5 +6,7 @@ export interface ExampleDescription {
   data: any;
   schema: JsonSchema;
   uiSchema: UISchemaElement;
+  translations?: any;
+  locale?: String;
```
eneufeld: you could assign a default value to locale, so you don't have wo handle this in the changeExample method

packages/core/src/helpers/label.ts
```diff
@@ -54,3 +54,12 @@ export const createLabelDescriptionFrom = (withLabel: ControlElement): LabelDesc
     };
   }
 };
+
+export const translateLabel = (translations, label: LabelDescription): LabelDescription => {
```
eneufeld: a test case is missing for this

packages/core/src/json-forms.tsx
```diff
@@ -102,6 +104,18 @@ export class JsonFormsElement extends HTMLElement {
     });
   }
 
+  set translations(_translations: Object) {
+    this._translations = _translations;
+    // TODO
```
eneufeld: what is missing here?

AhmetTanakol:
I updated the code regarding to feedbacks and added unit tests for label translation

packages/core/src/reducers/i18n.ts
```diff
@@ -0,0 +1,24 @@
+import { SET_LOCALE, SET_TRANSLATIONS } from '../actions';
+
+export const i18nReducer = (
+  state = {},
```
edgarmueller: I know I said otherwise the last time but I think the default intialization of the `locale` should happen here. Reason being is that users probably will not always use the `initJsonFormsStore` function to initialize the store and hence no default locale will be set.

packages/core/src/renderers/renderer.util.tsx
```diff
@@ -131,7 +132,8 @@ export const mapStateToControlProps = (state, ownProps) => {
   const path = composeWithUi(ownProps.uischema, ownProps.path);
   const visible = _.has(ownProps, 'visible') ? ownProps.visible :  isVisible(ownProps, state);
   const enabled = _.has(ownProps, 'enabled') ? ownProps.enabled :  isEnabled(ownProps, state);
-  const labelDesc = createLabelDescriptionFrom(ownProps.uischema);
+  let labelDesc = createLabelDescriptionFrom(ownProps.uischema);
+  labelDesc = translateLabel(getTranslations(state), labelDesc);
```
edgarmueller: Prefer nesting functions call instead of mutating `labelDesc`.

packages/examples/src/person.ts
```diff
@@ -156,12 +156,32 @@ export const data = {
...
+    nationality: 'Nationality',
+    occupation: 'Occupation',
+    birthday: 'Birthday',
+  },
+  'de-DE': {
+    name: 'Name',
+    height: 'Höhe',
+    age: 'Alter',
+    nationality: 'Staatsangehörigkeit',
+    occupation: 'Besetzung',
```
edgarmueller: I think the proper german term is 'Tätigkeit' or 'Beschäftigung' ;)

AhmetTanakol:
updated

AhmetTanakol:
fixed assigning default value of locale

edgarmueller:
@eneufeld Could you update your review on this?

AhmetTanakol:
updated

edgarmueller:
Thank you

PR #780, opened 2018-01-12
Replace global styling registry with reducer

edgarmueller:
Also:
- Change signature of initJsonFormsStore to expect single state object
- Replace dot-prop-immutable with lodash API
- Remove warnings


packages/core/src/renderers/renderer.util.tsx
```diff
@@ -146,19 +143,20 @@ export const mapStateToControlProps = (state, ownProps) => {
...
   if (trim) {
-    classNames = classNames.concat(JsonForms.stylingRegistry.get('control.trim'));
+    classNames = classNames.concat(getStyle(state)('control.trim'));
   }
   const inputClassName =
     ['validate']
       .concat(isValid ? 'valid' : 'invalid');
-  const labelClass = JsonForms.stylingRegistry.getAsClassName('control.label');
+  const labelClass = getStyleAsClassName(state)('control.label');
+  const descriptionClassName = getStyleAsClassName(state)('input-description');
```
eneufeld: I think this should be `control.description`

packages/core/src/renderers/renderer.util.tsx
```diff
@@ -54,31 +56,26 @@ export const mapStateToLayoutProps = (state, ownProps) => {
   return {
     renderers: state.renderers,
     visible,
-    path: ownProps.path
+    path: ownProps.path,
+    getStyle: getStyle(state),
```
eneufeld: this is vanilla renderer specific, maybe this should be handled similar to the vanillacontrolprops ?

packages/core/src/actions.ts
```diff
@@ -59,3 +63,19 @@ export const unregisterRenderer = (
   tester,
   renderer
 });
+
+export const registerStyle = (styleName: string, classNames: string[]) => ({
```
eneufeld: can we make this specific to vanilla?

packages/core/src/renderers/renderer.util.tsx
```diff
@@ -54,31 +56,26 @@ export const mapStateToLayoutProps = (state, ownProps) => {
...
+    getStyleAsClassName: getStyleAsClassName(state)
   };
 };
 
 export const renderChildren = (
```
eneufeld: I think we should move the whole method to the vanilla renderer set

packages/core/src/renderers/renderer.util.tsx
```diff
@@ -54,31 +55,24 @@ export const mapStateToLayoutProps = (state, ownProps) => {
   return {
     renderers: state.renderers,
     visible,
-    path: ownProps.path
+    path: ownProps.path,
   };
 };
 
 export const renderChildren = (
```
eneufeld: this is vanilla dependent now as material is using its own implementation

packages/material/src/controls/material-input.control.tsx
```diff
@@ -31,15 +33,18 @@ export class MaterialInputControl extends Control<ControlProps, ControlState> {
     if (!visible) {
       style = {display: 'none'};
     }
+    const classNames: string[] = !_.isEmpty(controlElement.scope) ?
```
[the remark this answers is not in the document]
edgarmueller: Material input tests are expecting this class in order to retrieve controls to assert upon. Of course, we could also update the test cases, I guess.

PR #783, opened 2018-01-16
added datetime widget for material ui

1903Daniel:
Fixes #


package.json
```diff
@@ -16,6 +16,11 @@
 		"tslint-loader": "^3.5.3",
 		"tslint-react": "^3.2.0",
 		"typescript": "^2.6.2",
-		"webpack": "^3.10.0"
+		"webpack": "^3.10.0",
```
eneufeld: do not add material dependencies to root package

packages/material/example/index.html
```diff
@@ -8,6 +8,8 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <script src="assets/native-shim.js"></script>
 
+  <!--- Material Icons -->
+  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
eneufeld: I already included the dependency to https://github.com/MODX-Club/material-ui-icons, I would suggest to use it instead of the cdn

1903Daniel: is used by the framework, so we can't use the dependency

1903Daniel: @eneufeld I don't have the "ignore" button, so can you please click ignore or else remove the request?

1903Daniel:
@eneufeld @edgarmueller 
fixed, please review

packages/material/example/index.html
eneufeld: Hi, 
I just found this: 
> If you dont want to use icon font, or you are already use material-ui-icons you can pass any icon to the components with the following props
leftArrowIcon - arrow left for datepicker
rightArrowIcon - arrow right for datepicker
dateRangeIcon - date tab icon for datetimepicker
timeIcon - time tab icon for datetimepicker
keyboardIcon - icon for keyboard end adornmentic

from the readme: https://github.com/dmtrKovalenko/material-ui-pickers

1903Daniel: cool, thank you 😄 is changed now

1903Daniel:
@edgarmueller So in the meeting, we discussed I should create a new PR without i18n. However, the concern of @eneufeld is regarding moment dependency. Can you please check with Jonas on how to proceed? Shall I still create an additional PR without i18n, and also for date only and time only widgets accordingly? Or shall I leave this as is, and just create one PR each for the other 2 components that also include i18n?

edgarmueller:
@1903Daniel Isn't the momentjs dependency result of the i18n efforts?

1903Daniel:
@edgarmueller no, the pickers framework needs moment:
https://github.com/dmtrKovalenko/material-ui-pickers

Available as npm package. Please note that we are using moment as a peer dependency

npm install moment material-ui-pickers -S

edgarmueller:
I see, thanks for clarifying. Nevertheless I'd prefer a PR without i18n (which will not be part of the initial release next week), because then we could still think about providing these widgets as packages.
@eneufeld What's your take on this? Do you think it makes sense to keep the 'native' widgets as default, but to provide an additional package which includes the moment dependency?
@1903Daniel Let's wait for Eugen's response

1903Daniel:
as discussed, no longer needed because new PR w/o i18n is open

PR #784, opened 2018-01-16
Introduce parentPath prop in controls and further minor fixes

lucas-koehler:
Added a `parentPath` property to input controls.
The parent property points to the data parent data that contains the
property specified in the UI Schema Element

* Hand down the parent path instead of the resolved path to a `DispatchField`
* This is necessary because the `DispatchField` resolves the path again against the UI Schema
* Adapted Vanilla and Material Input Control to propagate the `parentPath`

Other fixes:
* Tree Renderer: fixed `setSelection` method signature
* MaterialTableControl: Set `rowCount` to zero if the table's data is `undefined` or `zero` to avoid errors

### DIFF
--- a/packages/core/src/renderers/dispatch.field.tsx
+++ b/packages/core/src/renderers/dispatch.field.tsx
@@ -24,8 +24,6 @@ const Dispatch = (dispatchFieldProps: DispatchFieldProps) => {
         schema={schema}
         uischema={uischema}
         path={dispatchFieldProps.path}
-        data={dispatchFieldProps.data}
-        class
       />
     );
   }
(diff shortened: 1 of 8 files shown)


lucas-koehler:
Added Material boolean control

packages/core/src/renderers/dispatch.field.tsx
```diff
@@ -24,8 +24,6 @@ const Dispatch = (dispatchFieldProps: DispatchFieldProps) => {
         schema={schema}
         uischema={uischema}
         path={dispatchFieldProps.path}
-        data={dispatchFieldProps.data}
```
eneufeld: why don't we need to set the data anymore?

lucas-koehler: Because the field resolves it itself by using the path.

PR #787, opened 2018-01-17
Remove classNames property from material input and boolean controls

edgarmueller:
Also update Base Control along the way:

* Replace handleChange method in Control with props
* Delegate from mapDispatchToFieldProps to mapDispatchToControlProps
* Update test
* Add JSDoc to ControlProps
* Use correct type parameter in material input control


packages/core/src/renderers/Control.ts
```diff
@@ -1,25 +1,65 @@
...
   path: string;
+
+  /**
+   * Path of the parent renderer, if any.
+   */
   parentPath?: string;
-  classNames: ControlClassNames;
+
+  /**
+   * An unique ID that can be used to identify the rendered form.
```
eneufeld: that is the id of a single control, so maybe `element` instead of `form`

packages/core/src/renderers/renderer.util.tsx
```diff
@@ -153,6 +153,14 @@ export const mapStateToControlProps = (state, ownProps) => {
     required
   };
 };
+
+export const mapDispatchToControlProps = dispatch => ({
+  handleChange(path, value) {
```
eneufeld: do we have a case where the path of the handleChange is a different one than the one in the properties?
I think we only have the case where we pass in the path from the properties to the handleChange method.
We could then get rid of the path parameter by providing two parameter to the mapDispatchToControlProps as the second one could be the props: `mapDispatchToProps(dispatch, [ownProps]): dispatchProps`

edgarmueller: Not sure if I got this right, but the relevant `path` for updating the respective data bit of a control is determined is `mapStateToProps`, isn't it?

packages/core/src/renderers/Control.ts
edgarmueller: I'm fine with element, updated.

packages/examples/package.json.orig
```diff
@@ -0,0 +1,42 @@
+{
```
eneufeld: this should not be part of the commit I assume

PR #800, opened 2018-01-19
Add script for file header generation (#747)

### LINKED ISSUE #747: Set up file headers
All files should carry a copyright header with the name of the contributing company.

mathansen:
* Add add-file-header.sh (invoke without arguments to display usage)
* Add LICENSE file (MIT License)

Fixes #747


add-file-headers.sh
```diff
@@ -0,0 +1,136 @@
...
+#
+# Helper script to add headers to source files.
+# Invoke without arguments to display usage.
+#
+# date:   2018-19-01
+# author: Mat Hansen <mhansen@eclipsesource.com>
+#
+
+MODE="undefined"
+HEADER_FILE=""
```
edgarmueller: Is there no switch to set `HEADER_FILE`? Otherwise it should be hard-coded to `LICENSE`, I guess?

mathansen: In fact there is one (-h=*|--header=*).  But you are right, it is missing in the usage example. I will take care of that.

edgarmueller:
Thanks

PR #801, opened 2018-01-19
Make displaying of 'required' validation message configurable

doruktuncel:


edgarmueller:
Your proposed solution can not work as you do not check what kind of error is going to be displayed. Remember that we only want to hide error messages that say 'is a required property'.

edgarmueller:
@doruktuncel Any updates on this?

doruktuncel:
@edgarmueller I am on it, aiming to finalize it by tomorrow.

PR #808, opened 2018-01-21
added value label to slider, updated tester to check for option 'slider'

### LINKED ISSUE #795: Improve slider control
- Find out more reasonable criteria to display, e.g. for age we do not want a slider, maybe a new option?
- Display current value
- Label overlaps with control

eneufeld:
Fixes #795

### DIFF
--- a/packages/core/src/testers/index.ts
+++ b/packages/core/src/testers/index.ts
@@ -278,3 +278,18 @@ export const isArrayObjectControl = and(
     schema.type === 'object'
   )
 );
+
+/**
+ * Tests whether a given UI schema is of type Control and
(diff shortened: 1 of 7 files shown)


eneufeld:
In the demoes the initial value is undefined, so there is actually nothing to show. Any suggestion what to show?

edgarmueller:
Ah, makes sense, then I'd probably change the default initialization (`(resolvedSchema.maximum - resolvedSchema.minimum) / 2`) to the minimum, but maybe get some other opinions on that as well.

eneufeld:
as discussed with @sdirix  we will check for a default value and use this in the slider, I will also remove the uischema option

PR #809, opened 2018-01-21
added config json

1903Daniel:
@edgarmueller @eneufeld 
I have one question about why the config can't be passed into the test cases. Maybe you can already tell from looking at the test cases, or else we can discuss it on Tuesday.
Everything works fine when trying out the examples, but the test cases throw an error, e.g.

  renderers » text.field » maxLength not specified, attributes should have default values

  C:\Users\User\Documents\git\packages\vanilla\src\fields\text.field.tsx:36

   35:   );
    6: };
    7:

  Error thrown in test:

  TypeError {
    message: 'Cannot read property \'restrict\' of undefined',
  }

### DIFF
--- a/packages/core/src/actions/index.ts
+++ b/packages/core/src/actions/index.ts
@@ -6,6 +6,7 @@ import { RankedTester } from '../testers';
 export const INIT = 'INIT';
 export const UPDATE_DATA = 'UPDATE';
 export const UPDATE_UI = 'UPDATE_UI';
+export const UPDATE_CONFIG = 'UPDATE_CONFIG';
 export const VALIDATE = 'VALIDATE';
 export const SHOW = 'SHOW';
 export const HIDE = 'HIDE';
--- a/packages/core/src/configDefault.ts
+++ b/packages/core/src/configDefault.ts
@@ -1,0 +1,4 @@
+export const configDefault = {
+    restrict: true,     // [text] whether to restrict the number of enterable characters to maxLength, if specified in the JSON schema
+    trim: true          // [text] whether to resize the input's width to maxLength, if specified in the JSON schema
+};
--- a/packages/core/src/reducers/common.ts
+++ b/packages/core/src/reducers/common.ts
@@ -1,5 +1,21 @@
 import * as _ from 'lodash';
-import { INIT, UPDATE_DATA } from '../actions';
+import { INIT, UPDATE_DATA, UPDATE_CONFIG } from '../actions';
+import { configDefault } from '../configDefault';
+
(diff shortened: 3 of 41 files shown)


edgarmueller:
@1903Daniel Most likely because the `config` is undefined. Check the test cases whether the initialization of the store always contains the config. Maybe you can a default value of the config state within the reducer which from my POV makes sense and also should fix the problem?

1903Daniel:
@edgarmueller, hmm, yes the test cases contain the config where needed. Even if I don't supply it, within the `initJsonFormsStore` functions (in both setup.ts and store.ts) I already set a default value to it. So I don't see how moving the default to the reducer would fix it. But I can give it a try and let you know

1903Daniel:
@edgarmueller So as I suspected, moving the function where default values are set to the reducer does not resolve the error. If we can look at it in the meeting, that'd be great

edgarmueller:
I don't see any default initialization in the reducer happening (or did you not commit it?). Also, I'm quite sure that for the test cases currently failing you do not supply the initial config state. But sure, we can have together at today's meeting.

1903Daniel:
@edgarmueller 
can you please take a look at the common.ts from line 107 and the example configCustom.ts from L32.

when initializing a renderer, the reducer always seems to run into the default case (common.ts) So I figured I need to apply the default configuration in the default case from L107 to the passed in state.config, which I did.

In an example where I don't provide the config object, that works fine. But in the configCustom.ts example, I always get the default back. So I think that the config object does not get passed down to the reducer, neither from the example nor from the vanilla test cases - but I couldn't figure out why. Can you see it?

edgarmueller:
@1903Daniel Your `applyDefaultConfiguration` looks wrong. In `if (!initConfig[key])`  you want to check for the presence of the property `key`, not its value. You can do so either via `hasOwnProperty` or lodash's `has` function.

On a different note: why do you call `applyDefaultConfiguration` in the default case? Every action unknown should simply return the current state. Therefore, why don't you initialize the configuration within the parameter list?

1903Daniel:
@edgarmueller Ah, thank you for pointing out the bug.
By parameter list, do you mean the function arguments right after export const commonStateReducer = ... ?
I thought that right-hand side {} in the parameter list ( config : {} ) only applies when no config object was supplied to the function (that is undefined or null) - so I would not catch the case when I already have a config defined, but not certain keys. Or is my understanding wrong?

edgarmueller:
No, you understanding is correct, but if you'd extract the config reducer and move it to the root of the jsonforms reducer hierarchy, the approach should work (which I like better anyway since the common reducer is less complex and concerns are separated).

PR #810, opened 2018-01-22
Updated Webpack config to be modular

eneufeld:
* Using webpack-merge to allow modular webpack configs
* Defined externals to separate JSONForms Core and Renderers


packages/core/webpack.build.js
```diff
@@ -0,0 +1,9 @@
+const merge = require('webpack-merge');
+const baseConfig = require('../../webpack/webpack.build.base.js');
+
+module.exports = merge(baseConfig, {
+    output: {
+        filename: "jsonforms-core.js",
+        library: "JSONForms-Core"
```
edgarmueller: As discussed, remove hyphens.

PR #811, opened 2018-01-23
Remove $ref property from Scopable

### LINKED ISSUE #803: Remove $ref property from Scopable
We can shorten the definition of a Scopable by inlining the value of the $ref.

edgarmueller:
Change `Scopable` to interface with a property `scope` of type `string`.
Also remove obsolete test files which were found while updating the scope defintions.

Fixes #803


edgarmueller:
I did, why?

PR #814, opened 2018-01-24
Extract JSON Forms state into separate substate

### LINKED ISSUE #812: Move jsonforms state to jsonforms key
Reducers as well as extractors should act on the `state.jsonforms` substate so that usage when having a redux store already is eased. 
Check whether functions like `initJsonFormsStore` or `createJsonFormsStore` are still needed.

edgarmueller:
Fixes #812 
PR also contains changes to properly make re-use of the test package and removes/simplifies imports within tests.


packages/core/src/renderers/field.util.ts
```diff
@@ -55,7 +54,7 @@ export const mapStateToFieldProps = (state, ownProps) => {
   const path = composeWithUi(ownProps.uischema, ownProps.path);
   const visible = _.has(ownProps, 'visible') ? ownProps.visible : isVisible(ownProps, state);
```
eneufeld: must this not be state.jsonforms?
also the next line?

packages/core/src/store.ts
```diff
@@ -1,34 +1,23 @@
...
-    jsonformsReducer(),
-    initialState,
-    applyMiddleware(thunk)
-  );
-
-  return store as JsonFormsStore;
-};
 export interface JsonFormsStore extends Store<any> {
 }
 export interface JsonFormsState {
```
eneufeld: can't we simply define the JSONForms sub state, so leave it as it was? What is the use case of a full JsonformsState?

packages/core/src/renderers/field.util.ts
edgarmueller: No, since `isVisible` and `isEnabled` do not extract anything directly from the state, only via `getData` which expects the root state.

packages/core/src/store.ts
edgarmueller: No sure if I understood the question correctly. Is it about the index properties?

PR #832, opened 2018-01-27
Fixed initialization of store

### LINKED ISSUE #827: Generate schema example has undefined UI schema
UI schema is undefined and needs to be set.

### LINKED ISSUE #829: Validation errors are not being shown
see person example and minLength constrain on name, for instance.

eneufeld:
* fixed initialization of store in the webcomponent
* fixed resolve example

Fixes #827, Fixes #829


edgarmueller:
Thanks, looks good, doesn't fix #827 though

PR #852, opened 2018-02-03
Port of #779 to master

edgarmueller:
Port of i18n from `react` to `master` branch

### DIFF
--- a/packages/core/src/util/label.ts
+++ b/packages/core/src/util/label.ts
@@ -55,11 +55,18 @@ export const createLabelDescriptionFrom = (withLabel: ControlElement): LabelDesc
   }
 };
 
-export const translateLabel = (translations, label: LabelDescription): LabelDescription => {
-  if (translations && _.startsWith(label.text, '%')) {
-    const labelKey = label.text.substr(1, label.text.length);
-    label.text = translations[labelKey] ? translations[labelKey] : label.text;
-  }
+export interface Translations {
+  [key: string]: string;
+}
 
-  return label;
-};
+export const translateLabel: (translations: Translations,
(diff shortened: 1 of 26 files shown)


packages/vanilla/package.json
```diff
@@ -18,7 +18,6 @@
   },
   "dependencies": {
     "@jsonforms/core": "^2.0.0-rc.1",
-    "@jsonforms/webcomponent": "^2.0.0-rc.1",
```
eneufeld: must be a dev dependency at least, as otherwise the examples won't run as all the example still rely on the webcomponent and not react

packages/core/src/util/label.ts
```diff
@@ -54,3 +54,12 @@ export const createLabelDescriptionFrom = (withLabel: ControlElement): LabelDesc
     };
   }
 };
+
+export const translateLabel = (translations, label: LabelDescription): LabelDescription => {
```
eneufeld: type translations?

packages/core/src/util/label.ts
```diff
@@ -54,3 +54,12 @@ export const createLabelDescriptionFrom = (withLabel: ControlElement): LabelDesc
...
+  if (translations && _.startsWith(label.text, '%')) {
+    const labelKey = label.text.substr(1, label.text.length);
+    label.text = translations[labelKey] ? translations[labelKey] : label.text;
```
eneufeld: Should we bot at least remove the % and apply the default transformation?

edgarmueller: Jonas argued that we should stick with this behaviour in order to make it obvious for devs that a translation is missing. Personally, I'm fine with both approaches.

PR #857, opened 2018-02-06
Add JSDoc & refine typings

### LINKED ISSUE #698: Finalize API typings & JSDoc
Some parts of the code are not typed at all or only weakly typed. We need to make sure that at least all API relevant code is typed.

edgarmueller:
Fixes #698

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -2,9 +2,17 @@
  * Global JSONForms object that holds services and registries.
  */
 export class JsonFormsGlobal {
+  /**
+   * Array of renderers that should be picked up at startup.
+   * @type {any[]}
+   */
   public renderers = [];
+
+  /**
+   * Array of field renderers that should be picked up at startup.
+   * @type {any[]}
+   */
   public fields = [];
-  public reducers: {[key: string]: any} = {};
 }
 
 const JsonForms = new JsonFormsGlobal();
--- a/packages/core/src/renderers/UnknownRenderer.tsx
+++ b/packages/core/src/renderers/UnknownRenderer.tsx
@@ -1,8 +1,20 @@
 import * as React from 'react';
-import {Component} from "react";
+import { Component } from 'react';
 
-export class UnknownRenderer extends Component<any, any> {
+/**
+ * Props of an {@link UnknownRenderer}
+ */
+export interface UnknownRendererProps {
+  /**
+   * The type for which no renderer has been found.
+   */
+  type: 'renderer' | 'field';
+}
 
+/**
+ * A renderer that will be used in case no other renderer is applicable.
+ */
+export class UnknownRenderer extends Component<UnknownRendererProps, any> {
   render() {
     return (
       <div style={{color: 'red'}}>
--- a/packages/core/src/util/field.ts
+++ b/packages/core/src/util/field.ts
@@ -1,47 +1,24 @@
-import { Component } from 'react';
 import * as _ from 'lodash';
-import { ControlElement, UISchemaElement } from '../models/uischema';
+import { ControlElement } from '../models/uischema';
 import { RankedTester } from '../testers';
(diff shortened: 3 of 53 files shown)


packages/core/src/core.ts
```diff
@@ -2,8 +2,22 @@
...
+
+  /**
+   * Array of field renderers that should be picked up at startup.
+   * @type {any[]}
+   */
   public fields = [];
+
+  /**
+   * Object containing additional reducers.
+   * @type {{}}
```
eneufeld: shouldn't type be the type of the method?

packages/core/src/renderers/UnknownRenderer.tsx
```diff
@@ -1,8 +1,20 @@
...
 
-export class UnknownRenderer extends Component<any, any> {
+/**
+ * Props of an {@link UnknownRenderer}
+ */
+export interface UnknownRendererProps {
+  /**
+   * The type for which no renderer has been found.
+   */
+  type: 'renderer' | 'field';
```
eneufeld: do we use this prop anywhere?

packages/test/src/FakeLayout.tsx
```diff
@@ -49,7 +49,11 @@ const FakeLayout = (props: RendererProps) => {
   );
 };
 
-export default registerStartupRenderer(
+const ConnectedFakeLayout = connectToJsonForms(mapStateToLayoutProps, null)(FakeLayout);
+
+registerStartupRenderer(
     fakeLayoutTester,
-    connectToJsonForms(mapStateToLayoutProps, null)(FakeLayout)
+    FakeLayout
```
eneufeld: this should be the connectedFakeLayout

packages/core/src/util/field.ts
```diff
@@ -1,47 +1,24 @@
...
-  enabled: boolean;
-  dispatch: any;
-  isValid: boolean;
-
-  handleChange(string, any): (void);
-}
+/**
+ * Registers the given field renderer when a JSON Forms store is created.
+ * @param {RankedTester} tester
+ * @param field
```
eneufeld: doc uncomplete

packages/material/src/complex/index.ts
```diff
@@ -5,7 +5,9 @@ import {
 } from '@jsonforms/core';
 import MaterialArrayControlRenderer from './MaterialArrayControlRenderer';
 
-export default registerStartupRenderer(
+registerStartupRenderer(
     rankWith(3, isArrayObjectControl),
     MaterialArrayControlRenderer
```
eneufeld: shouldn't this be connected?

packages/vanilla/src/complex/array/index.ts
```diff
@@ -11,7 +11,9 @@ export {
   ArrayControlRenderer
 };
 
-export default registerStartupRenderer(
+registerStartupRenderer(
   rankWith(2, isArrayObjectControl),
   ArrayControlRenderer
```
eneufeld: why not connected?

packages/core/src/core.ts
edgarmueller: I've removed the `reducers` property, since it is not use anyway.

packages/core/src/renderers/UnknownRenderer.tsx
edgarmueller: Yes, check `DispatchField` and `DispatchRenderer`.

packages/material/src/complex/index.ts
edgarmueller: We're using the default export of the `MaterialArrayControlRenderer` which is connected.

packages/vanilla/src/complex/array/index.ts
edgarmueller: Same as with the material array renderer, default export is already connected.

PR #862, opened 2018-02-08
Remove registerStartupRenderer and replace with explicit initialization

edgarmueller:
* Remove JsonFormsInit
* Use `connectToJsonForms` consistently, make 'react-redux` import obsolete
* make `react-redux` dependency a dev dependency in renderer sets and remove react-dom dependency from material renderer set
* Streamline naming of testers


edgarmueller:
@eneufeld this should be fine now, can you approve?

PR #864, opened 2018-02-08
Re-order controls in person example

### LINKED ISSUE #846: Fix UI schema of person example

edgarmueller:
Fixes #846

### DIFF
--- a/packages/examples/src/person.ts
+++ b/packages/examples/src/person.ts
@@ -79,28 +79,19 @@ export const uischema = {
         },
         {
           type: 'Control',
-          label: '%height',
-          scope: '#/properties/personalData/properties/height'
-        },
-        {
-          type: 'Control',
-          label: {
-            text: '%age'
-          },
-          scope: '#/properties/personalData/properties/age'
-        },
-        {
-          type: 'Control',
-          scope: '#/properties/personalData/properties/drivingSkill',
-          label: {
-            text: '%drivingskill'
-          }
+          label: '%birthday',
+          scope: '#/properties/birthDate'
         },
       ]
     },
     {
       type: 'HorizontalLayout',
       elements: [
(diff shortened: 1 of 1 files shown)


packages/examples/src/person.ts
```diff
@@ -79,22 +79,8 @@ export const uischema = {
...
-        },
-        {
-          type: 'Control',
-          label: {
-            text: '%age'
-          },
-          scope: '#/properties/personalData/properties/age'
-        },
-        {
-          type: 'Control',
```
eneufeld: we should at least move the driving skill example to an own example eg "numbers" to demonstrate different number behaviors

packages/examples/src/person.ts
```diff
@@ -103,39 +89,20 @@ export const uischema = {
...
-          scope: '#/properties/birthDate'
-        }
-      ]
-    },
-    {
-      type: 'HorizontalLayout',
-      elements: [
-        {
-          type: 'Control',
-          scope: '#/properties/postalCode',
```
eneufeld: same here, we should have an example which shows the use of trim and restrict

edgarmueller:
I agree with both points in general, but perhaps not in scope of this PR:
* I've removed `drivingSkil` because I think it doesn't make much sense in the context of a person. A dedicated `numbers` example is a good idea, but perhaps as a separate PR then.
* Demonstrating trim and restrict without any explanation might not be ideal. I'd probably also move this to a text/string section together with some text that explains what's happening.

I can create issues for both points if you agree. 
I think the person example should demonstrate the basic feature set of JSON Forms and hence I'd stick with how it is structured right now.

edgarmueller:
Created https://github.com/eclipsesource/jsonforms/issues/865

PR #868, opened 2018-02-10
Pull out react into own package

eneufeld:
* pull out react into an own package
* update all dependencies in existing packages accordingly
* create angular package
* create an angular-material package for an angular renderer set based on material


eneufeld:
Thanks for the review.

> there one issue I've seen: when running npm run dev and modifying any of the two text fields twice I get an error ERROR TypeError: Cannot read property 'unsubscribe' of undefined.

Did you run npm run dev for the angular-material project?

edgarmueller:
Yes.

PR #879, opened 2018-02-13
Add i18n description translation

edgarmueller:


edgarmueller:
Yeah, good point, working on it..

PR #880, opened 2018-02-14
Add i18n number formatting

edgarmueller:


packages/core/src/renderers/Formatted.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Interface for mapping a given type to a formatted string and back.
+ */
+export interface Formatted<A> {
+  /**
+   * Format the given value
+   *
+   * @param {A} a the value to be formatted
```
eneufeld: why not name the variable "value"?

PR #884, opened 2018-02-15
Introduce config sub-state and reducer

edgarmueller:


packages/core/src/reducers/config.ts
```diff
@@ -0,0 +1,22 @@
+import * as _ from 'lodash';
+import { SET_CONFIG } from '../actions';
+import { configDefault } from '../configDefault';
+
+const applyDefaultConfiguration = (config?: any) => {
```
eneufeld: can't we default the config to an empty object and always merge?

packages/examples/src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+  'en-US': {
+    postalcode: 'Postal Code'
+  },
+  'de-DE': {
+    postalcode: 'Postleitzahl'
+  }
+};
+
+const config = {
+  restrict: false,
```
eneufeld: we should use non default values to show off the config

edgarmueller: should be both `true`

packages/core/src/util/renderer.ts
```diff
@@ -139,6 +140,7 @@ export const mapStateToControlProps = (state, ownProps): StatePropsOfControl =>
       controlElement.scope !== undefined && isRequired(ownProps.schema, controlElement.scope);
   const resolvedSchema = Resolve.schema(ownProps.schema, controlElement.scope);
   const description = resolvedSchema !== undefined ? resolvedSchema.description : '';
+  const config = getConfig(state);
```
eneufeld: didn't we want to merge the config of the state with the options of the uischema?

packages/core/src/util/field.ts
```diff
@@ -27,6 +27,7 @@ export const mapStateToFieldProps = (state, ownProps): StatePropsOfField => {
   const controlElement = ownProps.uischema as ControlElement;
   const id = controlElement.scope || '';
   const inputClassName = ['validate'].concat(isValid ? 'valid' : 'invalid');
+  const config = getConfig(state);
```
eneufeld: didn't we want to merge the config of the state with the options of the uischema?

PR #889, opened 2018-02-19
Reduce linter warnings

edgarmueller:
* Some linter rules updated
* Fix most remaining warnings


packages/core/src/actions/index.ts
```diff
@@ -6,13 +6,19 @@ import { generateDefaultUISchema, generateJsonSchema } from '../generators';
 const NAMESPACE = 'jsonforms';
 
 export const INIT = `${NAMESPACE}/INIT`;
-export const UPDATE_DATA = `${NAMESPACE}/UPDATE`;
+export const UPDATE_DATA: 'jsonforms/UPDATE' = 'jsonforms/UPDATE';
```
eneufeld: why not use the Namespace constant?

packages/core/src/actions/index.ts
```diff
@@ -6,13 +6,19 @@ import { generateDefaultUISchema, generateJsonSchema } from '../generators';
...
 export const VALIDATE = `${NAMESPACE}/VALIDATE`;
 export const ADD_RENDERER = `${NAMESPACE}/ADD_RENDERER`;
 export const REMOVE_RENDERER = `${NAMESPACE}/REMOVE_RENDERER`;
 export const ADD_FIELD = `${NAMESPACE}/ADD_FIELD`;
 export const REMOVE_FIELD = `${NAMESPACE}/REMOVE_FIELD`;
 
+export interface UpdateAction {
+  type: 'jsonforms/UPDATE';
```
eneufeld: why not use UPDATE_DATA constant?

Issue #901, opened 2018-02-22
Update make-it-happen example


edgarmueller:
The name property should also have a `minLength` restriction.

eneufeld:
Closed via #903

PR #903, opened 2018-03-05
Updates examples

edgarmueller:
* Add minLength constraint to name property of task which demonstrates validation better in the blog series (instead of name only being required)
* Replace description in person example to avoid displaying label


eneufeld:
Fixed #901

Issue #894, opened 2018-02-20
Material array renderer issues


edgarmueller:
Two minor problems I encountered with Stefan yesterday when looking at the material renderer:
* When having a boolean control inside the array and having the 'select all' state checked, clicking the boolean control inside the array clears the toggled state of the 'select all' checkbox
* Labels should be capitalized

PR #904, opened 2018-03-07
Switch to jest for material renderers, add tests for array

edgarmueller:
This was initially motivated because of popper.js, which throws exception about Range being missing. Popper is used by the material tooltips.

### DIFF
--- a/packages/material/src/complex/MaterialArrayControlRenderer.tsx
+++ b/packages/material/src/complex/MaterialArrayControlRenderer.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-
+import * as _ from 'lodash';
 import {
   mapDispatchToTableControlProps,
   mapStateToTableControlProps,
@@ -20,17 +20,15 @@ import Grid from 'material-ui/Grid';
 export class MaterialArrayControlRenderer extends RendererComponent<TableControlProps, TableState> {
   constructor(props) {
     super(props);
-
     this.state = {
-      selected: [],
+      selected: _.fill(Array(this.props.data.length), false),
       openConfirmDelete: false
     };
   }
 
   render() {
     const { visible } = this.props;
-
-    const numSelected = this.state.selected ? this.state.selected.length : 0;
+    const numSelected = this.state.selected ? _.filter(this.state.selected, v => v).length : 0;
     const tableProps = {
       selectAll: this.selectAll,
       select: this.select,
@@ -44,6 +42,7 @@ export class MaterialArrayControlRenderer extends RendererComponent<TableControl
       numSelected,
       ...this.props
     };
+    const selectedCount = _.filter(this.state.selected, v => v).length;
 
     return (
       <Grid container direction='column' hidden={{ xsUp: !visible }} spacing={0}>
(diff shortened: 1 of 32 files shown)


packages/material/src/complex/MaterialArrayControlRenderer.tsx
```diff
@@ -81,33 +80,18 @@ export class MaterialArrayControlRenderer extends RendererComponent<TableControl
...
   }
   private selectAll = (_event, checked) => {
     if (checked) {
-      this.setState({ selected: this.props.data });
-
+      this.setState({ selected: _.fill(Array(this.props.data.length), true) });
       return;
     }
-    this.setState({ selected: [] });
+    this.setState({selected: _.fill(Array(this.props.data.length), false)});
```
eneufeld: can't we move the _.fill (...) stuff to a function so that we don't repeat it?

packages/material/src/complex/MaterialArrayControlRenderer.tsx
```diff
@@ -81,33 +80,18 @@ export class MaterialArrayControlRenderer extends RendererComponent<TableControl
...
-    } else if (selectedIndex === selected.length - 1) {
-      newSelected = newSelected.concat(selected.slice(0, -1));
-    } else if (selectedIndex > 0) {
-      newSelected = newSelected.concat(
-        selected.slice(0, selectedIndex),
-        selected.slice(selectedIndex + 1),
-      );
-    }
+  private select = (_event, index) => {
+    const copy = this.state.selected.slice();
```
eneufeld: Does this fix the select issue described in  #894 ?

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -33,7 +33,7 @@ const EmptyTable = () => (
 );
 
 const TableHeaderCell = ({cellProperty}) =>
-  <TableCell >{cellProperty}</TableCell>;
+  <TableCell >{_.capitalize(cellProperty)}</TableCell>;
```
eneufeld: Does this fix the Label issue from #894?

packages/material/test/renderers/MaterialBooleanField.test.tsx
```diff
@@ -1,378 +1,324 @@
...
-  store.dispatch(update('foo', () => null));
-  t.is(input.value, '');
-});
 
-test('update with wrong ref', t => {
-  const store = initJsonFormsStore({
-    data: t.context.data,
-    schema: t.context.schema,
-    uischema: t.context.uischema,
+  test.skip('should update with undefined value', () => {
```
eneufeld: why skip all those tests?

packages/material/src/complex/MaterialArrayControlRenderer.tsx
edgarmueller: Yes, the whole PR was actually meant to fix #894. I forgot to mention that.

packages/material/src/complex/MaterialTableControl.tsx
edgarmueller: Yes.

packages/material/test/renderers/MaterialBooleanField.test.tsx
edgarmueller: See follow up issue #907

Issue #820, opened 2018-01-25
Configure coveralls


edgarmueller:
Since we have multiple packages I have no clue about how to configure coveralls integration in order to enable github checks. Someone (preferrably not me :wink:) should investigate.

edgarmueller:
Fixed via #905

PR #905, opened 2018-03-07
Merge coverage reports

### LINKED ISSUE #687: Check test coverage
Check test coverage and check whether there are areas where we can increase test coverage.

edgarmueller:
Fixed #687 and #820

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ docs
 
 .idea/
 .vscode/
+
+.coveralls.yml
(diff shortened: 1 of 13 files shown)


.gitignore
```diff
@@ -8,3 +8,5 @@ docs
 
 .idea/
 .vscode/
+
+.coveralls.yml
```
eneufeld: why ignore the coveralls file, doesn't it contain config?

edgarmueller: It's used to run coveralls locally and contains the repo token, which should not be made public

PR #919, opened 2018-03-14
Updated Readmes

### LINKED ISSUE #909: Review all READMEs
We pulled out react and must thus update the readme and the docu

eneufeld:
* Updated existings readmes to reflect pull out of react from core
* Added missing Readmes

Fixes #909

### DIFF
--- a/packages/angular-material/README.md
+++ b/packages/angular-material/README.md
@@ -1,7 +1,7 @@
 # JSONForms - More Forms. Less Code
 ### Complex Forms in the blink of an eye
 
-JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
+JSONForms eliminates the tedious task of writing fully-featued forms by hand by leveraging the capabilities of JSON, JSON Schema and Javascript.
 
 # Angular Material Renderers Package
 This is the JSONForms Angular Material Renderers Package. This package only contains renderers and must be combined with  [JSONForms Angular](https://github.com/eclipsesource/jsonforms/blob/master/packages/angular).
--- a/packages/angular-material/README.md
+++ b/packages/angular-material/README.md
@@ -3,16 +3,10 @@
 
 JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
 
-# Core Package
-This is the JSONForms core package. This package provides basic functionality needed to render forms and is based on [Redux](https://redux.js.org/).
+# Angular Material Renderers Package
+This is the JSONForms Angular Material Renderers Package. This package only contains renderers and must be combined with  [JSONForms Angular](https://github.com/eclipsesource/jsonforms/blob/master/packages/angular).
 
-In order to use JSONForms Core you need to decide which UI framework you would like to use.
-
-JSONForms currently supports [React](https://www.npmjs.com/package/@jsonforms/react). A proof of concept exists for [Angular](https://github.com/eclipsesource/jsonforms/blob/master/packages/angular).
-
-See the [Example Package](https://www.npmjs.com/package/@jsonforms/examples) on how to integrate JSONForms with your application.
-
-Check https://www.npmjs.com/search?q=%40jsonforms for all published JSONForms packages.
+**This is a proof of concept currently and will evolve in the future.**
 
 # License
 The JSONForms project is licensed under the MIT License. See the [LICENSE file](https://github.com/eclipsesource/jsonforms/blob/master/LICENSE) for more information.
--- a/packages/angular/README.md
+++ b/packages/angular/README.md
@@ -4,9 +4,9 @@
 JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
 
 # Angular Package
-This is the JSONForms angular package which provides the necessary bindings for Angular. It uses [JSONForms Core](https://www.npmjs.com/package/@jsonforms/core).
+This is the JSONForms Angular package which provides the necessary bindings for Angular. It uses [JSONForms Core](https://www.npmjs.com/package/@jsonforms/core).
 
-You can combine the angular package with any angular-based renderer set you want, for example the [Material Renderers](https://github.com/eclipsesource/jsonforms/blob/master/packages/angular-material).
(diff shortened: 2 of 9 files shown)


eneufeld:
It would probably make sense if @edgarmueller also checks the readmes

packages/angular-material/README.md
```diff
@@ -0,0 +1,22 @@
+# JSONForms - More Forms. Less Code
+### Complex Forms in the blink of an eye
+
+JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
```
edgarmueller: This sentence actually isn't entirely true anymore since we only use custom elements in the context of the webcomponent package. Maybe we should update the wording here (this need to be updated across all READMEs, I think).

packages/angular/README.md
```diff
@@ -0,0 +1,24 @@
+# JSONForms - More Forms. Less Code
+### Complex Forms in the blink of an eye
+
+JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
+
+# Angular Package
+This is the JSONForms angular package which provides the necessary bindings for Angular. It uses [JSONForms Core](https://www.npmjs.com/package/@jsonforms/core).
```
edgarmueller: casing: Angular

packages/angular/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+You can combine the angular package with any angular-based renderer set you want, for example the [Material Renderers](https://github.com/eclipsesource/jsonforms/blob/master/packages/angular-material).
```
edgarmueller: casing again; not the same as in angular-material

packages/core/README.md
```diff
@@ -4,9 +4,11 @@
 JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
 
 # Core Package
-This is the JSONForms core package which does not contain any renderers by itself.
+This is the JSONForms core package. This package provides basic functionality needed to render forms and is based on [Redux](https://redux.js.org/).
 
-You can combine the core package with any renderer you want, for example the [Material Renderers](https://www.npmjs.com/package/@jsonforms/material-renderers) or the [Vanilla Renderers](https://www.npmjs.com/package/@jsonforms/vanilla-renderers).
+In order to use JSONForms Core you need to decide which UI framework you would like to use.
+
+JSONForms currently supports [React](https://www.npmjs.com/package/@jsonforms/react). A proof of concept exists for [Angular](https://github.com/eclipsesource/jsonforms/blob/master/packages/angular).
```
edgarmueller: Should we mention the Angular package as it's private?

packages/test/README.md
```diff
@@ -0,0 +1,24 @@
+# JSONForms - More Forms. Less Code
+### Complex Forms in the blink of an eye
+
+JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
+
+# Test Package
+This is the JSONForms test package. It can be used for testing packages which do not care about a specific renderer set being used, e.g. the i18n or the webcomponent package.
+
+This package provides:
+* FakeLayout which wraps all children into a `div` with the className `layout`.
```
edgarmueller: use `FakeLayout` and add comma before which, likewise for `FakeControl`

packages/i18n/README.md
```diff
@@ -0,0 +1,26 @@
+# JSONForms - More Forms. Less Code
+### Complex Forms in the blink of an eye
+
+JSONForms uses HTML custom elements and eliminates the need to write HTML templates in order to create forms by leveraging the capabilities of JSON and JSON schema.
+
+# I18n Package
+This is the JSONForms i18n package. This package provides reducers and actions to support i18n.
```
edgarmueller: Off-topic, but I'd propose to make this package private for the 2.0.0 release as it hasn't been tested thoroughly and it might be the case that we need to update the API.

edgarmueller: @eneufeld Do you agree?

eneufeld: agreed

packages/core/README.md
eneufeld: I mentioned it as a proof of concept, but this can be removed, @koegel  what is your opinion?

packages/angular-material/README.md
edgarmueller: Maybe something along the line of: JSONForms eliminates the tedious task of writing fully-featued forms by hand by leveraging the capabilities of JSON, JSON Schema and Javascript? @koegel Your take on this?

packages/i18n/README.md
edgarmueller: Ok, I'll submit a PR later today

packages/core/README.md
koegel: I think mentioning Angular even only as POC is good

packages/i18n/README.md
koegel: +1

packages/core/README.md
samtsai: Hi JSONForms team! TBH, I didn't realize Angular package was private until reading this thread, even if it is a POC. It took me a moment to realize my working example was using the `material` renderer which under the hood is using the `material-ui` library.

Maybe an extra note of `private` or unreleased would be helpful for others so they don't get confused with it not showing up in `npm`: https://www.npmjs.com/search?q=%40jsonforms

eneufeld: Hi @samtsai,
thanks for pointing this out.
We will update the Readme of the Angular package accordingly.

Just to clarify:
You used the angular package and the angular-material packages right? Because both of those have no react dependencies as the angular-material package is based on [angular-material](https://material.angular.io/).

samtsai: Hi @eneufeld,
I wanted to install the packages via yarn/npm but didn't find them in the registry so for now I'm basing my example off of your example repo: https://github.com/eclipsesource/make-it-happen-angular (which is still using `material-renderer` not `angular-material-renderer`).

eneufeld: Ok, 
now I get it.

The example repo you use is simply using the Webcomponent to wrap the react based application.
The ugly part about this solution is, that you have to use react if you want to provide custom renderers for your application.
This is why we created the angular and angular-material packages.
The angular-material package also contains an example application you can start. The package doesn't contain many renderers yet though.

We are always happy to accept contributions.

samtsai: Thanks for the update! I forked the project and will take a look, hopefully I can contribute something useful :)

PR #927, opened 2018-03-22
Angular Boolean renderer

samtsai:
@eneufeld I've been working on the `angular-material` renderers and I've added a `boolean` renderer and set up some other things:

- Redux devtools support
- lint rule updates
- grammar fixes

I've noticed I get an `undefined` error in the console when changing the state of the data. Functionally everything works fine but it seems that there are some cases where `ngOnDestroy` is called where `this.subscription` no longer exists... I'm really not sure why.

```
ERROR TypeError: Cannot read property 'unsubscribe' of undefined
    at TextControlRenderer.ngOnDestroy (text.renderer.ts:100)
    at callProviderLifecycles (core.js:12712)
    at callElementProvidersLifecycles (core.js:12673)
    at callLifecycleHooksChildrenFirst (core.js:12656)
    at destroyView (core.js:14022)
    at callWithDebugContext (core.js:15056)
    at Object.debugDestroyView [as destroyView] (core.js:14607)
    at ViewContainerRef_.clear (core.js:11363)
    at SafeSubscriber._next (jsonforms.component.ts:69)
    at SafeSubscriber.__tryOrUnsub (Subscriber.js:239)
```

Love to hear your feedback on all this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,8 @@ For a more detailed tutorial about the usage of JSON Forms, please see [this tut
 * Install [node.js](https://nodejs.org/)(version > 4.x.x)
 * Clone this repository
 * Install dependencies: `npm install`
-* Hook up dependencies between packages: `lerna bootstrap --hoist`
+* Hook up dependencies between packages: `lerna bootstrap --hoist`  
+  or `npx lerna bootstrap --hoist` (if npm >= 5.2.0 and don't want to globally install `lerna`)
 
 ## Build & Testing
 * Build (all packages): `lerna run build`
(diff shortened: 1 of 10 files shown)


eneufeld:
Thank you for the contribution.
I encountered the same error.

I didn't look into the error yet, but I checked the examples in the angular-redux project and they don't call destroy at all.
And I just checked the angular-redux project and with version 4 they provide a connect method which is even more similar to react now.
Maybe it would make sense to give it a try, as currently the ngOnInit has a lot of code duplication.

PR #928, opened 2018-03-26
Fixed redux store binding in angular

eneufeld:
* Updated angular dependencies
* Fixed store binding by adding explicit comparators

By providing the comparator the number of recreations is reduced significantly.
The comparators are needed as we map the whole store, would we use selects to retrieve single properties we could get rid of the comparators.


edgarmueller:
@eneufeld I'm afraid you need to update this one once more

PR #935, opened 2018-04-10
Add explicit examples for numbers and text controls

### LINKED ISSUE #865: Create additional examples
Create one or more examples that feature more advanced controls and options, e.g.
* `trim` and `restrict` options of text-based controls and describe the behaviour
* slider control and when how it is triggered

AhmetTanakol:
Added new example files to separate examples for numbers and texts.

Fix #865

### DIFF
--- a/packages/examples/src/numbers.ts
+++ b/packages/examples/src/numbers.ts
@@ -5,9 +5,15 @@ export const schema = {
   properties: {
     price: {
       type: 'number',
-      maximum: 300,
+      maximum: 100,
       minimum: 1,
       default: 50
+    },
+    age: {
+      type: 'integer'
+    },
+    height: {
+      type: 'number'
     }
   },
 };
@@ -25,6 +31,14 @@ export const uischema = {
             text: 'Price'
           }
         },
+        {
+          type: 'Control',
+          scope: '#/properties/age'
+        },
+        {
+          type: 'Control',
+          scope: '#/properties/height'
+        }
       ]
     },
   ]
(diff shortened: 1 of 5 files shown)


packages/examples/src/text.ts
```diff
@@ -0,0 +1,73 @@
...
+export const uischema = {
+  type: 'VerticalLayout',
+  elements: [
+    {
+      type: 'HorizontalLayout',
+      elements: [
+        {
+          type: 'Control',
+          scope: '#/properties/address',
+          label: 'Address',
```
edgarmueller: Explicitly point that this is using `trim` and `restrict` options

packages/examples/src/text.ts
```diff
@@ -0,0 +1,73 @@
+import { registerExamples } from './register';
+
+export const schema = {
```
edgarmueller: Maybe we can pick a different example, since an address is not typically entered in a single text field. Also reduce the `maxLength` to a size of 10 which is beneficial when playing around with the examples.

packages/examples/src/numbers.ts
```diff
@@ -0,0 +1,43 @@
+import { registerExamples } from './register';
+
+export const schema = {
+  type: 'object',
+  properties: {
+    price: {
+      type: 'number',
+      maximum: 300,
```
edgarmueller: Lower to 100

packages/examples/src/numbers.ts
```diff
@@ -0,0 +1,43 @@
...
+      minimum: 1,
+      default: 50
+    }
+  },
+};
+
+export const uischema = {
+  type: 'VerticalLayout',
```
edgarmueller: Add controls for integer and number as well

AhmetTanakol:
updated

edgarmueller:
Thank you

PR #947, opened 2018-04-27
Update ajv to 6.4.0

edgarmueller:
Update ajv to most recent version: `validate` is also capable of returning a `PromiseLike` object now, it seems.


eneufeld:
Don't we need to explicitly add the v04 schema like stated in the ajv readme?

> To use Ajv with draft-04 schemas in addition to explicitly adding meta-schema you also need to use option schemaId:
> 
> var ajv = new Ajv({schemaId: 'id'});
> // If you want to use both draft-04 and draft-06/07 schemas:
> // var ajv = new Ajv({schemaId: 'auto'});
> ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-04.json'));

edgarmueller:
I guess ajv defaults to v7 and it still works as we likely do not use any features which are specific to version 4. I'll add another commit that adds the registers call and the schemaId.

edgarmueller:
Coverage decreased as I've included the draft4 schema as an object literal instead of requiring a loader to handle JSON files.

PR #952, opened 2018-05-02
[Editor] Improve label mapping capabilities

lucas-koehler:
The values of a label mapping can now be either specified as a string or as an object.
In case of a string the behavior does not change. This guarantees old label mappings to behave as always.
The object allows to specify two properties:
`constant`: A constant value that is always displayed
`property`: A property whose content is displayed
If both are specified, the constant value is displayed followed by a space and the property value.

Additionally fixed an import to be resolvable when the editor package is consumed in other packages.


edgarmueller:
Likely, but only later as we tried to get rid of the assumptions we make about ids, hence I'm ok with the change.

PR #957, opened 2018-05-03
support references on uischema generation

AhmetTanakol:


packages/core/src/generators/uischema.ts
```diff
@@ -121,12 +128,22 @@ const generateUISchema =
             if (!_.isEmpty(jsonSchema.properties)) {
                 // traverse properties
                 const nextRef: string = currentRef + '/properties';
+                const nextRelativePath: string = relativePath + '/properties';
                 Object.keys(jsonSchema.properties).map(propName => {
-                    const value = jsonSchema.properties[propName];
+                    let value = jsonSchema.properties[propName];
+                    relativePath = `${nextRelativePath}/${propName}`;
+                    let ref = `${nextRef}/${propName}`;
+                    if (value.$ref !== undefined &&
```
edgarmueller: Why do we only resolve refs in case we also have a definitions block? Since we include json-refs anyway, we handle this case in a more generic fashion I think

packages/core/test/generators/uischema.test.ts
```diff
@@ -27,6 +27,153 @@ import { generateDefaultUISchema } from '../../src/generators/uischema';
...
+  const uischema: VerticalLayout = {
+    type: 'VerticalLayout',
+    elements: [
+      typeControl,
+      labelControl,
+      scopeControl,
+      ruleLayout
+    ]
+  };
+  const uiSchema = generateDefaultUISchema(schema);
```
edgarmueller: Rename `uiSchema` to `generatedUiSchema`

packages/editor/src/ide.ts
```diff
@@ -61,33 +62,41 @@ export class JsonEditorIde extends HTMLElement implements Editor {
...
+          configureUploadButton(this, uploadButton);
+        }
+
+        // configure button to download model data.
+        const downloadButton = document.getElementById('download-data-button') as HTMLButtonElement;
+        if (downloadButton !== null) {
+          configureDownloadButton(this, downloadButton);
+        }
+      },
+             err => {
```
edgarmueller: Fix formatiting

packages/core/src/generators/uischema.ts
AhmetTanakol: I think we still need it, because json-refs doesn't resolve all the references such as circular references. We need to specify where to look for their definitions. As you said, for inlined references we don't need this.

AhmetTanakol:
Updated

packages/editor/src/ide.ts
```diff
@@ -61,33 +62,42 @@ export class JsonEditorIde extends HTMLElement implements Editor {
...
+    JsonRefs.resolveRefs(getSchema(this._store.getState()))
+      .then(
+        resolvedSchema => {
+        this.editor.setInnerComponent(
+          TreeRenderer,
+          {
+            uischema: getUiSchema(this._store.getState()),
+            schema: getSchema(this._store.getState()),
+            schemaService,
+            resolved: resolvedSchema.resolved
```
edgarmueller: Don't we always want to work with the resolved schema? So what happens if we use `resolvedSchema.resolved` as `schema`, do we still need the `resolved` property?

AhmetTanakol: I don't think so, they are same so we can directly use `resolvedSchema.resolved` as our root schema

edgarmueller: Could you check if that's possible and update if that's the case?

AhmetTanakol:
Updated

edgarmueller:
Thank you

PR #960, opened 2018-05-04
Add support for primitive types to array renderers

edgarmueller:
Adds support for primitive array types. Following changes have been made:
* Refactor typings of vanilla table controls
* Adds basic test case for vanilla `ArrayControl`
* Fix signature of `addItem` of table dispathc props
* Adds missing clone in core reducer if target path is empty
* Adds new example for demo purposes

### DIFF
--- a/packages/vanilla/src/complex/TableArrayControl.tsx
+++ b/packages/vanilla/src/complex/TableArrayControl.tsx
@@ -26,7 +26,6 @@ import * as React from 'react';
 import * as _ from 'lodash';
 import {
   ControlElement,
-  ControlProps,
   formatErrorMessage,
   Helpers,
   isPlainLabel,
@@ -34,6 +33,7 @@ import {
   mapStateToTableControlProps,
   Paths,
   RankedTester,
+  TableControlProps,
   Test,
 } from '@jsonforms/core';
 import { connectToJsonForms, DispatchField, RendererComponent } from '@jsonforms/react';
@@ -45,7 +45,9 @@ const {
 } = Helpers;
 
 const {
-  isArrayObjectControl,
+  or,
+  isObjectArrayControl,
+  isPrimitiveArrayControl,
   rankWith
 } = Test;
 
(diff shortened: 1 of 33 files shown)


edgarmueller:
@AlexandraBuzila Did you sign the CLA?

packages/vanilla/src/complex/TableArrayControl.tsx
```diff
@@ -83,10 +87,10 @@ class TableArrayControl extends RendererComponent<TableProps, void> {
     const buttonClass = getStyleAsClassName('array.table.button');
     const controlClass = [getStyleAsClassName('array.table'),
       convertToValidClassName(controlElement.scope)].join(' ');
-    const createControlElement = (key: string): ControlElement => ({
+    const createControlElement = (key = ''): ControlElement => ({
```
eneufeld: shouldn't this be rather optional, as in the case of an object array you want to have the key and in the case of primitives you never want to have this key

packages/vanilla/src/complex/TableArrayControl.tsx
```diff
@@ -110,42 +114,71 @@ class TableArrayControl extends RendererComponent<TableProps, void> {
...
+                            schema={scopedSchema}
+                            uischema={createControlElement()}
+                            path={childPath}
+                          />
+                        </td>
+                    }
+                    <td>
+                      {
+                        errorsPerEntry ?
+                          <span style={{ color: 'red'}}>
```
eneufeld: shouldn't this be a class only, with hardcoding this to red, one can never change this otherwise?

clayroach:
This doesn't seem to be fixed... at least not for React components.  Something simple as the following doesn't find a renderer.

Schema:
```
{
  "type": "object",
  "properties": {
    "simplearrayofstrings": {
      "type": "array",
      "items": {
        "type": "string"
      }
    }
  }
}
```

JSON:
```
{
  "simplearrayofstrings": [
    "This is an example message",
    "Get ready for booohay"
  ]
}
```

edgarmueller:
@clayroach What UI schema do you use? If you have none, I could be the case that the UI schema generator hasn't been updated yet to support primitive arrays as well.

clayroach:
@edgarmueller UI Schema is:

```
{
  "type": "VerticalLayout",
  "elements": [
    {
      "type": "Control",
      "scope": "#/properties/simplearrayofstrings"
    }
  ]
}
```

edgarmueller:
It seems to work for me; using the definitions you have provided and the current jsonforms-react-seed, this is what I get: 
primitivearray

clayroach:
Bizarre.... I'm using the `jsonforms-react-seed` to test and getting...  "No applicable field found."

image

clayroach:
re-cloning, and re-installing npm libraries appears to resolve the issue.  false alarm...

PR #963, opened 2018-05-09
[Editor] Refactor schema service, support imported UI Schema model

AhmetTanakol:
- ContainmentProperty interface removed
- Unused array functions removed from schema service
- Instead of calculating container properties each time, add the calculated ones into store for reuse
- Add support for imported UI schema model


packages/editor/test/services/container.service.test.ts
```diff
@@ -0,0 +1,719 @@
...
+  test('object with object array ', () => {
+    const schema = fooBarArraySchema;
+
+    const properties = getContainerProperties(schema, schema);
+    expect(properties.length).toEqual(1);
+    expect(getPropertyLabel(properties[0])).toEqual('foo');
+    expect(properties[0].schema).toMatchObject(schema.properties.foo.items);
+  });
+
+  test('object with simple array ', () => {
```
edgarmueller: array[s]

packages/editor/test/services/container.service.test.ts
```diff
@@ -0,0 +1,719 @@
...
+      type: 'object',
+      properties: {
+        persons: {type: 'array', items: {$ref: '#/definitions/person'}},
+        robots: {type: 'array', items: {$ref: '#/definitions/robot'}}
+      }
+    };
+
+    const properties = getContainerProperties(schema, schema);
+    expect(properties.length).toEqual(2);
+    const personCopy = JSON.parse(JSON.stringify(schema.definitions.person));
```
edgarmueller: Use spread operator instead of copy via `parse/stringify`

packages/editor/test/services/container.service.test.ts
```diff
@@ -0,0 +1,719 @@
...
+        b: {type: 'object'}
+      },
+      anyOf: [
+        {$ref: '#/definitions/a'},
+        {$ref: '#/definitions/b'}
+      ]
+    };
+
+    const properties = getContainerProperties(schema, schema);
+    expect(properties.length).toEqual(0);
```
edgarmueller: Does this return 0 because the `a` and `b` definitions do not have `properties`?

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+};
+
+/**
+ * Calculate references that are used in parentSchema and add copy them into schema
+ *
+ * @param parentSchema root schema which is used to find all the schema refs
+ * @param schema current subschema without resolved references
+ * @returns JsonSchema current subschema with resolved references
+ */
+export const calculateSchemaReferences = (parentSchema: JsonSchema,
```
edgarmueller: This is the actual function that takes care of creating a self-contained schema, I'd call that way then, e.g. `makeSchemaSelfContained`

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+/**
+ * Finding required definitions from parentSchema by using schema refs
+ *
+ * @param parentSchema The root schema
+ * @param allRefs All the refs of root schema
+ * @param schemaRefs The current schema refs
+ * @param extractedReferences Contains the definition attributes for the current schema
+ * @returns SchemaRefs all the required reference paths for subschema
+ */
+const findReferencePaths = (parentSchema: JsonSchema,
```
edgarmueller: Why not call this only `findReferences`?

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+/**
+ * Interface wraps SchemaRef
+ */
+interface SchemaRefs {
+  [id: string]: SchemaRef;
+}
+
+/**
+ * Returns the label of a container
+ * @param {Property} containerProperty
```
edgarmueller: Rename param

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+ * @returns {string}
+ */
+export const getPropertyLabel =  (containerProperty: Property): string => {
```
edgarmueller: Rename to `findPropertyLabel`

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * Interface wraps SchemaRef
+ */
+interface SchemaRefs {
+  [id: string]: SchemaRef;
+}
+
+/**
+ * Returns the label of a container
```
edgarmueller: Update JsDoc and mention that this a best-effort solution to obtain a label.

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+  return schema;
+};
+
+/**
+ * Retieves a self contained schema.
+ * @param parentSchema The schema to use for resolvement
+ * @param refPath The path to resolve
+ * @return a JsonSchema that is self-contained
+ */
+export const getSelfContainedSchema = (parentSchema: JsonSchema, refPath: string): JsonSchema => {
```
edgarmueller: Rename to `makeSchemaSelfContained`

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    if (internal) {
+      // If internal is true the schema service does not need to resolve further,
+      // because only directly contained containments are wanted.
+      // This prevents running into circles
+      return [];
+    }
+
+    return Object.keys(schema.properties)
+      .reduce(
+        (prev, cur) =>
```
edgarmueller: Rename `cur` to `prop` or `currentProp`

packages/editor/src/services/container.service.ts
```diff
@@ -0,0 +1,266 @@
...
+          prev.concat(
+            getContainerProperty(
+              cur,
+              cur,
+              schema.properties[cur],
+              rootSchema,
+              false,
```
edgarmueller: Is passing `false` as the value for `isInContainment` not contradicting its own name here? Also why is `internal` set to `true`? These seem like arbitrary choices. Can you explain? Maybe update the JsDoc accordingly while you are at it

packages/editor/src/services/schema.service.impl.ts
```diff
@@ -435,14 +388,15 @@ export class SchemaServiceImpl implements SchemaService {
     this.selfContainedSchemas[editorContext.dataSchema.id] = this.editorContext.dataSchema;
   }
 
-  getContainmentProperties(schema: JsonSchema): ContainmentProperty[] {
-    return this.getContainment('root', 'root', schema,
-                               this.editorContext.dataSchema, false, null, null, null);
+  getContainerProperties(schema: JsonSchema): Property[] {
```
edgarmueller: Rename to `findContainerProperties`

packages/editor/src/services/schema.service.impl.ts
```diff
@@ -546,33 +500,27 @@ export class SchemaServiceImpl implements SchemaService {
...
-  private getContainment(key: string, name: string, schema: JsonSchema, rootSchema: JsonSchema,
-                         isInContainment: boolean,
-                         addFunction: (data: object) => (valueToAdd: object,
-                                                         neighbourValue?: object,
-                                                         insertAfter?: boolean) => void,
-                         deleteFunction: (data: object) => (valueToDelete: object) => void,
-                         getFunction: (data: object) => Object,
-                         internal = false
-  ): ContainmentProperty[] {
+  private getContainerProperty(key: string,
```
edgarmueller: Rename to `findContainerProperties` (plural)

packages/editor/src/tree/Dialog.tsx
```diff
@@ -17,61 +22,72 @@ const Dialog = (
...
-                  (d, key) => {
-                    if (prop.schema.properties[key].default) {
-                      d[key] = prop.schema.properties[key].default;
-                    }
+    rootData,
+    containersProperties,
+    rootSchema
+  }) => {
+  let containerProps;
+  if (containersProperties !== undefined && containersProperties[schema.id]) {
```
edgarmueller: Move this logic to `mapStateToProps`

packages/editor/src/tree/ObjectListItem.tsx
```diff
@@ -76,9 +82,11 @@ export interface ObjectListItemProps {
     onAdd: any;
     onSelect: any;
   };
-  schemaService: SchemaService;
   imageMapping?: any;
   labelMapping?: any;
+  containersProperties?: any;
```
edgarmueller: Add JsDoc for newly introduced properties.

packages/editor/example/uischema-example/config.ts
```diff
@@ -47,3 +47,24 @@ export const modelMapping = {
 };
 
 export const detailSchemata = {};
+
+export const importedUISchemaModel = {
```
edgarmueller: Does this belong here, where is it used?

packages/editor/src/reducers/index.ts
```diff
@@ -112,7 +146,23 @@ export const editorReducer = (
                 labelMapping: state.labelMapping,
                 modelMapping: state.modelMapping,
                 resources: state.resources,
-                identifyingProperty: state.identifyingProperty
+                identifyingProperty: state.identifyingProperty,
+                containersProperties: state.containersProperties
+            };
+        case ADD_CONTAINER_PROPERTIES:
+            if (_.isEmpty(action.containerProperties) ||
```
edgarmueller: Are there test cases that explicitly check whether the `containerProperites` have been updated?

packages/editor/src/services/container.service.ts
AhmetTanakol: I will update JsDoc. `isInContainment` is set to true if the schema has items and these items are non primitive properties, so the term `container` is used for these property objects. So this container property is directly related to type `array`. The term `property` is quite confusing but I don't know which term is correct to define this logic. The logic itself makes sense since we add children into parent so we actually need to consider parents as `containers`. `internal` value also depends on this logic. If the schema type is object, then we don't actually need to resolve it further because we already know the results from future iterations will be the children of the corresponding schema

AhmetTanakol: Maybe we should use the term `getChildren` instead of `getProperties`

packages/editor/example/uischema-example/config.ts
AhmetTanakol: To test if the uploaded ui schemata can be rendered or not but I will remove it and move to ui editor application

AhmetTanakol:
updated

packages/editor/src/reducers/index.ts
```diff
@@ -40,6 +51,24 @@ export const addUiSchema = (schemaId: string, uiSchema): AddUiSchemaAction => {
...
+/**
+ * Creates an action to add a calculated container properties for the given schema id
+ * If there already are container properties for the given schema id
+ * it is overwritten with the given one
+ *
+ * @param {JsonSchema} schema
+ * @param {[schemaId: string]: Property[]} containerProperties
+ * @returns {AddContainerPropertiesAction}
+ */
+export const addContainerProperties =
```
edgarmueller: This should rather be called `setContainerProperties` then?

packages/editor/src/helpers/util.ts
```diff
@@ -72,6 +72,7 @@ export const parentPath = (path: string): string => {
  * @param modelMapping The model mapping
  * @param uiSchemata The object containing the UI Schemata for the data types displayed by the
  *                   editor. The keys are the schema ids and the values the actual UI Schemata
+ * @param containersProperties Calculated properties of containers
```
edgarmueller: Typo in param name

packages/editor/src/helpers/containment.util.ts
```diff
@@ -1,18 +1,18 @@
...
  * @param data The data object to match
- * @param properties The array of ContainmentProperties
- * @param modelMapping The model mapping used to match the data to a containment property
- * @return The matching {@link ContainmentProperty}
+ * @param properties The array of properties of root schema
+ * @param modelMapping The model mapping used to filter properties of the given data
+ * @return The matching {@link Property}
  */
 // TODO type model mapping
 export const matchContainmentProperty =
```
edgarmueller: This is inconsistent with the naming schema of containers/containment.

packages/editor/src/reducers/index.ts
```diff
@@ -107,12 +142,13 @@ export const editorReducer = (
...
-                uiSchemata: uiSchemata,
-                imageMapping: state.imageMapping,
-                labelMapping: state.labelMapping,
-                modelMapping: state.modelMapping,
-                resources: state.resources,
-                identifyingProperty: state.identifyingProperty
+                ...state,
+                uiSchemata: uiSchemata
+            };
+        case ADD_CONTAINER_PROPERTIES:
```
edgarmueller: Rename to `SET_CONTAINER_PROPERTIES`

packages/editor/example/uischema-example/schema.ts
```diff
@@ -127,7 +127,10 @@ export const taskSchema = {
           'default': 'Categorization'
```
edgarmueller: The variable holding the schema is still called `taskSchema`.

packages/editor/src/tree/ExpandArray.tsx
```diff
@@ -36,7 +35,7 @@ export interface ExpandArrayProps {
  * a suitable delete function for the expanded elements is created.
  *
  * @param data the array to expand
- * @param property the {@link ContainmentProperty} defining the property that the array belongs to
+ * @param property the {@link Property} defining the property that the array belongs to
```
edgarmueller: JsDoc unclear: Is this just the array prop?

packages/editor/src/reducers/index.ts
```diff
@@ -1,26 +1,37 @@
 import * as _ from 'lodash';
+import { Property } from '../services/property.util';
 
 export const ADD_UI_SCHEMA: 'jsonforms/editor/ADD_UISCHEMA' = 'jsonforms/editor/ADD_UISCHEMA';
+export const ADD_CONTAINER_PROPERTIES:
```
edgarmueller: Rename to `SET_CONTAINER_PROPERTIES`

packages/editor/src/services/property.util.ts
```diff
@@ -0,0 +1,307 @@
...
+ *
+ * @param {string} property The schema key from which this property was created.
+ * @param {string} label The name of the schema the property describes.
+ * @param {JsonSchema} schema The schema is the JsonSchema this property describes.
+ * @param {JsonSchema} rootSchema The parent schema
+ * @param {boolean} isInContainer To indicate whether the schema is in a container or not
+ *                                Properties that are described in array are considered as
+ *                                they are in a container.
+ * @param {boolean} hasOnlyOwnChildren To accept only corresponding children of a property
+ * @returns {Property[]}
```
edgarmueller: Describe return type and mention that the returned properties all have a self-contained schema.

packages/editor/src/services/property.util.ts
```diff
@@ -0,0 +1,307 @@
...
+          ...containerProperties,
+          ...{[schema.id]: props}
+        };
+        return _.reduce(
+          props, (prev, curProp) => {
+            return findAllContainerProperties(curProp.schema, rootSchema, prev);
+          },
+          containerProperties
+        );
+      } else {
```
edgarmueller: Remove else statements and fall through to outer return statement

AhmetTanakol:
updated

packages/editor/src/services/property.util.ts
```diff
@@ -0,0 +1,306 @@
...
+ * Calculates label for an item which is resolved in anyOf
+ * @param {string} id The schema id
+ * @returns {string}
+ */
+const calculateLabel = (id: string): string => (
+  _.startsWith(id, '#') ? id.substring(id.lastIndexOf('#') + 1) : id
+);
+
+/**
+ * Constructing container properties of a given schema. A container means
```
edgarmueller: Rephrase (if correct): Returns a flattened list of container properties. A property is being considered a container property if it is an array of objects, meaning that all its children are non-primitive.

packages/editor/src/tree/Dialog.tsx
```diff
@@ -5,73 +5,90 @@ import * as _ from 'lodash';
...
+                  closeDialog();
+                }}
+              >
+                {prop.label}
+              </button>
+            )
+        }
+      </div>
+      <button
+        className='btn button waves-effect waves-light jsf-treeMasterDetail-dialog-close'
```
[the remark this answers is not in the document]
AhmetTanakol: Yes, at least until we migrate to MUI

packages/editor/test/services/reference.service.test.ts
```diff
@@ -436,15 +423,14 @@ describe('Schema Service Reference Tests', () => {
...
     const refProperty =
-      service.getReferenceProperties(schema.properties.classes.items as JsonSchema)[0];
+      getReferenceProperties(schema.properties.classes.items as JsonSchema)[0];
     expect(isReferenceProperty(refProperty)).toBe(true);
     expect(isContainmentProperty(refProperty)).toBe(false);
     const containmentProperty = service.getContainmentProperties(schema)[0];
     expect(isReferenceProperty(containmentProperty)).toBe(false);
     expect(isContainmentProperty(containmentProperty)).toBe(true);
-  });
+  });*/
```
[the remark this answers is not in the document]
AhmetTanakol: Actually, I forgot to remove this. Since we do not have an schema service interface there is no need to check for implementor

packages/editor/test/helpers/containment.util.test.ts
```diff
@@ -1,6 +1,6 @@
 import * as _ from 'lodash';
```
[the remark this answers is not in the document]
AhmetTanakol: yes

AhmetTanakol:
updated

PR #964, opened 2018-05-09
Remove transformProps reducer

edgarmueller:
Remove transformProps as it is a function within store which breaks redux principles.


edgarmueller:
Coverage decrease as i18n reducer is not triggered anymore. This will be work for a later sprint, as discussed.

PR #965, opened 2018-05-09
Generate UI schema if schema ownProps is available

edgarmueller:
If the schema prop is available on the `JsonForms` component generate an UI schema accordingly instead of using the one from the store.

### DIFF
--- a/packages/core/test/util/renderer.test.ts
+++ b/packages/core/test/util/renderer.test.ts
@@ -27,7 +27,8 @@ import * as _ from 'lodash';
 import {
   createDefaultValue,
   mapDispatchToControlProps,
-  mapStateToControlProps, mapStateToDispatchRendererProps
+  mapStateToControlProps,
+  mapStateToDispatchRendererProps
 } from '../../src/util';
 import configureStore from 'redux-mock-store';
 import { UPDATE_DATA, UpdateAction } from '../../src/actions';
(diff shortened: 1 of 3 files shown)


packages/core/test/util/renderer.test.ts
```diff
@@ -27,10 +27,12 @@ import * as _ from 'lodash';
 import {
   createDefaultValue,
   mapDispatchToControlProps,
-  mapStateToControlProps
+  mapStateToControlProps, mapStateToDispatchRendererProps
```
eneufeld: weird formatting

packages/core/test/util/renderer.test.ts
```diff
@@ -296,3 +295,30 @@ test('createDefaultValue', t => {
...
+  };
+
+  const props = mapStateToDispatchRendererProps(
+    store.getState(),
+    { schema }
+  );
+  t.deepEqual(props.uischema, generateDefaultUISchema(schema));
+});
+
+test(`mapStateToDispatchRendererProps should use registered UI schema given no ownProps`, t => {
```
eneufeld: shouldn't we add the test case where a uischema is passed in using the ownProps?

edgarmueller:
@eneufeld Updated

eneufeld:
Thank you, looks good.

PR #970, opened 2018-05-16
Avoid rendering unknown fields and empty layouts

AhmetTanakol:
Return null if an unknown field is rendered
Set the layout as null if the type of field is object with empty properties when generating uischema

### DIFF
--- a/packages/editor/src/helpers/NonEmptyLayout.tsx
+++ b/packages/editor/src/helpers/NonEmptyLayout.tsx
@@ -1,0 +1,49 @@
+import {
+  Layout,
+  mapStateToLayoutProps,
+  RankedTester,
+  rankWith,
+  RendererProps,
+  Tester,
+} from '@jsonforms/core';
+import { connectToJsonForms } from '@jsonforms/react';
+import * as _ from 'lodash';
+
+/**
+ * Checks whether the given UI schema contains only label field or not.
+ */
+export const isInstanceOfLayout: Tester =
+  (uischema: Layout): boolean => {
+      if (_.isEmpty(uischema)) {
+        return false;
+      }
+      if (!uischema.hasOwnProperty('elements')) {
+        return false;
+      }
+      return _.isEmpty(uischema.elements) ||
+        (
+             uischema.elements.length === 1 &&
+             _.some(uischema.elements, element => element.type === 'Label')
+        );
+    };
+/**
+ * Tester for empty layouts
+ * @type {RankedTester}
+ */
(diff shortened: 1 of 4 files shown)


packages/editor/src/helpers/NonEmptyLayout.tsx
```diff
@@ -0,0 +1,49 @@
...
+  RendererProps,
+  Tester,
+} from '@jsonforms/core';
+import { connectToJsonForms } from '@jsonforms/react';
+import * as _ from 'lodash';
+
+/**
+ * Checks whether the given UI schema contains only label field or not.
+ */
+export const isInstanceOfLayout: Tester =
```
edgarmueller: Rename tester to `isEmptyLayout`

packages/editor/src/helpers/NonEmptyLayout.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const nonEmptyLayoutTester: RankedTester =
+  rankWith(3, isInstanceOfLayout);
+
+/**
+ * NonEmptyLayoutRenderer is a custom layout renderer to prevent rendering empty layouts
+ *
+ * @returns null
+ * @constructor
+ */
+export const NonEmptyLayoutRenderer = (
```
edgarmueller: Add test case for `NonEmptyLayoutRenderer`.

packages/material/src/controls/MaterialInputControl.tsx
```diff
@@ -61,23 +63,27 @@ export class MaterialInputControl extends Control<ControlProps, ControlState> {
...
-          {computeLabel(isPlainLabel(label) ? label : label.default, required)}
-        </InputLabel>
-        <DispatchField uischema={uischema} schema={schema} path={parentPath} />
-        <FormHelperText error={!isValid}>
-          {!isValid ? formatErrorMessage(errors) : showDescription ? description : null}
-        </FormHelperText>
-      </FormControl>
-    );
+    const field = _.maxBy(fields, r => r.tester(uischema, schema));
+    if (field === undefined || field.tester(uischema, schema) === -1) {
```
edgarmueller: A couple of things: 
1. Use `NOT_APPLICABLE` constant.
2. Are there any other controls that need to be updated to make the behavior consistent, e.g. `MaterialBooleanControl`? What about the vanilla renderer set? Ideally, this check could be moved into a [HOC](https://reactjs.org/docs/higher-order-components.html)
3. Add test case(s)

AhmetTanakol:
Moving this to jsonforms-ui-editor

Issue #954, opened 2018-05-02
[Material] Horizonally aligned control next to a enum control is not properly aligned


lucas-koehler:
The non-enum control's line is a few pixels lower than the enum control's one. See Clock Speed in screenshot. This was recorded in Theia.

![image](https://user-images.githubusercontent.com/6959840/39530271-ea4a42d2-4e28-11e8-97d3-117ba7e72fca.png)

AhmetTanakol:
I also have a similar issue while working on generic editor. No margin between text field and date field.

![screen shot 2018-05-07 at 5 43 03 pm](https://user-images.githubusercontent.com/10375761/39710863-50d953dc-521e-11e8-8c06-4ee66f5d474a.png)

eneufeld:
I think this is fixed with Material-UI 1.0.0
see:
![image](https://user-images.githubusercontent.com/1176277/40448803-e208411e-5ed6-11e8-9b91-f9cd5f3e952b.png)

edgarmueller:
@eneufeld Right, closing it

PR #982, opened 2018-05-23
Added missing space in material horizontal layout renderer

eneufeld:
This error was also mentioned in #954


edgarmueller:
Thank you!

PR #984, opened 2018-05-24
Add custom input label style to not wrap the label

### LINKED ISSUE #936: Text label overlaps with the input field when the input field is trimmed
screen shot 2018-04-12 at 11 31 42 am

eneufeld:
Fixes #936


edgarmueller:
A bit awkward, but I guess there's no other way, thanks!

PR #996, opened 2018-06-06
Etract naming and image mapping

AhmetTanakol:


packages/editor/example/index.ts
```diff
@@ -28,12 +29,52 @@ window.onload = () => {
...
+  const imageGetterPredicate = (schemaId: string) =>
+    !_.isEmpty(imageProvider) ? `icon ${imageProvider[schemaId]}` : '';
+
   const store = createEditorStore({}, taskSchema, uischema, materialFields,
                                   materialRenderers, imageProvider, labelProvider, modelMapping,
                                   detailSchemata, {});
 
   ide.filterPredicate = filterPredicate;
 
+  ide.namingPredicate = getNamingFunction;
```
edgarmueller: These aren't predicates since they don't return a boolean, I'd rather stick with naming schema of `labelProvider` and `imageProvider` throughout the code.

packages/editor/example/index.ts
```diff
@@ -28,12 +29,52 @@ window.onload = () => {
     };
   };
 
+  const getNamingFunction =
```
edgarmueller: Rename to `labelProvider`

packages/editor/src/ide.ts
```diff
@@ -42,10 +44,31 @@ export class JsonEditorIde extends HTMLElement implements Editor {
     return getSchema(this._store.getState());
   }
 
+  /**
+   * Filtering selected list item from available items
```
edgarmueller: While we're at it, let's update the JSDoc here as well: Rephrase to "Set the predicate function that is used by the array renderer to filter container properties'

AhmetTanakol:
updated

PR #999, opened 2018-06-06
Fix react/MUI related warnings

edgarmueller:
Fix warnings that appeared in the console:

* Wrap disabled button in div within `TableToolbar`
* Fix stylenames in input control

### DIFF
--- a/packages/material/src/complex/TableToolbar.tsx
+++ b/packages/material/src/complex/TableToolbar.tsx
@@ -79,14 +79,16 @@ export const TableToolbar = (
             </Grid>
             <Grid item>
               <Tooltip title='Delete'>
-                <Button
-                  variant='fab'
-                  aria-label={`Delete`}
-                  disabled={numSelected === 0}
-                  onClick={openConfirmDeleteDialog}
-                >
-                  <DeleteIcon />
-                </Button>
+                <div>
+                  <Button
+                    variant='fab'
+                    aria-label={`Delete`}
+                    disabled={numSelected === 0}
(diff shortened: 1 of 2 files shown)


packages/material/src/complex/TableToolbar.tsx
```diff
@@ -79,14 +79,16 @@ export const TableToolbar = (
             </Grid>
             <Grid item>
               <Tooltip title='Delete'>
-                <Button
```
eneufeld: why do we need to wrap this in a div?

packages/material/src/controls/MaterialInputControl.tsx
```diff
@@ -61,11 +61,11 @@ export class MaterialInputControl extends Control<ControlProps, ControlState> {
       style.display = 'none';
     }
     const inputLabelStyle: {[x: string]: any} = {
-      'white-space': 'nowrap',
-      'overflow' : 'hidden',
-      'text-overflow': 'ellipsis',
+      whiteSpace: 'nowrap',
```
eneufeld: is this auto resolved to `white-space`?

edgarmueller: Yes, hyphens are replaced with casing.

packages/material/src/complex/TableToolbar.tsx
edgarmueller: Because the tooltip doesn't pop up otherwise, see https://stackoverflow.com/a/18113961

PR #1004, opened 2018-06-13
Allow overriding of handleChange and data field props

AhmetTanakol:


packages/core/src/util/field.ts
```diff
@@ -138,3 +139,18 @@ export const defaultMapStateToEnumFieldProps = (state, ownProps): StatePropsOfEn
...
 export const mapDispatchToFieldProps: (dispatch) => DispatchPropsOfControl =
   mapDispatchToControlProps;
+
+/**
+ * Default map dispatch to enum field props to customize handleChange action
+ *
+ * @param dispatch
+ * @param ownProps
+ */
+export const defaultMapDispatchToEnumFieldProps = (dispatch, ownProps) => {
```
edgarmueller: Isn't this actually more generic in the sense that this is `defaultMapDispatchToControlProps`?

packages/core/src/util/field.ts
```diff
@@ -138,3 +139,18 @@ export const defaultMapStateToEnumFieldProps = (state, ownProps): StatePropsOfEn
  */
 export const mapDispatchToFieldProps: (dispatch) => DispatchPropsOfControl =
   mapDispatchToControlProps;
+
+/**
+ * Default map dispatch to enum field props to customize handleChange action
+ *
+ * @param dispatch
```
edgarmueller: Remove useless param descriptions

packages/core/src/util/field.ts
AhmetTanakol: Yes, you are right. User can overwrite `handleChange` method for all fields. I will change it

AhmetTanakol:
Moved default data value control to `mapStateToFieldProps`

PR #1005, opened 2018-06-14
Added docu for styling the vanilla renderers

eneufeld:
Also updated the example styles to include function case

### DIFF
--- a/packages/vanilla/src/util/index.ts
+++ b/packages/vanilla/src/util/index.ts
@@ -36,7 +36,7 @@ import { VanillaControlStateProps, VanillaLayoutProps } from '../index';
  */
 export interface StyleDef {
   name: string;
-  classNames: string[] | ((args: any[]) => string[]);
+  classNames: string[] | ((...args: any[]) => string[]);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/vanilla/src/util/index.ts
```diff
@@ -153,6 +153,10 @@ export const vanillaStyles = [
     name: 'horizontal.layout',
     classNames: ['horizontal-layout']
   },
+  {
+    name: 'horizontal.layout.item',
+    classNames: numberOfChildren => ['horizontal-layout-' + numberOfChildren[0]]
```
edgarmueller: I think our typings are broken here right now, because `numberOfChildren` is a nested array ATM. We can fix this the following way (which also reflects the intended usage).
`StyleDef` should be defined this way (note varargs siganture on `args`):
```ts
export interface StyleDef {
  name: string;
  classNames: string[] | ((...args: any[]) => string[]);
}
```
If `classNames` is a function, we need to spread the `args` as well (see `styles.ts`):
```ts
    return foundStyle.classNames(...args);
```
This way, a `classNames` function can be used the following way (can we also use this as the example?):
```ts
{
    name: 'horizontal.layout.item',
    classNames: ([size, somethingElse]) => {
      const cols = 12 / size;
      console.log(somethingElse);
      return [`col-md-${cols}`];
    }
  }
```

PR #1013, opened 2018-06-21
Define a uischema registry in the core (#994)

eneufeld:
* Add UI schema registry reducer
This allows to register, unregister and retrieve uischemas.
The user can now configure how the detail in an ArrayLayout renderer should look like.


edgarmueller:
Thanks!

PR #1018, opened 2018-07-02
 Rename editor package to material-tree-renderer 

edgarmueller:
Some parts related to the reference service haven't been renamed, but marked as deprecated. These should be moved/removed when we will be moving the renderer to the respective renderer-set/core.


edgarmueller:
@eneufeld Yes, correct.

PR #1020, opened 2018-07-02
improve TableToolBar styling

AhmetTanakol:


packages/material/src/complex/MaterialArrayControlRenderer.tsx
```diff
@@ -164,4 +172,4 @@ export interface TableState {
 export default connectToJsonForms(
   mapStateToTableControlProps,
   mapDispatchToTableControlProps
-)(MaterialArrayControlRenderer);
+)(withStyles(styles, { name: 'MaterialArrayControlRenderer' })(MaterialArrayControlRenderer));
```
edgarmueller: Could we use `withStyles` with the `ValidationIcon` component directly instead of passing down the `classes` prop all the way down.

AhmetTanakol:
Updated

edgarmueller:
Thanks!

PR #1024, opened 2018-07-12
Add schema based condition

edgarmueller:
Adds a new 'SchemaBasedCondition' condition type

### DIFF
--- a/packages/core/src/util/runtime.ts
+++ b/packages/core/src/util/runtime.ts
@@ -25,48 +25,70 @@
 import * as _ from 'lodash';
 // TODO: pass in uischema and data instead of props and state
 import { getData } from '../reducers';
-import { LeafCondition, RuleEffect, UISchemaElement } from '../models/uischema';
+import {
+  Condition,
+  LeafCondition,
+  RuleEffect,
+  SchemaBasedCondition,
+  UISchemaElement
+} from '../models/uischema';
 import { resolveData } from './resolvers';
 import { toDataPath } from './path';
+import { createAjv } from './validator';
 
-const isRuleDefined = (uischema: UISchemaElement): boolean =>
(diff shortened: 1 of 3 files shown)


packages/core/src/util/runtime.ts
```diff
@@ -25,48 +25,70 @@
...
+  (!_.has(uischema, 'rule.condition.expectedValue') && !_.has(uischema, 'rule.condition.schema'));
 
-export const evalVisibility = (uischema: UISchemaElement, data: any) => {
-  // TODO condition evaluation should be done somewhere else
-  if (isRuleDefined(uischema)) {
+const isLeafCondition = (condition: Condition): condition is LeafCondition =>
+  condition.type === 'LEAF';
 
+const isSchemaCondition = (condition: Condition): condition is SchemaBasedCondition =>
+  condition.type === undefined;
```
eneufeld: I would check for the schema attribute

packages/core/src/util/runtime.ts
```diff
@@ -25,48 +25,70 @@
...
 };
 
-export const evalEnablement = (uischema: UISchemaElement, data: any) => {
-
-  if (isRuleDefined(uischema)) {
+export const evalVisibility = (uischema: UISchemaElement, data: any) => {
+  const fulfilled = isConditionFulfilled(uischema, data);
 
-    return true;
+  switch (uischema.rule.effect) {
```
eneufeld: can we unify the formatting for evalVisibility and evalEnablement

packages/core/src/util/runtime.ts
edgarmueller: Makes sense, fixed

packages/core/src/util/runtime.ts
edgarmueller: Fixed

PR #1025, opened 2018-07-12
Port slider renderer to material-ui

### LINKED ISSUE #583: Services should use common naming scheme for listeners registration
`DataService`, `Runtime` as well as any additional future service implementation should use a common naming scheme for listener registration and de-registration. This issue is about either fixing that naming scheme or somehow enforce it, if applicable.

mathansen:
* Introduces dependency to mui/lab as the slider is not included in mui/core
* Refactor existing slider field to control
* Adapt tests
* Fixes #583


mathansen:
Test case for multipleOf missing. Will submit it tomorrow.

mathansen:
Added the check for multipleOf (verifies that the step size is set correctly).

Issue #989, opened 2018-05-30
Multiple checkboxes


danielclariondoor:
Nice package. Thanks.
Couple of simple things I can't figure out from the examples.

How can I create a checkbox input group?

edgarmueller:
@danielclariondoor Glad you like it! Currently, we do not support creating input groups (but you can always provide your own custom controls). Could you elaborate about your use case a bit with regards to schema and the expected controls? Maybe we can come up with something. Also, which renderer set are you using?

sdirix:
This issue will be closed as it has been open for a long time with no activity. Feel free to reopen this issue if needed.

PR #1029, opened 2018-07-13
Add radio group controls for enums (#989)

AlexandraBuzila:
Adds MaterialRadioGroupControl  and vanilla RadioGroupControl that renders enums as radio groups.

selection_242

### DIFF
--- a/packages/material/src/controls/MaterialRadioGroupControl.tsx
+++ b/packages/material/src/controls/MaterialRadioGroupControl.tsx
@@ -1,0 +1,112 @@
+/*
+  The MIT License
+
+  Copyright (c) 2018 EclipseSource Munich
+  https://github.com/eclipsesource/jsonforms
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
+  furnished to do so, subject to the following conditions:
+
+  The above copyright notice and this permission notice shall be included in
+  all copies or substantial portions of the Software.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+import * as React from 'react';
+import {
+    computeLabel,
+    ControlElement,
+    ControlProps,
+    ControlState,
+    formatErrorMessage,
+    isDescriptionHidden,
+    isPlainLabel,
+    mapDispatchToControlProps,
+    mapStateToControlProps,
+    resolveSchema,
+} from '@jsonforms/core';
+import { connectToJsonForms, Control } from '@jsonforms/react';
+import Radio from '@material-ui/core/Radio';
+import RadioGroup from '@material-ui/core/RadioGroup';
+import {
+    FormControl,
+    FormControlLabel,
+    FormHelperText,
+    FormLabel
+} from '@material-ui/core';
+
+export class MaterialRadioGroupControl extends Control<ControlProps, ControlState> {
+    render() {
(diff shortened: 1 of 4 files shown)


packages/material/src/controls/MaterialRadioGroupControl.tsx
```diff
@@ -0,0 +1,130 @@
...
+            uischema,
+            schema,
+            visible
+        } = this.props;
+        const isValid = errors.length === 0;
+        const style: { [x: string]: any } = {};
+        if (!visible) {
+            style.display = 'none';
+        }
+        const inputLabelStyle: { [x: string]: any } = {
```
edgarmueller: Do we need this label style?

packages/examples/src/rule.ts
```diff
@@ -41,7 +41,7 @@ export const schema = {
 };
 
 export const uischema = {
-  type: 'HorizontalLayout',
+  type: 'VerticalLayout',
```
edgarmueller: I'd rather a new, explicit example for demonstrating the radio button group and revert this change.

packages/material/src/controls/MaterialRadioGroupControl.tsx
```diff
@@ -0,0 +1,130 @@
...
+import Radio from '@material-ui/core/Radio';
+import RadioGroup from '@material-ui/core/RadioGroup';
+import {
+    FormControl,
+    FormControlLabel,
+    FormHelperText,
+    FormLabel
+} from '@material-ui/core';
+
+export class MaterialRadioGroupControl extends Control<ControlProps, ControlState> {
```
edgarmueller: Do we have a vanilla version of this?

packages/material/test/renderers/MaterialRadioGroupControl.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+    const inputs: HTMLInputElement[] = TestUtils.scryRenderedDOMComponentsWithTag(tree, 'input');
+    const radioButtons = _.filter(
+      inputs,
+      i => i.type === 'radio'
+    );
+    expect(radioButtons.length).toBe(4);
+    let currentlyChecked = _.filter(radioButtons, radio => radio.checked);
+    expect(currentlyChecked.length).toBe(1);
+    expect(currentlyChecked[0].value).toBe('D');
```
edgarmueller: We tested this expectation in the test before this one. Perhaps remove the assertions here and rename the tests accordingly. This test should only whether the binding works as intended.

packages/material/src/controls/MaterialRadioGroupControl.tsx
AlexandraBuzila: Pushed new commit containing vanilla renderer as well.

packages/vanilla/src/controls/RadioGroupControl.tsx
```diff
@@ -0,0 +1,112 @@
...
+                className={classNames.wrapper}
+                hidden={!visible}
+                onFocus={this.onFocus}
+                onBlur={this.onBlur}
+            >
+                <label htmlFor={id} className={classNames.label} >
+                    {computeLabel(isPlainLabel(label) ? label : label.default, required)}
+                </label>
+
+                <form>
```
eneufeld: why wrap in a form?

packages/vanilla/src/controls/RadioGroupControl.tsx
```diff
@@ -0,0 +1,112 @@
...
+                    {
+                        options.map(optionValue =>
+                            (
+                                <div>
```
eneufeld: why wrap each input into an own div?

packages/vanilla/src/controls/RadioGroupControl.tsx
```diff
@@ -0,0 +1,112 @@
...
+    isPlainLabel,
+    mapDispatchToControlProps,
+    mapStateToControlProps,
+    resolveSchema
+} from '@jsonforms/core';
+import { connectToJsonForms, Control } from '@jsonforms/react';
+import { VanillaControlProps } from '../index';
+import { addVanillaControlProps } from '../util';
+
+export class RadioGroupControl extends Control<VanillaControlProps, ControlState> {
```
eneufeld: In my opinion this can be a field, at least in the way it is currently implemented

packages/vanilla/src/controls/RadioGroupControl.tsx
AlexandraBuzila: I replaced it with a div, needed for styling.

packages/vanilla/src/controls/RadioGroupControl.tsx
AlexandraBuzila: Because we need to return an input and a label.

packages/vanilla/src/controls/RadioGroupControl.tsx
AlexandraBuzila: Discussed it with Edgar, we left it as a control for now.

AlexandraBuzila:
> the FormLabel is much bigger then all other labels in the form.

It is the same in the material examples: the FormLabel is big and does not change size when a value is selected in the radio group.
https://material-ui.com/demos/selection-controls/#radio-buttons

> Shouldn't we display Material and Vanilla RadioGroup in the same way?
> Material is showing the options as row, and vanilla as column.

I fixed it in the new commit.

PR #1030, opened 2018-07-13
Make ids unique

edgarmueller:


packages/core/src/util/ids.ts
```diff
@@ -0,0 +1,29 @@
...
+  let tries = 0;
+  while (!isUniqueId(proposedID, tries)) {
+    tries++;
+  }
+  const newID = makeId(proposedID, tries);
+  idMappings.set(element, newID);
+  return newID;
+};
+
+export const removeId = uischema => idMappings.delete(uischema);
```
edgarmueller: Add type annotation to `uischema`. Can we add a test case for this, i.e. upon unmounting the id should be deleted?

packages/core/src/util/ids.ts
```diff
@@ -0,0 +1,29 @@
+import { UISchemaElement } from '..';
+
+const idMappings: Map<UISchemaElement, string> = new Map<UISchemaElement, string>();
+
+export const createId = (element: UISchemaElement, proposedID: string) => {
```
edgarmueller: Rename `element` to `uischema` for consistency

packages/react/src/DispatchField.tsx
```diff
@@ -26,32 +26,55 @@ import * as React from 'react';
...
+  render() {
+    const { uischema, schema, path, fields } = this.props as DispatchFieldProps;
+    const field = _.maxBy(fields, r => r.tester(uischema, schema));
+
+    if (field === undefined || field.tester(uischema, schema) === -1) {
+      return <UnknownRenderer type={'field'}/>;
+    } else {
+      const Field = field.field;
+
+      if (uischema.type !== undefined && uischema.type === 'Control') {
```
edgarmueller: Do we need to check `type` for being undefined?

packages/react/src/DispatchField.tsx
```diff
@@ -26,32 +26,55 @@ import * as React from 'react';
...
-    const Field = field.field;
-
-    return (
-      <Field
-        schema={schema}
-        uischema={uischema}
-        path={dispatchFieldProps.path}
-      />
-    );
+class Dispatch extends React.Component<DispatchFieldProps, { id: string }> {
```
edgarmueller: Extract state type to explicit State Interface (also see controls).

edgarmueller:
@eneufeld Please also have a look.

edgarmueller:
Follow up: #1048

Issue #1037, opened 2018-07-18
Clean-up example package


edgarmueller:
* Package name should be updated
* Fix hoisting package warnings

eneufeld:
refactor examples, so that they can be used by the native react example and a future native angular example packages

PR #1038, opened 2018-07-18
Update package versions and locks

sdirix:


edgarmueller:
Thanks, this also partly fixes #1037.

PR #1041, opened 2018-07-19
Fixed the no data state of the array renderer

### LINKED ISSUE #1040: Empty array renderer looks weird
When rendering an empty array, it looks like follows:

selection_062

Two minor issues are:
1. The checkbox should not be checked
2. The 'No data' label is too small and does not span the entire width of the control

eneufeld:
Fixes #1040

### DIFF
--- a/packages/material/src/complex/MaterialTableControl.tsx
+++ b/packages/material/src/complex/MaterialTableControl.tsx
@@ -73,9 +73,14 @@ const generateHeaderCellForPrimitives = toolbarProps => {
     </TableCell>
   );
 };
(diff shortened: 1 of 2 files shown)


packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -73,17 +73,22 @@ const generateHeaderCellForPrimitives = toolbarProps => {
     </TableCell>
   );
 };
+const getValidColumnProps = (scopedSchema: JsonSchema) =>
```
edgarmueller: This seems to crash when first clicking 'select all', followed by a delete. We should also add a test case covering this scenario.

PR #1047, opened 2018-07-22
Clean up Example Packages

### LINKED ISSUE #1037: Clean-up example package

eneufeld:
The examples must still be started from a concrete renderer set.
They can now be debugged with React tools as they run as react apps.

Changes in detail:
* Changed 'examples' package to only provide data
  * it has no dependencies to react or any renderer set
* Changed 'example' package to be the react examples util
  * it provides a method to render the examples
  * it extends the provided examples to add actions
    * allows to register/unregister rating control
    * allows to register/unregister custom ui schema
    * allows to show a full data switch
  * the renderers and fields to be used must be provided
  * it has no dependencies to any renderer set
* Changed 'material' and 'vanilla' packages to use the 'example' package

Fixes #1037


packages/material/example/index.ts
```diff
@@ -22,11 +22,11 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 */
-import { createExampleSelection } from '../../examples/src/register';
+import { renderExample } from '../../example/src/index';
 import { materialFields, materialRenderers } from '../src';
 
 window.onload = () => {
```
edgarmueller: I don't think we need to registerd this via `onload` anymore, could be removed.

packages/examples/src/config.ts
```diff
@@ -74,16 +65,14 @@ registerExamples([
     label: 'Configuration (Default)',
     data,
     schema,
-    uiSchema: uischema,
-    translations
+    uiSchema: uischema
```
edgarmueller: While we are it, could we unify casing here, so using `uischema` instead of `uiSchema`?

packages/examples/src/nestedArrays.ts
```diff
@@ -0,0 +1,114 @@
+/*
```
edgarmueller: This example currently doesn't work with the vanilla renderers, since it's not implemented properly. Would it make sense to exlucde it somehow from the vanila examples?

packages/example/src/reduxUtil.ts
```diff
@@ -0,0 +1,79 @@
...
+      dispatch(Actions.init(example.data, example.schema, example.uiSchema));
+      if (example.config) {
+        Actions.setConfig(example.config(dispatch));
+      }
+    }
+  });
+const mergeProps = (stateProps, dispatchProps, ownProps) => {
+    return Object.assign({}, ownProps, {
+      ...stateProps,
+      dispatch: dispatchProps.dispatch,
```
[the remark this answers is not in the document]
eneufeld: how, I didn't come up with a good idea?

packages/example/src/reduxUtil.ts
```diff
@@ -0,0 +1,79 @@
...
+      selectedExample
+    };
+  };
+const mapDispatchToProps = dispatch => ({
+    dispatch: dispatch,
+    changeExampleData: example => {
+      dispatch({ type: 'changeExample', selectedExample: example });
+      dispatch(Actions.init(example.data, example.schema, example.uiSchema));
+      if (example.config) {
+        Actions.setConfig(example.config(dispatch));
```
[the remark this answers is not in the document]
eneufeld: you are right, messed up the parenthesis. 
I think, that we should refactor setConfig to not return a function but a simple action in the first place.

packages/examples/src/config.ts
eneufeld: sure

PR #1048, opened 2018-07-23
Unique ID support in Core, React and Material

### LINKED ISSUE #691: Control IDs should be unique and escaped
IDs are not unique since the schema path is used for them currently. Also, if we use a variation of the schema path, we should make sure to escape it.

sdirix:
Fixes #691 

Content:
 * Adds unique ID support in ```core```
 * Uses unique ID support in ```react``` to provide IDs
 * Uses generated IDs in ```react``` to derive unique IDs for HTML elements in Material renderers

Open Tasks:
 * Implement tests for ALL material fields
 * Add unique ID support for arrays

### DIFF
--- a/packages/core/src/util/ids.ts
+++ b/packages/core/src/util/ids.ts
@@ -1,0 +1,26 @@
+const usedIDs: Set<string> = new Set<string>();
+
+export const createId = (proposedID: string) => {
+  if (proposedID === undefined) {
+    // failsafe to avoid endless loops in error cases
+    proposedID = 'undefined';
+  }
+  let tries = 1;
+  while (!isUniqueId(proposedID, tries)) {
+    tries++;
+  }
+  const newID = makeId(proposedID, tries);
+  usedIDs.add(newID);
+  return newID;
+};
+
+export const removeId = id => usedIDs.delete(id);
+
+const isUniqueId = (idBase: string, iteration: number) => {
+  const newID = makeId(idBase, iteration);
+  return !usedIDs.has(newID);
+};
+
+const makeId = (idBase: string, iteration: number) => iteration <= 1 ? idBase : idBase + iteration;
+
+export const clearAllIds = () => usedIDs.clear();
--- a/packages/react/src/DispatchField.tsx
+++ b/packages/react/src/DispatchField.tsx
@@ -27,31 +27,30 @@ import { connect } from 'react-redux';
 import * as _ from 'lodash';
 import { UnknownRenderer } from './UnknownRenderer';
 import { DispatchFieldProps, mapStateToDispatchFieldProps } from '@jsonforms/core';
+
 /**
- * Props of the {@link DispatchField} renderer.
+ * Dispatch renderer component for fields.
  */
-  /**
-   * The available field renderers.
-   */
+class Dispatch extends React.Component<DispatchFieldProps> {
+  render() {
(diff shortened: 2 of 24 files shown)


packages/core/src/util/ids.ts
```diff
@@ -0,0 +1,26 @@
...
+  let tries = 1;
+  while (!isUniqueId(proposedID, tries)) {
+    tries++;
+  }
+  const newID = makeId(proposedID, tries);
+  usedIDs.add(newID);
+  return newID;
+};
+
+export const removeId = id => usedIDs.delete(id);
```
edgarmueller: Add type annotation to `id`

packages/react/src/JsonForms.tsx
```diff
@@ -26,26 +26,57 @@ import * as React from 'react';
...
 import { UnknownRenderer } from './UnknownRenderer';
-import { JsonFormsProps, mapStateToDispatchRendererProps } from '@jsonforms/core';
+import {
+  createId,
+  JsonFormsProps,
+  mapStateToDispatchRendererProps,
+  removeId
+} from '@jsonforms/core';
+
+class JsonFormsDispatchRenderer extends React.Component<JsonFormsProps, { id: string }> {
```
edgarmueller: Extract state to explicit type

packages/react/src/JsonForms.tsx
```diff
@@ -26,26 +26,57 @@ import * as React from 'react';
...
+  }
 
-const JsonFormsDispatchRenderer =
-  ({ uischema, schema, path, renderers }: JsonFormsProps) => {
-  const renderer = _.maxBy(renderers, r => r.tester(uischema, schema));
-  if (renderer === undefined || renderer.tester(uischema, schema) === -1) {
-    return <UnknownRenderer type={'renderer'}/>;
-  } else {
-    const Render = renderer.renderer;
+  isManageOwnID(props) {
```
edgarmueller: Re-use `isControl`

packages/react/test/renderers/JsonForms.test.tsx
```diff
@@ -186,3 +210,61 @@ test('deregister an unregistered renderer should be a no-op', t => {
...
+        {children}
+      </div>
+    );
+  };
+
+  const store = initJsonFormsStore(
+    {
+      data: t.context.data,
+      schema: t.context.schema,
+      uischema: t.context.uischema2,
```
edgarmueller: Make `uischema2` a local var as it doesn't seem to be used anywhere else.

edgarmueller:
For the remaining open tasks, let's create corresponding issues.

packages/react/src/DispatchField.tsx
```diff
@@ -27,31 +27,30 @@ import { connect } from 'react-redux';
...
-   * The available field renderers.
-   */
-
-const Dispatch = (dispatchFieldProps: DispatchFieldProps) => {
-  const uischema = dispatchFieldProps.uischema;
-  const schema = dispatchFieldProps.schema;
-  const field = _.maxBy(dispatchFieldProps.fields, r => r.tester(uischema, schema));
+class Dispatch extends React.Component<DispatchFieldProps> {
+  render() {
+    const { uischema, schema, path, fields, id } = this.props as DispatchFieldProps;
```
eneufeld: I think the cast is not needed

edgarmueller:
Closing in favor of #1050

PR #1052, opened 2018-08-08
Remove $id dependency from property util

edgarmueller:
This is a 1st attempt at removing the assumption that relevant subschemas have to be annotated with `$id` when using the `TreeWithDetail` renderer. These changes might cause the renderer to be somewhat slower as we can not lookup the container properties anymore, but instead re-calculate them. It also depends on a deep equal comparison of the involved schemas. Tthe `modelMapping`, label and image provider are pure functions now. Declarative maps can still be provided by creating helpers around these functions, e.g. we could do so in the theia-tree-editor extension. I believe these changes have simplified the code a bit, and we can still re-introduce lookups later as an optimization step, in case the schema provides `$id`s.

Relevant changes also include:

* Move `traverse` to common util
* Remove reference service (if this is in use anywhere it should be moved there, this is currently dead code)
* Export `UISchemaTester` type
* Remove `treeWithDetail`/`editor` reducer
* Use UI schemas state to register detail UI schemas
* Rename types in `TreeWithDetailRenderer` for consistency reasons


packages/material-tree-renderer/src/services/property.util.ts
```diff
@@ -56,7 +53,7 @@ export const findPropertyLabel = (property: Property): string => {
     [
       property.schema.title,
       property.label,
-      property.schema.$id,
+      (property.schema as any).$id || (property.schema as any).id,
```
edgarmueller: Fix casts

packages/material-tree-renderer/src/services/property.util.ts
```diff
@@ -1,17 +1,14 @@
 import * as _ from 'lodash';
-import { JsonSchema7 } from '@jsonforms/core';
+import { JsonSchema } from '@jsonforms/core';
```
AhmetTanakol: Are we using version 4? If yes, this might cause inconsistencies for the applications that depend on material-tree-renderer. As I remembered using version 7 in ui-schema-editor with material-tree-renderer which depends on version 4 was problematic.

packages/material-tree-renderer/src/services/property.util.ts
```diff
@@ -171,84 +156,93 @@ const calculateLabel = (id: string): string => (
...
-                                        schema: JsonSchema7,
-                                        rootSchema: JsonSchema7,
-                                        isInContainer: boolean,
-                                        hasOnlyOwnChildren = false): Property[] => {
+const findContainerProps = (property: string,
+                            label: string,
+                            schema: JsonSchema,
+                            rootSchema: JsonSchema,
+                            isInContainer: boolean,
+                            visisted: Map<string, string>,
```
AhmetTanakol: there is a typo

packages/material-tree-renderer/src/services/property.util.ts
AhmetTanakol: Since using `id` is not mandatory, should we really check this?

packages/material-tree-renderer/example/index.ts
```diff
@@ -1,131 +1,120 @@
...
+  },
+  {
+    tester: schema => _.keys(schema.properties).some(key => key === 'users') ?
+      1 : NOT_APPLICABLE,
+    uischema: userGroupView
+  }
+];
+
+const filterPredicate = () => (property: Property): boolean =>
+  traverse(taskSchema, subSchema => subSchema === property.schema) !== undefined;
```
AhmetTanakol: Maybe instead of `===`, we can use `_.isEqual()` for deep comparison

edgarmueller: This always returned false, I'm wondering why this actually worked.

edgarmueller: Ah right, got it, `traverse` always returns a `boolean` (which is kinda weird, but that's a different story), hence the expression evaluated to true. I hopefully fixed this by tracking the schema path in `findContainerProps` as well and maintaining an additional property `orgSchema` in `Property` which is the reference to the sub schema of the root schema, so that we are able to compare by reference here.

eneufeld:
I like the change, but I would still suggest to split it into at least two part, one being the refactoring of the label and image provider and a second being the change of the resolver.

edgarmueller:
@eneufeld I've rebased my changes, please have a look again

packages/material-tree-renderer/example/index.ts
```diff
@@ -1,131 +1,117 @@
...
-        if (!_.isEmpty(info.property) && !_.isEmpty(element[info.property])) {
-          label = _.isEmpty(label) ?
-            element[info.property] :
-            `${label} ${element[info.property]}`;
-        }
-        if (label !== undefined) {
-          return label;
-        }
-      }
+  (schema: JsonSchema7) => (element: any): string => {
```
eneufeld: should we remove all the JsonSchema7 stuff and replace it with JsonSchema?

PR #1057, opened 2018-08-28
Fix error when resovling anyOf containing $refs

edgarmueller:
Provide `labelProvider` to `AddtItemDialog`, adapt example code

### DIFF
--- a/packages/material-tree-renderer/src/tree/AddItemDialog.tsx
+++ b/packages/material-tree-renderer/src/tree/AddItemDialog.tsx
@@ -31,6 +31,7 @@ export interface AddItemDialogProps {
   dialogProps: any;
   setSelection: any;
   add?: any;
+  labelProvider?(schema: JsonSchema): (data?: any) => string;
 }
 
 class AddItemDialog extends React.Component<AddItemDialogProps, {}> {
@@ -46,7 +47,8 @@ class AddItemDialog extends React.Component<AddItemDialogProps, {}> {
       /**
        * Self contained schemas of the corresponding schema
        */
-      containerProperties
+      containerProperties,
+        labelProvider
     } = this.props;
 
     return (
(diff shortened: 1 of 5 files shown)


packages/material-tree-renderer/src/tree/AddItemDialog.tsx
```diff
@@ -113,7 +117,8 @@ const mapStateToProps = (state, ownProps) => {
     schema: ownProps.schema,
     closeDialog: ownProps.closeDialog,
     dialogProps: ownProps.dialogProps,
-    setSelection: ownProps.setSelection
+    setSelection: ownProps.setSelection,
+      labelProvider: ownProps.labelProvider
```
AhmetTanakol: Is there an extra space?

packages/material-tree-renderer/src/tree/AddItemDialog.tsx
```diff
@@ -86,7 +88,9 @@ class AddItemDialog extends React.Component<AddItemDialogProps, {}> {
                       closeDialog();
                     }}
                   >
-                    <ListItemText primary={prop.label} />
+                    <ListItemText
+                        primary={`${prop.property} [${(labelProvider && labelProvider(prop.schema)()) || prop.label}]`}
```
AhmetTanakol: Do we really want to display the property that the label is derived from?

packages/material-tree-renderer/test/services/property.util.test.ts
```diff
@@ -352,6 +352,77 @@ describe('Property util', () => {
...
+    test('expect to retrieve properties from object with array object by using keyword anyOf recursive', () => {
+        const schema: JsonSchema = {
+            definitions: {
+                a: {
+                    anyOf: [
+                        {$ref: '#/definitions/b'},
+                        {$ref: '#/definitions/a'}
+                    ]
+                },
+                b: { type: 'object', properties: { foo: { type: 'string' } } }
```
AhmetTanakol: Maybe we can make this example a bit more complex by adding the `anyOf` definition of `a` into `b` 's properties. We can create circular references between definitions, then we can check whether any of these definitions are visited more than once. What do you think?

PR #1068, opened 2018-08-31
Provide renderers for allOf/anyOf/oneOf and objects

edgarmueller:
This is a draft for implementing `allOf/anyOf/oneOf`renderers for the material renderer set. Any comments welcome.

### DIFF
--- a/packages/material/src/complex/MaterialObjectRenderer.tsx
+++ b/packages/material/src/complex/MaterialObjectRenderer.tsx
@@ -3,6 +3,7 @@ import * as _ from 'lodash';
 import {
     ControlElement,
     ControlProps,
+    Generate,
     isObjectControl,
     JsonSchema,
     mapStateToControlProps,
@@ -17,13 +18,7 @@ import { MaterialLayoutRenderer } from '../util/layout';
 
 const createControls = (schema: JsonSchema, path: string): UISchemaElement[] => {
     const obj =  resolveSchema(schema, path);
-    return Object.keys(obj).map(key => {
-        return {
-            label: key,
-            type: 'Control',
-            scope: `${path}/${key}`
-        };
-    });
+    return Object.keys(obj).map(key => Generate.controlElement(key, `${path}/${key}`));
 };
 
 class MaterialObjectRenderer extends React.Component<ControlProps, any> {
--- a/packages/material/src/complex/MaterialObjectRenderer.tsx
+++ b/packages/material/src/complex/MaterialObjectRenderer.tsx
@@ -1,61 +1,63 @@
 import * as React from 'react';
-import * as _ from 'lodash';
 import {
-    ControlElement,
     ControlProps,
-    Generate,
+    findUISchema,
     isObjectControl,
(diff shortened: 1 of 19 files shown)


packages/material/src/complex/MaterialObjectRenderer.tsx
```diff
@@ -0,0 +1,67 @@
...
+    render() {
+        const {
+            uischema,
+            schema,
+            visible,
+           } = this.props;
+
+        const controlElement = uischema as ControlElement;
+        const scope = controlElement.scope;
+        const elements: UISchemaElement[] = createControl(schema, (`${scope || '#'}/properties`));
```
eneufeld: we should retrieve or at least generate the elements using one of the existing functions instead of doing this with custom code

packages/material/src/complex/MaterialObjectRenderer.tsx
```diff
@@ -0,0 +1,67 @@
...
+        const style: {[x: string]: any} = { marginBottom: '10px' };
+        if (!visible) {
+            style.display = 'none';
+        }
+
+        return (
+            <Card style={style}>
+                {!_.isEmpty(controlElement.label) && <CardHeader title={controlElement.label}/>}
+                <CardContent>
+                    <MaterialLayoutRenderer
```
eneufeld: why not delegate to jsonforms similar to the allof renderer?

packages/material/src/complex/MaterialObjectRenderer.tsx
edgarmueller: I've exposed `createControlElement` from the UI schema generator. Does this address your concern?

packages/material/src/complex/MaterialObjectRenderer.tsx
edgarmueller: Usage of the `JsonForms` component is not applicable here, since we have to deal with an array of UI schema elements.

edgarmueller:
@eneufeld I' ve updated the PR, let me know if you have any other concers.

packages/material/src/complex/MaterialObjectRenderer.tsx
eneufeld: why can't we simply generate a uischema or use the uischemaservice as we do in the tree?

edgarmueller: What do you have in mind beside the usage of the `createControlElement` function?

eneufeld: I could imagine that you would want to provide a full uischema and not let it generate.
I don't really see the difference here between the detail view of the tree and the inline view of the object renderer

eneufeld:
I suggest to merge this, and open a new issue to discuss possible features of the object renderer there

edgarmueller:
I've created a [follow-up issue](https://github.com/eclipsesource/jsonforms/issues/1076) for tracking any remaining work, such that we can merge this one.

PR #1070, opened 2018-09-07
Add reducer for registering default data

edgarmueller:
Introduce a new reducer for adding a substate that keeps track of any pre-defined data that can be queried by clients. Currently, the only such client is the `AddItemDialog` which reads the registry and merges any result with any default values found in the schema, if such.
Registration is happening via a given schema path and a JSON value. No validation of the registered data or whatsoever is taking place.
The material-tree-renderer example has been updated to demonstrate this new feature.


packages/core/src/actions/index.ts
```diff
@@ -56,6 +59,20 @@ export const init = (
       uischema
     });
 
+export const registerDefaultData = (
+    schemaPath: string,
+    data: any
```
eneufeld: wouldn't it be better to have a function here, which gets the schema default (if available) and returns a new value to use?
As of know I can only provide exactly one static value for a path, with a function I could make this value dynamic based on whatever information available to me.

edgarmueller: Yes, true, but that would introduce additional functions into the store's state, which we should avoid IMHO (although I know that we do the same with testers already).

eneufeld:
I suggest to merge this and create a new issue in order to discuss how we can improve the way default data can be provided to jsonforms.

PR #1087, opened 2018-09-27
Feat/angular material/boolean tests

eneufeld:
~This is based on #1081~ 
~The new tests are failing we should discuss what we want to fix~

Tests should be running now and I rebased on master

### DIFF
--- a/packages/angular-material/test/boolean-control.spec.ts
+++ b/packages/angular-material/test/boolean-control.spec.ts
@@ -141,7 +141,7 @@ describe('Boolean control', () => {
         mockSubStore.complete();
         fixture.detectChanges();
         component.ngOnInit();
-        expect(component.value).toBe(true);
+        expect(component.data).toBe(true);
         expect(checkboxInstance.checked).toBe(true);
         expect(checkboxInstance.disabled).toBe(false);
     });
@@ -163,7 +163,7 @@ describe('Boolean control', () => {
         fixture.detectChanges();
         component.ngOnInit();
 
-        expect(component.value).toBe(true);
+        expect(component.data).toBe(true);
         expect(checkboxInstance.checked).toBe(true);
 
         const spy = spyOn(component, 'onChange');
@@ -187,15 +187,12 @@ describe('Boolean control', () => {
                 }
             }
         });
-        mockSubStore.complete();
         fixture.detectChanges();
         component.ngOnInit();
-        expect(component.value).toBe(true);
+        expect(component.data).toBe(true);
         expect(checkboxInstance.checked).toBe(true);
 
-        MockNgRedux.reset();
-        const mockSubStore2 = MockNgRedux.getSelectorStub();
-        mockSubStore2.next({
+        mockSubStore.next({
             jsonforms: {
                 core: {
                     data: { foo: false },
@@ -203,9 +200,9 @@ describe('Boolean control', () => {
                 }
             }
         });
-        mockSubStore2.complete();
-        component.subscribe();
-        expect(component.value).toBe(false);
+        mockSubStore.complete();
+        fixture.detectChanges();
+        expect(component.data).toBe(false);
         expect(checkboxInstance.checked).toBe(false);
     });
     it('should update with undefined value', () => {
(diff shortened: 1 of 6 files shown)


packages/angular-material/test/boolean-control.spec.ts
```diff
@@ -22,15 +22,73 @@
...
+import { NOT_APPLICABLE } from '@jsonforms/core';
+import { BooleanControlRenderer, booleanControlTester } from '../src';
 
+const uischema = {
+    type: 'Control',
+    scope: '#/properties/foo'
+};
+
+describe('Material boolean field tester', () => {
+    it('should fail', () => {
```
edgarmueller: I don't think we need the 'it should fail' spec, as we already test these in `testers.test.ts`

packages/angular-material/test/boolean-control.spec.ts
```diff
@@ -41,18 +99,16 @@ describe('Boolean control', () => {
...
-        scope: '#/properties/foo'
-    };
 
-    beforeEach(async (() => {
+    const getImports = () => [
+        MatCheckboxModule,
+        MatFormFieldModule
+    ];
+
+    beforeEach(async(() => {
```
edgarmueller: Do we need the `async`?

packages/angular-material/test/boolean-control.spec.ts
```diff
@@ -96,23 +160,181 @@ describe('Boolean control', () => {
...
         });
         mockSubStore.complete();
+        fixture.detectChanges();
         component.ngOnInit();
-        MockNgRedux.reset();
-        const mockSubStore2 = MockNgRedux.getSelectorStub();
-        mockSubStore2.next({
+
+        expect(component.data).toBe(true);
+        expect(checkboxInstance.checked).toBe(true);
```
edgarmueller: I don't think we necessarily have to expect these (here and in the following tests) as we already covered that case

packages/angular-material/test/boolean-control.spec.ts
```diff
@@ -96,23 +160,181 @@ describe('Boolean control', () => {
...
+        });
+        fixture.detectChanges();
+        component.ngOnInit();
+        expect(component.data).toBe(true);
+        expect(checkboxInstance.checked).toBe(true);
+
+        mockSubStore.next({
+            jsonforms: {
+                core: {
+                    data: { foo: true, bar: false },
```
edgarmueller: I'm not sure how the test name relates to the `data` here. Mind to clarify?

packages/angular-material/test/boolean-control.spec.ts
```diff
@@ -96,23 +160,181 @@ describe('Boolean control', () => {
...
+        fixture.detectChanges();
+        expect(component.data).toBe(true);
+        expect(checkboxInstance.checked).toBe(true);
+    });
+    it('can be disabled', () => {
+        const mockSubStore = MockNgRedux.getSelectorStub();
+        component.uischema = uischema;
+        component.disabled = true;
+
+        mockSubStore.next({
```
edgarmueller: We don't think we need the store at all in this test case.

packages/angular-material/test/boolean-control.spec.ts
```diff
@@ -96,23 +160,181 @@ describe('Boolean control', () => {
...
+        component.ngOnInit();
+        expect(checkboxInstance.disabled).toBe(true);
+
+    });
+    it('id should be present in output', () => {
+        const mockSubStore = MockNgRedux.getSelectorStub();
+        component.uischema = uischema;
+        component.id = 'myId';
+
+        mockSubStore.next({
```
edgarmueller: Here as well.

packages/angular-material/test/boolean-control.spec.ts
eneufeld: it makes sure that the wiring was done correctly, but I'm all for removing those tests, the hard question is what do we need to retest when testing wiring and what not

packages/angular-material/test/boolean-control.spec.ts
eneufeld: yes, we could leave them out

packages/angular-material/test/boolean-control.spec.ts
eneufeld: the idea is to test, that only changes on the relevant field are reflected, probably doesn't make much sense here, and should only be tested for the mapToControlProps

packages/angular-material/test/boolean-control.spec.ts
eneufeld: probably not, I will recheck

eneufeld:
@edgarmueller should I do the fixes here or would you be fine with a clean up commit?

edgarmueller:
@eneufeld Yes, I'm fine with that.

PR #1092, opened 2018-10-01
Add Angular Material Test case

eneufeld:
Based on #1089


packages/angular/src/control.ts
```diff
@@ -98,4 +97,11 @@ export class JsonFormsControl extends JsonFormsBaseRenderer implements OnInit, O
         const dispatch = mapDispatchToControlProps(this.ngRedux.dispatch);
         return {...props, ...dispatch};
     }
+
+    private triggerValidation() {
```
edgarmueller: I'm not sure whether thsi also applies to angular, does it?

PR #1093, opened 2018-10-01
Feat/angular material/unify text test

eneufeld:
based on #1092

### DIFF
--- a/packages/angular-test/src/text.ts
+++ b/packages/angular-test/src/text.ts
@@ -23,7 +23,7 @@ export const textTest = <C extends JsonFormsControl>(
             type: 'object',
             properties: {
                 foo: {
-                    type: 'text'
+                    type: 'string'
                 }
             }
         };
--- a/packages/angular-test/src/text.ts
+++ b/packages/angular-test/src/text.ts
@@ -10,11 +10,13 @@ export const textTest = <C extends JsonFormsControl>(
     imports: any[],
     providers: any[],
     componentUT: Type<C>,
-    errorTestInformation: ErrorTestInformation) => () => {
+    instance: string,
+    errorTestInformation: ErrorTestInformation,
+    elementToUse: (element: DebugElement) => any) => () => {
         let fixture: ComponentFixture<any>;
-        let textNativeElement: HTMLInputElement;
-        let component: C;
         let textElement: DebugElement;
+        let textNativeElement;
+        let component: C;
 
         const data = { foo: 'foo' };
         const schema: JsonSchema = {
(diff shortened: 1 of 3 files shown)


packages/angular-test/src/text.ts
```diff
@@ -10,11 +10,13 @@ export const textTest = <C extends JsonFormsControl>(
...
+    elementToUse: (element: DebugElement) => any) => () => {
         let fixture: ComponentFixture<any>;
-        let textNativeElement: HTMLInputElement;
-        let component: C;
         let textElement: DebugElement;
+        let textNativeElement;
+        let component: C;
 
         const data = { foo: 'foo' };
         const schema: JsonSchema = {
```
edgarmueller: I missed this before, but on line 26 foo has a type of 'text', which should be 'string', I think. Which brings me to the question whether this `schema` has any actual effect on the tests. Can we clarify that?

packages/angular-test/src/text.ts
```diff
@@ -250,7 +254,7 @@ export const textTest = <C extends JsonFormsControl>(
             mockSubStore.complete();
             fixture.detectChanges();
             component.ngOnInit();
-            expect(textNativeElement.id).toBe('myId');
+            expect(textElement.nativeElement.id).toBe('myId');
```
edgarmueller: Shouldn't we use `textNativeElement` here?

packages/angular-test/src/text.ts
```diff
@@ -10,11 +10,13 @@ export const textTest = <C extends JsonFormsControl>(
     imports: any[],
     providers: any[],
     componentUT: Type<C>,
-    errorTestInformation: ErrorTestInformation) => () => {
+    instance: string,
+    errorTestInformation: ErrorTestInformation,
+    elementToUse: (element: DebugElement) => any) => () => {
```
edgarmueller: Minor: Use consistent naming for parameter `elementToUse` and arguments `toSelect`, maybe `nativeElementSelector` or similar.

packages/angular-test/src/text.ts
eneufeld: no as textNativeElement could be in a wrapper and we can't be sure that the propagation is working as expected, eg for angular-material your suggestions works, but for ionic it fails

packages/angular-test/src/text.ts
eneufeld: I would add this to #1100 and do a full clean up

packages/angular-test/src/text.ts
edgarmueller: We shoud improve naming then (maybe via #1100), I'd think that both `textNativeElement` and `textElement.nativeElement` are likely to be used as synonymous.

packages/angular-test/src/text.ts
eneufeld: should be fixed with 2 commit

PR #1102, opened 2018-10-04
Fixes #1090: Rule examples are potentially broken

### LINKED ISSUE #1090: Rule examples are potentially broken
Conditions in rules need to set the type property, e.g. "type": "LEAF". The react seed was not updated correctly and therefore in the UI Schema the type property was not set for a condition:
https://github.com/eclipsesource/jsonforms-react-seed/pull/21
This might be the case in other seeds, too.

edgarmueller:
Use schema based conditions instead of leaf condiation for examples.

### DIFF
--- a/packages/examples/src/day4.ts
+++ b/packages/examples/src/day4.ts
@@ -59,7 +59,7 @@ export const uischema = {
           effect: 'HIDE',
           condition: {
               scope: '#/properties/recurrence',
-              expectedValue: 'Never'
+              schema: { enum: ['Never'] }
           }
       }
     }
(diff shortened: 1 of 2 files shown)


edgarmueller:
@eneufeld Do you have anything specific on your mind?

eneufeld:
If you check how we decide whether the condition is evaluated using schema or an expected value, you will see, that it is inkonsistent. 
For the schema we check whether the schema parameter is there
for the expectedvalue we check for the condition type, which we don't check otherwise.
I would propose to check for the expectedvalue property directly.

PR #1109, opened 2018-10-12
Add ionic categorization tab layout

edgarmueller:
* Updates Ionic `PlatformMock`
* Fix minor issues like fixing missing type annotations, usage of own props in layouts & usage of pipe operators


packages/ionic/src/layouts/categorization/categorization-menu-layout.ts
```diff
@@ -51,7 +51,7 @@ import { JsonFormsIonicLayout } from '../JsonFormsIonicLayout';
       </ion-content>
   `
 })
-export class CategorizationLayoutRenderer extends JsonFormsIonicLayout {
+export class CategorizationMenuLayoutRenderer extends JsonFormsIonicLayout {
```
eneufeld: Is this renderer event needed ? the tab based one will always win ...

edgarmueller: I think we should keep it for now, but the default should be the tab one.

PR #1112, opened 2018-10-15
Added Slider renderer for Angular Material

eneufeld:
~Based #1111~

### DIFF
--- a/packages/angular-test/src/range.ts
+++ b/packages/angular-test/src/range.ts
@@ -1,0 +1,382 @@
+import { DebugElement, Type } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { MockNgRedux } from '@angular-redux/store/testing';
+
+import { JsonFormsControl } from '@jsonforms/angular';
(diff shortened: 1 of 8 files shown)


packages/angular-test/src/range.ts
```diff
@@ -0,0 +1,382 @@
...
+                                 false, /* ctrlKey */
+                                 false, /* altKey */
+                                 false, /* shiftKey */
+                                 false, /* metaKey */
+                                 0, /* button */
+                                 null /* relatedTarget */);
+            // `initMouseEvent` doesn't allow us to pass the `buttons` and
+            // defaults it to 0 which looks like a fake event.
+            Object.defineProperty(event, 'buttons', { get: () => 1 });
+            rangeNativeElement.dispatchEvent(event);
```
edgarmueller: Could this be simplified as in `fixture.debugElement.query(By.css('.mat-slider-wrapper')).nativeElement.click()` ?

eneufeld: I will try this out and comment with the results

eneufeld:
@edgarmueller  tried out the suggestion, seem to work

PR #1113, opened 2018-10-16
Date Control Renderer for Angular Material

eneufeld:
~based on #1112~

### DIFF
--- a/packages/angular-material/src/controls/date.renderer.ts
+++ b/packages/angular-material/src/controls/date.renderer.ts
@@ -15,10 +15,10 @@ import { JsonFormsControl } from '@jsonforms/angular';
         placeholder="{{ description }}"
         [id]="id"
         [formControl]="form"
-        [matDatepicker]="myDatepicker"
+        [matDatepicker]="datepicker"
     >
-    <mat-datepicker-toggle matSuffix [for]="myDatepicker"></mat-datepicker-toggle>
-    <mat-datepicker #myDatepicker></mat-datepicker>
+    <mat-datepicker-toggle matSuffix [for]="datepicker"></mat-datepicker-toggle>
+    <mat-datepicker #datepicker></mat-datepicker>
 
     <mat-error>{{ error }}</mat-error>
   </mat-form-field>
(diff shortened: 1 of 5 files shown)


packages/angular-material/src/controls/date.renderer.ts
```diff
@@ -0,0 +1,38 @@
...
+        matInput
+        (dateChange)="onChange($event)"
+        [value]="data"
+        placeholder="{{ description }}"
+        [id]="id"
+        [formControl]="form"
+        [matDatepicker]="myDatepicker"
+    >
+    <mat-datepicker-toggle matSuffix [for]="myDatepicker"></mat-datepicker-toggle>
+    <mat-datepicker #myDatepicker></mat-datepicker>
```
edgarmueller: I'd change the ref name

packages/angular-material/test/date-control.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+        expect(component.data).toBe('2018-01-01');
+        expect(inputElement.value).toBe('1/1/2018');
+
+        const cells = document.querySelectorAll('.mat-calendar-body-cell');
+        const firstCell: HTMLElement = cells[1] as HTMLElement;
+        firstCell.click();
+        fixture.detectChanges();
+        flush();
+
+        if (inputElement.dispatchEvent) {
```
edgarmueller: Why do we need the condition here?

packages/angular-material/test/date-control.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+        fixture.detectChanges();
+        component.ngOnInit();
+
+        const spy = spyOn(component, 'onChange');
+
+        fixture.debugElement.query(By.directive(MatDatepicker)).componentInstance.open();
+        fixture.detectChanges();
+        flush();
+
+        expect(component.data).toBe('2018-01-01');
```
edgarmueller: I'd remove these expectatations.

packages/angular-material/src/controls/date.renderer.ts
eneufeld: do you whether the ref name must be unique across components, or only unique inside a component?

eneufeld: As I understand this article https://blog.angular-university.io/angular-viewchild/#visibilityscopeofviewchildtemplatequeries correctly, then the view ref is bound to the component scope, and thus this should be fine

PR #1114, opened 2018-10-16
Feat/angular material/label

eneufeld:
~Based on #1113~

### DIFF
--- a/packages/angular-material/test/label-renderer.spec.ts
+++ b/packages/angular-material/test/label-renderer.spec.ts
@@ -1,0 +1,94 @@
+/*
+  The MIT License
+  
+  Copyright (c) 2018 EclipseSource Munich
+  https://github.com/eclipsesource/jsonforms
+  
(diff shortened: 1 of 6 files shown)


packages/angular-material/test/label-renderer.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+                core: {
+                    data: data,
+                    schema: schema,
+                }
+            }
+        });
+        mockSubStore.complete();
+        fixture.detectChanges();
+        component.ngOnInit();
+        // auto? shown with US layout
```
edgarmueller: Misplaced comment?

eneufeld: yes this should have been part of the date test, i will clean this up

eneufeld: yes this should have been part of the date test, i will clean this up

PR #1115, opened 2018-10-17
Add master/detail tests

edgarmueller:


packages/angular/src/control.ts
```diff
@@ -55,7 +56,9 @@ export class JsonFormsControl extends JsonFormsBaseRenderer implements OnInit, O
     ngOnInit() {
         this.subscription = this.ngRedux
             .select()
-            .map((s: JsonFormsState) => this.mapToProps(s))
+            .pipe<ControlProps>(
```
eneufeld: what do we gain using the pipe?

packages/ionic/test/horizontal-layout.spec.ts
```diff
@@ -62,17 +62,15 @@ describe('Horizontal layout', () => {
             ],
             providers: [
                 {provide: Platform, useClass: PlatformMock},
-                {provide: NgRedux, useFactory: MockNgRedux.getInstance}
+                {provide: NgRedux, useFactory: MockNgRedux.getInstance},
+                {provide: Nav, useFactory: mockNav }
```
eneufeld: do we need this in the horizontal test?

packages/ionic/test/master-detail.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+                {provide: NgRedux, useFactory: MockNgRedux.getInstance},
+                {provide: Nav, useValue: mockNav()},
+                MasterDetailNavService,
+            ],
+        }).compileComponents();
+
+        MockNgRedux.reset();
+    }));
+
+    beforeEach(() => {
```
eneufeld: I think we can merge the two beforeeach, similar to the horizontal layout test

packages/angular/src/control.ts
edgarmueller: Avoiding 'unused imports' warnings

PR #1116, opened 2018-10-17
Feat/angular material/toggle

eneufeld:
Based on #1113


packages/angular-material/src/controls/toggle.renderer.ts
```diff
@@ -0,0 +1,63 @@
...
+    template: `
+    <div>
+        <mat-slide-toggle
+            (change)="onChange($event)"
+            [checked]="isChecked()"
+            [disabled]="!enabled"
+            [id]="id">
+            {{ label }}
+        </mat-slide-toggle>
+        <mat-error>{{ error }}</mat-error>
```
edgarmueller: Probably offtopic, but one thing I noted during testing is that the font used for errors is not the same as for the other controls.

PR #1119, opened 2018-10-18
Added Categorization Renderer for Angular Material

eneufeld:
* added renderer
* added test


eneufeld:
I somehow missed to add the material tabs module to the module.
should be fixed now

PR #1123, opened 2018-10-23
Angular material/list with detail

edgarmueller:


packages/angular-material/test/group-layout.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    });
+    mockSubStore.complete();
+    fixture.detectChanges();
+    component.ngOnInit();
+    const card: DebugElement[] = fixture.debugElement.queryAll(By.directive(MatCard));
+    // title
+    expect(card[0].nativeElement.children.length).toBe(1);
+  });
+
+  it('render with children', () => {
```
eneufeld: wouldn't it make sense to assert the label?

packages/angular-material/test/vertical-layout.spec.ts
```diff
@@ -25,109 +25,113 @@
...
-      declarations: [
-        VerticalLayoutRenderer,
-        JsonFormsOutlet
-      ],
-      imports: [],
-      providers: [
-        { provide: NgRedux, useFactory: MockNgRedux.getInstance }
-      ]
+    beforeEach(() => {
+        TestBed.configureTestingModule({
```
eneufeld: we could extract this, as it is the same for group, horizontal and vertical

packages/angular-material/test/vertical-layout.spec.ts
```diff
@@ -25,109 +25,113 @@
...
-    component = fixture.componentInstance;
-  });
 
-  it('render with undefined elements', () => {
-    const uischema: UISchemaElement = {
-      type: 'VerticalLayout'
-    };
-    const mockSubStore = MockNgRedux.getSelectorStub();
-    component.uischema = uischema;
+    it('render with undefined elements', () => {
```
eneufeld: all three tests are also similar.
only group needs a 4 test to check for the correct label.

packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    constructor(
+        ngRedux: NgRedux<JsonFormsState>,
+        private masterDetailService: MasterDetailService
+    ) {
+        super(ngRedux);
+        this.selectedItem = this.masterDetailService.getSelectedItem();
+    }
+
+    ngOnInit() {
```
eneufeld: why not use mapAdditionalProps

packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const removeSchemaKeywords = (path: string) => {
+    return path.split('/').filter(s => !_.some(keywords, key => key === s)).join('.');
+};
+
+@Component({
+    selector: 'jsonforms-list-with-detail-master',
+    template: `
+    <div fxLayout='row' fxLayoutGap='16px'>
+        <div flFlex>
```
eneufeld: I would suggest to use https://v5.material.angular.io/components/sidenav/overview instead of plain divs
Even when using plain divs, the full width of the screen should be used imho

PR #1133, opened 2018-10-29
Improve nested array renderer

eneufeld:


eneufeld:
This is only a suggestion, I'm open for suggestions

edgarmueller:
#1146 was based on this, hence closing this now.

PR #1134, opened 2018-10-29
[ionic] Add support for formats

edgarmueller:


packages/ionic/test/text-control.spec.ts
```diff
@@ -66,3 +75,62 @@ describe('Text control Base Tests', textBaseTest(testConfig, 'ion-input', toSele
...
+                schema
+            });
+            mockSubStore.next({
+                jsonforms: {
+                    core: {
+                        data: defaultTextTestData.data,
+                        schema,
+                        errors: [{
+                            dataPath: 'foo',
+                            message: `should match format "email"`
```
eneufeld: why do we need to test for erros here, isn't error checking part of the error tests (`textErrorTest(testConfig, errorTest)`)?

edgarmueller: Fixed

PR #1135, opened 2018-10-29
[material] Export util from material

edgarmueller:


packages/material/src/util/layout.tsx
```diff
@@ -25,13 +25,13 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import {
-    JsonSchema,
+    JsonSchema, OwnPropsOfRenderer,
```
eneufeld: formatting

edgarmueller: Fixed

PR #1136, opened 2018-10-29
Remove jsonforms/webcompomnent dev dep from packages

edgarmueller:
Although custom element support has been added to FF (https://blog.nightly.mozilla.org/2018/09/06/developer-tools-support-for-web-components-in-firefox-63/) there's no point in keeping needless dependencies.


eneufeld:
@edgarmueller build is failing

PR #1138, opened 2018-10-30
Added Object renderer

eneufeld:


eneufeld:
I skipped one test as we didn't decide on whether or how we want to support it.

packages/angular-material/test/object-control.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+    });
+    mockSubStore.complete();
+    fixture.detectChanges();
+    component.ngOnInit();
+    fixture.whenStable().then(() => {
+      expect(fixture.nativeElement.querySelector('input').disabled).toBe(true);
+    });
+  }));
+  it('should be enabled by default', async(() => {
```
edgarmueller: Does this test actually test something, or put another way: can an object renderer propagate a disabled state to its children?

packages/angular-material/test/object-control.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    mockSubStore.complete();
+    fixture.detectChanges();
+
+    fixture.whenRenderingDone().then(() => {
+      fixture.detectChanges();
+      expect(fixture.nativeElement.querySelectorAll('input').length).toBe(2);
+    });
+  }));
+
+  it('render only itself elements', async(() => {
```
edgarmueller: Sloppy test name

packages/angular-material/test/object-control.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  })
+  .compileComponents();
+
+    MockNgRedux.reset();
+    fixture = TestBed.createComponent(ObjectControlRenderer);
+    component = fixture.componentInstance;
+  }));
+
+  it('render all elements', async(() => {
+    const mockSubStore = MockNgRedux.getSelectorStub();
```
edgarmueller: If the `setupMockStore` method would receive the next state instead of the triplet of schema/UI schema/data, we could re-use it here to avoid some boilerplate

packages/angular-material/src/module.ts
```diff
@@ -110,7 +114,8 @@ import {
     LabelRenderer,
     CategorizationTabLayoutRenderer,
     MasterListComponent,
-    JsonFormsDetailComponent
+    JsonFormsDetailComponent,
+    ObjectControlRenderer
   ],
   schemas: [CUSTOM_ELEMENTS_SCHEMA],
```
edgarmueller: Off-topic, but do we still need `CUSTOM_ELEMENTS_SCHEMA`  to be registered?

packages/angular-material/test/object-control.spec.ts
eneufeld: I think it would be great if this would work, but I'm not sure whether it is doable and whether we want to support it.

packages/angular-material/src/module.ts
eneufeld: I would recheck after this is merged.

packages/angular-material/test/object-control.spec.ts
eneufeld: We should double check the tests anyway. I'll open an issue to recheck this.

PR #1148, opened 2018-11-09
Add ionic range control

edgarmueller:
Furthermore:
* Re-organize angular-test structure
* Fix rule evaluation
* Make `componentRef` lazily initialized 
* Fix group layout label property

### DIFF
--- a/packages/core/src/util/renderer.ts
+++ b/packages/core/src/util/renderer.ts
@@ -305,9 +305,14 @@ export interface ControlState {
  */
 export const mapStateToControlProps =
     (state: JsonFormsState, ownProps: OwnPropsOfControl): StatePropsOfControl => {
+        const { uischema } = ownProps;
         const path = composeWithUi(ownProps.uischema, ownProps.path);
-        const visible = _.has(ownProps, 'visible') ? ownProps.visible :  isVisible(ownProps, state);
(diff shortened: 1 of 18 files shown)


packages/core/src/util/renderer.ts
```diff
@@ -305,9 +305,14 @@ export interface ControlState {
  */
 export const mapStateToControlProps =
     (state: JsonFormsState, ownProps: OwnPropsOfControl): StatePropsOfControl => {
+        const { uischema } = ownProps;
         const path = composeWithUi(ownProps.uischema, ownProps.path);
-        const visible = _.has(ownProps, 'visible') ? ownProps.visible :  isVisible(ownProps, state);
-        const enabled = _.has(ownProps, 'enabled') ? ownProps.enabled :  isEnabled(ownProps, state);
+        const rulePath = _.has(ownProps.uischema, 'rule')
```
eneufeld: I assume you fixed this because of a bug?
if so we should add tests for this.
how can I reproduce the bug?

edgarmueller: It's related to the `ListWithDetail` renderer where a scope is set and a detail view with a rule has been specified as well. The rule's scope is based on a schema ref, but actually should be depenent on an instance path because of array indices. I added a test case demonstrating this.

eneufeld:
LGTM, thank you

PR #1152, opened 2018-11-13
[ionic] Add toggle renderer, rename checkbox renderer, rename test utils

edgarmueller:


packages/ionic/src/controls/boolean/boolean-toggle-control.scss
```diff
@@ -0,0 +1,3 @@
+jsonforms-toggle-control {
```
eneufeld: why do we need this?

edgarmueller: Currently not, I just added it for consistency reasons.

PR #1154, opened 2018-11-13
Remove template value bindings

edgarmueller:
Form value is set via `form` attribute. This does not apply to the ionic renderer set, e.g. because of https://forum.ionicframework.com/t/disable-enable-of-ion-input-using-angular-reactive-forms/116901 (for the `disabled` property). Also, removing the `value` property on string/multiline tests causes them to fail, hence I left it there for the time being.

### DIFF
--- a/packages/angular-material/src/controls/autocomplete.renderer.ts
+++ b/packages/angular-material/src/controls/autocomplete.renderer.ts
@@ -69,7 +69,6 @@ import { startWith } from 'rxjs/operators/startWith';
                 matInput
                 type="text"
                 (change)="onChange($event)"
-                [value]="getValue()"
                 placeholder="{{ description }}"
                 [id]="id"
                 [formControl]="form"
@@ -92,7 +91,6 @@ export class AutocompleteControlRenderer extends JsonFormsControl {
     constructor(ngRedux: NgRedux<JsonFormsState>) {
         super(ngRedux);
     }
-    getValue = () => this.data || '';
     getEventValue = (event: any) => event.target.value;
 
     ngOnInit() {
--- a/packages/angular-material/src/controls/date.renderer.ts
+++ b/packages/angular-material/src/controls/date.renderer.ts
@@ -11,7 +11,6 @@ import { JsonFormsControl } from '@jsonforms/angular';
     <input
         matInput
         (dateChange)="onChange($event)"
-        [value]="data"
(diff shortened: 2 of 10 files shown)


packages/angular-material/src/controls/boolean.renderer.ts
```diff
@@ -34,7 +34,6 @@ import { isBooleanControl, JsonFormsState, RankedTester, rankWith } from '@jsonf
         <mat-checkbox
             (change)="onChange($event)"
             [checked]="isChecked()"
-            [disabled]="!enabled"
```
eneufeld: I think we can't remove the disable here as the checkbox is not controlled by the reactive form control `form`

packages/angular-material/src/controls/toggle.renderer.ts
```diff
@@ -41,7 +41,6 @@ import {
         <mat-slide-toggle
             (change)="onChange($event)"
             [checked]="isChecked()"
-            [disabled]="!enabled"
```
eneufeld: I think we have the same issue here as with the checkbox renderer, the toggle is not controlled by the reactive control

packages/angular-test/src/boolean.ts
```diff
@@ -205,7 +204,7 @@ export const booleanBaseTest = <C extends JsonFormsControl, I>
         mockSubStore.complete();
         fixture.detectChanges();
         component.ngOnInit();
-        expect(checkboxInstance.disabled).toBe(true);
+        expect(component.form.disabled).toBe(true);
```
eneufeld: imho we must check the checkbox instance and not the component, see comment on the renderer

edgarmueller:
Ouch, right, fixed it.

packages/angular-test/src/boolean.ts
```diff
@@ -70,7 +70,6 @@ export const booleanBaseTest = <C extends JsonFormsControl, I>
         component.ngOnInit();
         expect(component.data).toBe(true);
         expect(checkboxInstance.checked).toBe(true);
-        expect(checkboxInstance.disabled).toBe(false);
```
eneufeld: why did you remove this, the idea was to check the default behavior is correct

edgarmueller: Restored as well..

PR #1159, opened 2018-11-14
Fix Angular MasterDetail Renderer

eneufeld:
* The layout renderer always provided the same rendererProps
to the jsonforms-outlet independent of the input changes.
To not rerender the layout on any change we need to fix the ngFor,
as we can't provide an useful trackBy function we reuse the array.
* The master-detail.service is unnecessary as we fixed the disposal.
* The master part didn't refresh item labels on changes in detail,
thus added the track by.


edgarmueller:
Tests seem to fail because of the the missing `MasterDetailService`. Also, I'd update the commit message as this actually about fixing the layout renderer, right?

eneufeld:
pushed fixed test case and updated commit message

PR #1172, opened 2018-11-21
Improved nested Array tester

### LINKED ISSUE #1169: "No applicable field found." in deeply nested schema
**Describe the bug**
In a deeply nested schema a valid JSON schema can fail to render form elements, instead presenting "No applicable field found.".

**To Reproduce**
Can be reproduced with the following Schema:
```
{
   "type":"object",
   "properties":{
      "Level1":{
         "type":"array",
         "items":{
            "type":"object",
            "properties":{
               "Level2":{
                  "type":"object",
                  "properties":{
                     "Level3":{
                        "type":"string"
                     }
                  }
               }
            }
         }
      }
   }
}
```
Interestingly if the bottom most element that is failing to render is nested further in an array, it does render correctly but not an object within an object.

**Expected behavior**
Text file to be rendered correctly.

 - Browser: Safari
 - Version: 12.0.1

 - Framework: React
 - RendererSet: Material

eneufeld:
Added Example and test case for problem
Fixed #1169

### DIFF
--- a/packages/core/src/testers/index.ts
+++ b/packages/core/src/testers/index.ts
@@ -345,14 +345,18 @@ export const isObjectArrayWithNesting =
     }
     const schemaPath = (uischema as ControlElement).scope;
     const resolvedSchema = resolveSchema(schema, schemaPath);
-    const minimalElementsMap: { [key: string]: number; } = { 'object': 1, 'array': 0 };
+    const wantedNestingByType: { [key: string]: number; } = { 'object': 2, 'array': 1 };
     return _.has(resolvedSchema, 'items') &&
       traverse(resolvedSchema.items, val => {
         if (val === schema) { return false; }
+        // we don't support multiple types
         if (typeof val.type !== 'string') { return true; }
-        const typeCount = minimalElementsMap[val.type];
+        const typeCount = wantedNestingByType[val.type];
         if (typeCount === undefined) { return false; }
-        if (typeCount === 0) { return true; } else { minimalElementsMap[val.type] = typeCount - 1; }
+        wantedNestingByType[val.type] = typeCount - 1;
+        if (wantedNestingByType[val.type] === 0) {
+          return true;
+        }
         return false;
       });
   };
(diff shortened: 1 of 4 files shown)


packages/core/src/testers/index.ts
```diff
@@ -345,17 +345,32 @@ export const isObjectArrayWithNesting =
     }
     const schemaPath = (uischema as ControlElement).scope;
     const resolvedSchema = resolveSchema(schema, schemaPath);
+    const minimalElementsMap: { [key: string]: number; } = { 'object': 1, 'array': 0 };
     return _.has(resolvedSchema, 'items') &&
-      traverse(resolvedSchema.items, val => val !== schema && _.has(val, 'items'));
+      traverse(resolvedSchema.items, val => {
+        if (val === schema) { return false; }
+        if (typeof val.type !== 'string') { return true; }
```
edgarmueller: Add comment to make it obvious why this guard is needed

packages/core/src/testers/index.ts
```diff
@@ -345,17 +345,32 @@ export const isObjectArrayWithNesting =
     }
     const schemaPath = (uischema as ControlElement).scope;
     const resolvedSchema = resolveSchema(schema, schemaPath);
+    const minimalElementsMap: { [key: string]: number; } = { 'object': 1, 'array': 0 };
```
edgarmueller: The name of `minimalElementsMap` is a bit misleading. Maybe rename this to `minNestingDepth`? What do you think?

packages/core/src/testers/index.ts
```diff
@@ -345,17 +345,32 @@ export const isObjectArrayWithNesting =
...
     return _.has(resolvedSchema, 'items') &&
-      traverse(resolvedSchema.items, val => val !== schema && _.has(val, 'items'));
+      traverse(resolvedSchema.items, val => {
+        if (val === schema) { return false; }
+        if (typeof val.type !== 'string') { return true; }
+        const typeCount = minimalElementsMap[val.type];
```
edgarmueller: Maybe rename `typeCount` to `depth`, `currDepth` or similar?

PR #1181, opened 2018-11-29
[feedback wanted] Extend core/i18n substates with actions to set schema and uischema

edgarmueller:
This PR introduces changes that are meant to be used in the following fashion:

During initialization one is ought to register the respective localized versions of the schema/UI schemas, e.g.:
```js
const localizedSchemas = new Map<string, JsonSchema>();
localizedSchemas.set('de-DE', deSchema);
localizedSchemas.set('en-US', usSchema);
store.dispatch(Actions.setLocalizedSchemas(localizedSchemas));
```
Likewise for UI schemas.

Then, within the locale switcher of the respective app (which needs to be custom anyway, I think), locales can be switched via the new actions (this could be probably be improved a bit since we need to pass the entire state ATM):

```js
changeLocale(locale: string) {
  store.dispatch(setLocale(locale));
  store.dispatch(setSchema(getLocalizedSchema(locale)(store.getState())));
  store.dispatch(setUISchema(getLocalizedUISchema(locale)(store.getState())));
}
```

This will cause the root schema/UI schema to be updated and hence the entire forms managed by JSON Forms will be re-rendered with the new schema/UI schema


eneufeld:
would it make sense to extend the examples in order to try this feature out?

PR #1186, opened 2018-12-03
String control is not used as fallback for controls anymore

### LINKED ISSUE #1103: Text Renderer is used as the fall back renderer for controls
**Describe the bug**
If you define a uischema with a control, then the fallback for all control renderers is the text renderer. This is not always the best solution and we should discuss whether it would be better to simply show the unknownRenderer instead.

eneufeld:
* Pins the tests in angular based renderers to isStringControl
* Cleanups the manual field dispatch in MaterialInputControl

Fixes #1103


edgarmueller:
Tests seem to fail.

PR #1190, opened 2018-12-10
Angular Material Tree styling & empty data label

eneufeld:


packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -49,7 +49,7 @@ export const removeSchemaKeywords = (path: string) => {
                <mat-icon aria-label="Add item to list">add</mat-icon>
            </button>
         </mat-sidenav>
-        <mat-sidenav-content class="content">
+        <mat-sidenav-content class="content" style="background-color:#fff">
```
edgarmueller: Moving this into the `styles` property would be a bit cleaner.

PR #1195, opened 2018-12-18
Add i18n for date & number controls

edgarmueller:
This PR also addresses other issues:
* a `ParamsService` for categories to support changing the UI schema/schema dynamically (hairy stuff, but I couldn't come up with something better, without triggering a re-render cycle)
* assignment of the `description` property in the angular-specific control
* Make all labels stacked (i.e. static at the top) in order to avoid overlap with potential description/placholders


packages/ionic/src/controls/number/number-control.ts
```diff
@@ -1,34 +1,39 @@
...
 @Component({
   selector: 'jsonforms-number-control',
   template: `
-      <ion-item no-padding no-lines>
-          <ion-label floating>{{label}}</ion-label>
+      <ion-item no-padding no-lines [hidden]="hidden">
+          <ion-label stacked>{{label}}</ion-label>
           <ion-label stacked *ngIf="error" color="error">{{error}}</ion-label>
           <ion-input
-                  type="number"
```
eneufeld: why not use the type 'string' ?

packages/ionic/src/controls/number/number-control.ts
```diff
@@ -1,34 +1,39 @@
...
                   placeholder="{{ description }}"
                   [id]="id"
+                  [value]="data !== undefined && data !== null && locale ?
```
eneufeld: maybe move this to a helper method, as the code is quite big?

packages/ionic/src/layouts/categorization/categorization-tab-layout.ts
```diff
@@ -29,44 +30,68 @@ import { CategoryRenderer } from './category/category';
...
+  categoryPages: any[];
 
-    constructor(ngRedux: NgRedux<JsonFormsState>) {
-        super(ngRedux);
-        this.categoryPages = [];
-        this.initializers.push(this.mapAdditionalProps);
-    }
+  constructor(ngRedux: NgRedux<JsonFormsState>, private paramsService: ParamsService) {
+    super(ngRedux);
+    this.categoryPages = [];
```
eneufeld: maybe add types?

packages/ionic/src/layouts/categorization/categorization-tab-layout.ts
```diff
@@ -29,44 +30,68 @@ import { CategoryRenderer } from './category/category';
...
-                                elements: category.elements,
-                                label: category.label,
-                                schema: this.schema,
-                                path: this.path
-                            }
-                        }
-                    });
-                }
+  mapAdditionalProps = (props: any) => {
+    if (this.categoryPages.length === 0) {
```
eneufeld: I don't get why we need to do so complicated stuff and can't simply rely on angular/ionic and the trackBy function to render the right stuff?

Issue #1168, opened 2018-11-16
Empty string should be converted to undefined


edgarmueller:
**Describe the bug**
We should consider converting empty strings to `undefined`, because in case of validation errors for non-required properties it can happen that you do not get rid of the error anymore.

**To Reproduce**
Create a schema with a non-required string property with an arbitrary minLength. Delete the respective value of the property in the generated form.

**Expected behavior**
When clearing the form, the validation message should disappear.

**Additional context**
Change must be done in the core

edgarmueller:
Fixed via #1197

PR #1197, opened 2018-12-19
Empty string should be converted to undefined #1168

AlexandraBuzila:


packages/core/src/reducers/core.ts
```diff
@@ -145,7 +145,11 @@ export const coreReducer = (
         };
       } else {
         const oldData: any = _.get(state.data, action.path);
-        const newData = action.updater(oldData);
+        let newData = action.updater(oldData);
+        if (newData === '' && newData.constructor.name === 'String') {
```
eneufeld: do we need the `newData.constructor.name === 'String' ` check?

edgarmueller: I don't think so

AlexandraBuzila: It seems like it's not needed, since only strings and dates produce an empty string when cleared and the data type for dates is string anyway. I pushed an updated commit.

Issue #1157, opened 2018-11-14
Ability to specify uischema of arrays inline


chrisjohnson:
**Is your feature request related to a problem? Please describe.**
Right now we are using jsonforms by storing a schema and a uischema in a database column for various forms we want to render. As it stands, the only way to apply uischema to array elements (consider an array of objects) is via the use of some extra javascript using registerUISChema where you specifically have to write testers that match when you get into a specific array and inject that array's uischema at the appropriate place. This is incompatible with our approach since aren't injecting extra javascript (and that doesn't really seem like an appropriate solution). Right now we are just living with the default generated uischema for array elements because it's the only viable option here.

**Describe the solution you'd like**
Some sort of mechanism for specifying the uischema inline to the overall uischema without having to call out to extra schemas. Perhaps borrowing from the json schema standard with some sort of $ref system? Or just the ability to specify a hunk of schema that will be pointing at an array and should repeat for each element in the array? I'm not quite sure the best solution here, because there would still need to be a way to add a new blank entry for adding elements to the array.

**Describe alternatives you've considered**
The only alternative is to just live with the default, or write some special case javascript using registerUISchema. The latter is not feasible so we are living with the former

**Describe for which setup you like to have the improvement**
Framework: react
RendererSet: material

edgarmueller:
This should be fixed via #1198. We'll still need to update the docs though

chrisjohnson:
<3

shivgarg5676:
Is there any documentation around this?

sdirix:
Hi @shivgarg5676 you can check https://github.com/eclipsesource/jsonforms/pull/1198 for a description. We'll add them to the official documentation at some point in the future.

PR #1198, opened 2018-12-19
Inline UiSchema for Arrays

sdirix:
UiSchemas for controls referring to arrays can now have an optional `options` attribute specifying how the array shall be rendered. Motivation: #1157 

## Default
```
options: {
    detail : 'DEFAULT'
}
```
The array will be rendered as before. The string is case insensitive.

## Inline UiSchema
```
options: {
    detail : {
        type: 'HorizontalLayout',
        ...
    }
}
```
The array will be rendered using the nested array renderer. The nested renderer will use the specified uischema to render the array elements.

## Generated UiSchema
```
options: {
    detail : 'GENERATED'
}
```
The array will be rendered using the nested array renderer. The nested renderer will use a generated uischema to render the array elements. The string is case insensitive.

## Registered / Generated UiSchema
```
options: {
    detail : 'REGISTERED'
}
```
The array will be rendered using the nested array renderer. The nested renderer will check if a uischema was registered for the type to be rendered or generate one itself. This case will be triggered if detail is any string besides `GENERATED` (case insensitiv) or `DEFAULT` (case insensitive).

### DIFF
--- a/packages/material/test/renderers/MaterialArrayLayout.test.tsx
+++ b/packages/material/test/renderers/MaterialArrayLayout.test.tsx
@@ -111,12 +111,53 @@ const nestedSchema2 = {
   }
 };
 
+const uischemaOptions: {
+  generate: ControlElement,
+  default: ControlElement,
+  inline: ControlElement,
+} = {
+  default: {
+    type: 'Control',
+    scope: '#',
+    options: {
+      detail : 'DEFAULT'
+    }
+  },
+  generate: {
+    type: 'Control',
+    scope: '#',
+    options: {
+      detail : 'GENERATE'
+    }
+  },
+  inline: {
+    type: 'Control',
+    scope: '#',
+    options: {
+      detail : {
+        type: 'HorizontalLayout',
+        elements: [
+          {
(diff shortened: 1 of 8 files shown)


packages/material/test/renderers/MaterialArrayLayout.test.tsx
```diff
@@ -129,4 +170,30 @@ describe('Material array layout', () => {
...
+    const store = initJsonFormsStore();
+    const tree = TestUtils.renderIntoDocument(
+      <Provider store={store}>
+        <MaterialArrayLayout schema={schema} uischema={uischemaOptions.generate}/>
+      </Provider>
+    );
+
+    const controls = TestUtils.scryRenderedDOMComponentsWithTag(tree, 'input');
+    // 1 data entries having 2 controls
+    expect(controls.length).toBe(4);
```
edgarmueller: Does the comment match the expectation?

sdirix: Good find. I accidentally tested the same case twice and only changed the comment.

PR #1209, opened 2019-01-04
Fix hasRef performance issues.

### LINKED ISSUE #1208: React performance problem with excessive calls to `hasRefs`
**Describe the bug**
When running with the `@jsonforms/react` there is a significant performance issue when rendering a form with a lot of fields. I believe I have tracked this down to [excessive calls to `hasRefs`](https://github.com/eclipsesource/jsonforms/blob/master/packages/react/src/JsonForms.tsx#L78). As I understand the code, that deferred call to resolve the refs on the schema, only needs to happen once on the root `JsonForms` component, as it will pass/provide the already resolve schema down to children, however, all of the children are also performing the `hasRefs` check, even though there will be no refs as the root `JsonForms` component has resolve them. `hasRefs` is an expensive call, especially when run on a large form. I was able to drop render time from 30s to 3 seconds by removing that call.

 - Framework: `react`
 - RendererSet: `material`

nowells:
When running with `@jsonforms/react` there is a significant performance issue when rendering a form with a lot of fields. I believe I have tracked this down to [excessive calls to `hasRefs`](https://github.com/eclipsesource/jsonforms/blob/master/packages/react/src/JsonForms.tsx#L78). As I understand the code, that deferred call to resolve the refs on the schema, only needs to happen once on the root `JsonForms` component, as it will pass/provide the already resolve schema down to children, however, all of the children are also performing the `hasRefs` check, even though there will be no refs as the root `JsonForms` component has resolve them. `hasRefs` is an expensive call, especially when run on a large form. I was able to drop render time from 30s to 3 seconds by removing that call.

Fixes #1208

### DIFF
--- a/packages/core/src/util/renderer.ts
+++ b/packages/core/src/util/renderer.ts
@@ -474,10 +474,12 @@ export const mapStateToLayoutProps =
 
 export interface JsonFormsProps extends StatePropsOfRenderer {
     renderers?: { tester: RankedTester, renderer: any }[];
+    isSubForm?: Boolean;
 }
 
 export interface StatePropsOfJsonFormsRenderer extends OwnPropsOfRenderer {
     renderers: JsonFormsRendererRegistryEntry[];
+    isSubForm: Boolean;
 }
 
 export const mapStateToJsonFormsRendererProps =
@@ -494,6 +496,7 @@ export const mapStateToJsonFormsRendererProps =
         return {
             renderers: _.get(state.jsonforms, 'renderers') || [],
             schema: ownProps.schema || getSchema(state),
-            uischema
+            uischema,
+            isSubForm: !!ownProps.schema,
         };
     };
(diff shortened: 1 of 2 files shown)


edgarmueller:
I like the change and the general idea of using schema to check whether we deal with a subform or not, but we should be aware of the fact that specifying the `schema` prop on the `JsonForms` component (as in the 'Multiple forms within single store' section in the [store docs](https://jsonforms.io/docs/store)) does not cause the schema to be resolved anymore, which is a change I can live with, but should be documented. Alternatively, we could maybe also think of replacing `isSubForm` with an explicit `resolveRefs` property which would default to false, but enables examples such as the `allOf` one, to work correctly by specifying true (this is rather just an idea, I'm not sure yet what implications this would have). @nowells @eneufeld What do you think?

nowells:
@edgarmueller another option that we might be able to go with is in `packages/react/src/JsonForms.tsx` we could split the `JsonForms` into two classes `SubJsonForms` and `JsonForms extends SubJsonForms` (or some better naming) where the `SubJsonForms` is what all renderers use and it is always provided a resolved `schema` and the external `JsonForms` class will always resolve the schema, but internally we never use that class again?

PR #1212, opened 2019-01-08
Update dependencies

edgarmueller:
Update necessary due to security warnings

### DIFF
--- a/packages/react/src/DispatchField.tsx
+++ b/packages/react/src/DispatchField.tsx
@@ -27,9 +27,9 @@ import { connect } from 'react-redux';
 import * as _ from 'lodash';
 import { UnknownRenderer } from './UnknownRenderer';
 import {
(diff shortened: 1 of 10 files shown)


packages/react/src/DispatchField.tsx
```diff
@@ -27,9 +27,9 @@ import { connect } from 'react-redux';
 import * as _ from 'lodash';
 import { UnknownRenderer } from './UnknownRenderer';
 import {
-    DispatchFieldProps,
-    formatErrorMessage,
-    mapStateToDispatchFieldProps,
+  DispatchFieldProps, DispatchFieldStateProps,
```
eneufeld: formatting, but should be fine, as it will be fixed by prettier

edgarmueller: Ok, great

PR #1214, opened 2019-01-09
Subclass JsonForms from ResolvedJsonForms.

nowells:
When running with @jsonforms/react there is a significant performance issue when rendering a form with a lot of fields. I believe I have tracked this down to excessive calls to hasRefs. As I understand the code, that deferred call to resolve the refs on the schema, only needs to happen once on the root JsonForms component, as it will pass/provide the already resolve schema down to children, however, all of the children are also performing the hasRefs check, even though there will be no refs as the root JsonForms component has resolve them. hasRefs is an expensive call, especially when run on a large form. I was able to drop render time from 30s to 3 seconds by removing that call.

This is an alternate solution to #1209 (and #1208) which takes the route of subclassing the JsonForms to allow renderers that are not at the root to explicitly include a JsonForms that already has the schema fully resolved and avoid the expensive penalties of hasRefs checks.

Fixes: #1209
Fixes: #1208

### DIFF
--- a/packages/material-tree-renderer/src/tree/TreeWithDetailRenderer.tsx
+++ b/packages/material-tree-renderer/src/tree/TreeWithDetailRenderer.tsx
@@ -15,7 +15,7 @@ import {
     Runtime, StatePropsOfControl,
     UISchemaElement,
 } from '@jsonforms/core';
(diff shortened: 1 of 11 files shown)


edgarmueller:
LGTM, thanks!

PR #1222, opened 2019-01-14
Pass ownProps.path to rule evaluation in mapStateToLayoutProps

### LINKED ISSUE #1220: Path does not work in rule
**Describe the bug**

When I put the path in JsonForms, the rule in uischema will still pick up the data from the root scope.

**To Reproduce**
1. Init JsonForms
```js
 store.dispatch(Actions.init(data, schema, uischema));
```

where data = { main: <---data itself---->}

So here, I have a path in data called main

2. To render JsonForms
```jsx
<JsonForms schema={schema} uischema={uischema} path='main'></JsonForms>
```

I specify the path - "main" here.

3. Here my uischema
```json
"type":"VerticalLayout",
"rule": {
    "effect": "SHOW",
    "condition": {
    "type": "LEAF",
    "scope": "#/properties/index",
    "expectedValue": 1
},
"elements": [
    {
        "type": "VerticalLayout",
        "elements": [
            {
                "type": "Control",
                 "scope": "#/properties/Name"
            }
    }
]
```

With the path, I can display the "Name" correctly. However, if I put a rule, the rule will NOT go to path (which is main) to check the value. It will result in not handling the rule correctly

**Expected behavior**
The element should show or hide based on the path but not the root.

**Browser:**
 - Browser: chrome
 - Version: 71

**Used Setup:**
 - Framework: react,
 - RendererSet: material

eneufeld:
* added example to reproduce the issue
* added test cases
* fixes #1220

### DIFF
--- a/packages/core/test/util/renderer.test.ts
+++ b/packages/core/test/util/renderer.test.ts
@@ -32,6 +32,7 @@ import {
   mapDispatchToControlProps,
   mapStateToControlProps,
   mapStateToJsonFormsRendererProps,
+  mapStateToLayoutProps,
   OwnPropsOfControl
 } from '../../src/util';
 import configureStore from 'redux-mock-store';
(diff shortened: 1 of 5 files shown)


packages/core/test/util/renderer.test.ts
```diff
@@ -460,3 +524,35 @@ test('mapDispatchToArrayControlProps should remove items from array', t => {
   t.is(store.getState().jsonforms.core.data.length, 1);
   t.is(store.getState().jsonforms.core.data[0], 'quux');
 });
+
+test('mapStateToLayoutProps - visible via state with path from ownProps ', t => {
```
edgarmueller: Maybe we should also a test case that expects visible to be false

eneufeld: yes we should do so

eneufeld: Added a commit providing the test case

Issue #1193, opened 2018-12-13
Preselect with default schema values


johnmorrell:
For enumerated properties presented as a drop down list, where a default value is specified in the schema, it would make sense to pre-select that value. For example, 'red' should be pre selected:

```
      "colours":{
         "type":"string",
         "enum":[
            "red",
            "green",
            "blue"
         ],
         "default":"red"
      }
```

edgarmueller:
@johnmorrell We have made changes that allow options to be passed in into the `createAjv` function. This enables us to use the `useDefaults: true` option, which should cover your use case (it's probably best if you take a look at the changes in #1224). Please let us know whether this solves your issue. This change will be part of the 2.2.0 release.

edgarmueller:
Please re-open if you experience any issues.

johnmorrell:
Thank you @edgarmueller, I have taken a look at this in v2.1.1-alpha.3

I can see that this is working for regular string types, but it does not appear to support the case where the string is of an enum type, eg:

```
"colours":{
  "type":"string",
    "enum":[
      "red",
      "green",
      "blue"
  ],
   "default":"red"
}
```

johnmorrell:
There is also an unusual behaviour when a default is set for basic strings (react/material).

If you select the field and try to delete the value, it deletes up to the last character then the default value returns. This is also the case if you select all of the text and attempt to delete.

edgarmueller:
Thanks for the update. We'll look into it again.

AlexandraBuzila:
@johnmorrell I just had a look at references and defaults. This is indeed not working and it seems to be a limitation of ajv: https://github.com/epoberezkin/ajv/issues/337#issuecomment-308864772

One way of working around this would be to add a default alongside your reference:

```javascript
{
   "type":"object",
   "properties":{
      "color":{
         "type":"string",
         "enum":[
            "red",
            "green",
            "blue"
         ],
         "default":"green"
      },
      "referencedColor":{
         "$ref":"#/definitions/referencedColor",
         "default":"blue"
      }
   },
   "definitions":{
      "referencedColor":{
         "type":"string",
         "enum":[
            "red",
            "green",
            "blue"
         ],
         "default":"green"
      }
   }
}
```

johnmorrell:
Thanks @AlexandraBuzila. I understand the issue now, and thanks for the workaround.

marshallmcdonnell:
[...]

PR #1224, opened 2019-01-14
Preselect with default schema values #1193

AlexandraBuzila:
- Adds options to `createAjv` method . Clients that want to have preselected
default values should pass in the `useDefaults: true` option to the
`createAjv` method and use the Ajv object in the `Actions.init` call.

### DIFF
--- a/packages/core/test/util/renderer.test.ts
+++ b/packages/core/test/util/renderer.test.ts
@@ -22,11 +22,12 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 */
+import { init, update, UPDATE_DATA, UpdateAction } from '../../src/actions';
 import test from 'ava';
-import * as _ from 'lodash';
 import * as Redux from 'redux';
 import {
   clearAllIds,
+  createAjv,
   createDefaultValue,
   mapDispatchToArrayControlProps,
   mapDispatchToControlProps,
@@ -36,7 +37,7 @@ import {
   OwnPropsOfControl
 } from '../../src/util';
 import configureStore from 'redux-mock-store';
-import { init, update, UPDATE_DATA, UpdateAction } from '../../src/actions';
+import * as _ from 'lodash';
 import { generateDefaultUISchema } from '../../src/generators';
 import {
   ControlElement,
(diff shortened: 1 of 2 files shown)


AlexandraBuzila:
I rebased the branch on master and added tests

packages/core/test/util/renderer.test.ts
```diff
@@ -588,3 +589,100 @@ test('mapStateToLayoutProps - hidden via state with path from ownProps ', t => {
...
+    }
+  };
+  const store: Store<JsonFormsState> = createStore(
+    combineReducers({ jsonforms: jsonformsReducer() }),
+    initState
+  );
+  store.dispatch(
+    init(data, schema, uischema, createAjv({ useDefaults: true }))
+  );
+  const ownProps: OwnPropsOfControl = { schema };
```
edgarmueller: I don't think we need `ownProps` and the `mapDispatchToArrayControlProps` call

packages/core/test/util/renderer.test.ts
```diff
@@ -588,3 +589,100 @@ test('mapStateToLayoutProps - hidden via state with path from ownProps ', t => {
...
+
+  mapDispatchToArrayControlProps(store.dispatch, ownProps);
+
+  t.is(store.getState().jsonforms.core.data.color, 'green');
+});
+
+test('should assign defaults to array', t => {
```
edgarmueller: A better name would be `"should assign defaults within nested object of an array"`

packages/core/test/util/renderer.test.ts
```diff
@@ -588,3 +589,100 @@ test('mapStateToLayoutProps - hidden via state with path from ownProps ', t => {
...
+  const ownProps: OwnPropsOfControl = {
+    schema,
+    uischema
+  };
+  const props = mapDispatchToArrayControlProps(store.dispatch, ownProps);
+
+  props.addItem('')();
+
+  t.is(store.getState().jsonforms.core.data.length, 2);
+  t.deepEqual(store.getState().jsonforms.core.data[0], { message: 'foo' });
```
edgarmueller: Split the expectation of having the default initialization kick-in for the 1st element into a separate test case. This test should only concern itself with checking the defaults of a newly added item.

packages/core/test/util/renderer.test.ts
AlexandraBuzila: Done, removed.

packages/core/test/util/renderer.test.ts
AlexandraBuzila: Done, changed the name.

packages/core/test/util/renderer.test.ts
AlexandraBuzila: Added a new test.

PR #1242, opened 2019-01-24
Change import style of 'ajv' and 'lodash' in all non test files

sdirix:


zummed:
I had to add:

    "allowSyntheticDefaultImports": true

to my tsconfig.json after trying out version "2.1.1-2.1.1-beta.0.0" just now, presumably because of this change.  Without it the follow error is generated:

    /data/work/analytics/tegu/node_modules/@jsonforms/core/lib/actions/index.d.ts
    (1,8): Module '"/data/work/analytics/tegu/node_modules/@jsonforms/core/node_modules/ajv/lib/ajv"' has no default export.

sdirix:
Hi @zummed. You are right, we'll update our README.MD for that. While `"allowSyntheticDefaultImports": true` allows you to compile, you might run into runtime issues (could you test that?). You might need to set `"esModuleInterop": true`.

edgarmueller:
Stefan is right, and I'm sorry about the broken version identifier. I have already released a new version with correct one, which is 2.1.1-beta.0.

zummed:
Ah thank you both :-) Yes no probs, I'll continue with the beta version so will keep an eye out for weirdness.

However I'm currently on react-redux 6.x, so aware some faults might be my doing.

zummed:
@edgarmueller That's good to know thanks.  I am using redux 4.x (4.0.1) so the same wariness applies :-)

Up until today I've had no issues (using react and the material renderer) so all good!  However, I started using an array in the schema today which is failing, hence my upgrade.  Not looked into the cause yet, but looks like it's my fault with the redux version from the initial error output.

PR #1247, opened 2019-01-28
Fix version ranges of own peer dependencies

edgarmueller:
Unify peer dependency version range within own scope.


eneufeld:
Looks good.
Do you think it might make sense to bound to a minor version in order to indicate that we might require specific changes from there and prepare the users for semantic versioning?

edgarmueller:
I'd certainly makes sense, we only need to make sure to update the number manually as lerna does not take care of updating peer dependencies.

edgarmueller:
Follow-up #1248 has been created.

Issue #1253, opened 2019-02-07
Invalid output from oneOf element/wrong field bindings


johnmorrell:
**Describe the bug**
Invalid JSON output from oneOf elements, where the output value is nested a level deeper than expected.

**To Reproduce**

Schema:
```
{
   "type":"object",
   "properties":{
      "vaule":{
         "oneOf":[
            {
              "title":"String",
              "type":"string"
            },
            {
              "title":"Number",
              "type":"number"
            }
         ]
      }
   }
}
```

UISchema:
```
{
   "type":"Control",
   "label":"Value",
   "scope":"#/properties/vaule"
}
```

Enter string and inspect data as JSON, to get:
```
{
  "vaule": {
    "vaule": [
      "abcdefg"
    ]
  }
}
```

**Expected behavior**

Example output as JSON:
```
{
    "vaule": "abcdefg"
}
```

**Browser (please complete the following information):**
 - Browser: Safari
 - Version: 12.0.3

**Used Setup (please complete the following information):**
 - Framework: React
 - RendererSet: Material

**Additional context**
JSON Forms v2.2.0

edgarmueller:
Thanks for the report! We experienced this issue ourselves in one of our projects and started working on a fix. I hope we'll be able to release a fix in the upcoming days, probably in the beginning of the next week.

edgarmueller:
@johnmorrell We've released a new 2.2.1-alpha.0 release which should address this issue. Let me know if it works.

johnmorrell:
Thank @edgarmueller, I've take a look at 2.2.1-alpha.0 and the data structure now looks good. The titles are big improvement too, but I did notice a case where the first title would be shown as ONEOF-0 rather than the title value. Simplified examples below:

**This works as expected**
```
{
   "oneOf":[
      {
         "title":"Option 1",
         "type":"object",
         "properties":{
            "option1":{
               "oneOf":[
                  {
                     "title":"Option 1A",
                     "type":"string"
                  },
                  {
                     "title":"Option 1B",
                     "type":"string"
                  }
               ]
            }
         }
      },
      {
         "title":"Option 2",
         "type":"object",
         "properties":{
            "option2":{
               "type":"number"
            }
         }
      }
   ]
}
```

**This results in first title being ONEOF-0**
```
{
   "oneOf":[
      {
         "title":"Option 1",
         "$ref":"#/definitions/moreOptions"
      },
      {
         "title":"Option 2",
         "type":"object",
[...]

PR #1263, opened 2019-02-15
Fix field binding

edgarmueller:
Fields do not alter their own path anymore, hence we need to pass it in accordingly. This should fix the most recent example in #1253

### DIFF
--- a/packages/material/src/complex/MaterialTableControl.tsx
+++ b/packages/material/src/complex/MaterialTableControl.tsx
@@ -141,9 +141,10 @@ const NonEmptyCell = ({
   scopedSchema,
   errors
 }: NonEmptyCellProps) => {
+  const path = rowPath + (scopedSchema.type === 'object' ? '.' + propName : '');
   const errorsPerEntry: any[] = filter(
     errors,
-    error => error.dataPath === rowPath + '.' + propName
+    error => error.dataPath === path
   ).map(e => e.message);
   const isValid = isEmpty(errorsPerEntry);
   return (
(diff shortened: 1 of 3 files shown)


packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -155,7 +155,7 @@ const NonEmptyCell = ({
             undefined,
             scopedSchema.type === 'object' ? `#/properties/${propName}` : '#'
           )}
-          path={rowPath}
+          path={rowPath + (scopedSchema.type === 'object' ? '.' + propName : '')}
```
eneufeld: good catch, we should probably fix line 146 accordingly

packages/vanilla/src/complex/TableArrayControl.tsx
```diff
@@ -162,7 +162,7 @@ class TableArrayControl extends React.Component<
                               <DispatchField
                                 schema={schema}
                                 uischema={createControlElement(prop)}
-                                path={childPath}
+                                path={childPath + '.' + prop}
```
eneufeld: shouldn't we do the same check here as in the material case? As far as I can see, this table is also used for primitive arrays

edgarmueller: We are inside a ternary expression here, see line 148

packages/material/src/complex/MaterialTableControl.tsx
edgarmueller: Fixed.

PR #1266, opened 2019-02-18
[material] Add oneOf tests

edgarmueller:
Adds initial set of tests for oneOf renderer, fix default value initialization

### DIFF
--- a/packages/material/src/complex/MaterialOneOfRenderer.tsx
+++ b/packages/material/src/complex/MaterialOneOfRenderer.tsx
@@ -60,7 +60,7 @@ class MaterialOneOfRenderer extends React.Component<ControlProps & OneOfProps, M
     const { path, schema, handleChange } = this.props;
     handleChange(
       path,
-      createDefaultValue(schema.oneOf[this.state.selectedOneOf])
+      createDefaultValue(schema.oneOf[this.state.newOneOfIndex])
     );
     this.setState({
       open: false,
@@ -69,7 +69,7 @@ class MaterialOneOfRenderer extends React.Component<ControlProps & OneOfProps, M
     });
   };
 
-  handleChange = (_event: any, newOneOfIndex: number) => {
+  handleTabChange = (_event: any, newOneOfIndex: number) => {
     this.setState({
       open: true,
       newOneOfIndex
@@ -79,7 +79,7 @@ class MaterialOneOfRenderer extends React.Component<ControlProps & OneOfProps, M
   render() {
 
     const oneOf = 'oneOf';
-    const { schema, path, rootSchema } = this.props;
+    const { schema, path, rootSchema, id } = this.props;
     const _schema = resolveSubSchemas(schema, rootSchema, oneOf);
     const oneOfRenderInfos = createCombinatorRenderInfos((_schema as JsonSchema).oneOf, rootSchema, oneOf);
 
@@ -90,7 +90,7 @@ class MaterialOneOfRenderer extends React.Component<ControlProps & OneOfProps, M
           combinatorKeyword={'oneOf'}
           path={path}
         />
-        <Tabs value={this.state.selectedOneOf} onChange={this.handleChange}>
+        <Tabs value={this.state.selectedOneOf} onChange={this.handleTabChange}>
           {oneOfRenderInfos.map(oneOfRenderInfo => <Tab key={oneOfRenderInfo.label} label={oneOfRenderInfo.label}/>)}
         </Tabs>
         {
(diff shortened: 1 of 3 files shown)


packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
```diff
@@ -0,0 +1,404 @@
...
+      wrapper.update();
+
+      // click add button
+      buttons = wrapper.find('button');
+      const addButton = buttons.at(2);
+      const nrOfRowsBeforeAdd = wrapper.find('tr');
+      addButton.simulate('click');
+      addButton.simulate('click');
+
+      // check if row has been added
```
eneufeld: this should say 'check if row**s** have been added as you added 2

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
```diff
@@ -0,0 +1,404 @@
...
+      const nrOfRowsAfterAdd = wrapper.find('tr');
+
+      // 1 header row + no data row
+      expect(nrOfRowsBeforeAdd.length).toBe(2);
+      // 1 header row + 2 data rows (one is replacing the 'No data' one)
+      expect(nrOfRowsAfterAdd.length).toBe(3);
+    });
+  });
+
+  it('should add an object within an array', () => {
```
eneufeld: isn't this exactly the same as the previous except one more assert?

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
```diff
@@ -0,0 +1,404 @@
...
+  return store;
+};
+
+describe('Material oneOf renderer', () => {
+
+  let wrapper: ReactWrapper;
+
+  afterEach(() => wrapper.unmount());
+
+  it('should add an item at correct path', () => {
```
eneufeld: can't we unify this with the `should switch to array based oneOf subschema, then switch back, then edit` test`? imho  the second one is a better test of the first one especially if you remove the intermediate objects which is tested in the array tests

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
edgarmueller: Yes, it's only the schema that differs.

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
edgarmueller: Depends on the view point: initially we already have two rows (header + no data), if we add two items, one is replacing the no data one, hence there's one row more than initially. But I can update if you want, if 'rows' seems to make more sense to you.

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
edgarmueller: Yes, we could do that, the schema is different though. Whatever makes more sense to you.

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
eneufeld: I'm fine with both, my thinking was that you had no data, and add two entries, but I see what you meant.

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
eneufeld: I now see the different schemas, but I just tried out using one schema for both tests and they still work. The array part which is the only one being checked is the same for both schemas.

packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
eneufeld: I now see the different schemas, but I just tried out using one schema for both tests and they still work. The array part which is the only one being checked is the same for both schemas.

edgarmueller:
@eneufeld I've extracted some utility methods to reduce boilerplate, but kept the tests as is otherwise.

packages/material/src/complex/MaterialOneOfRenderer.tsx
```diff
@@ -69,7 +69,8 @@ class MaterialOneOfRenderer extends React.Component<ControlProps & OneOfProps, M
     });
   };
 
-  handleChange = (_event: any, newOneOfIndex: number) => {
+  // TODO: rename
```
eneufeld: the todo is obsolete now

edgarmueller: Thanks, fixed

PR #1276, opened 2019-03-04
Make schemaMatches more stable

eneufeld:
* moved deriveType to util
* schemaMatches doesn't expect a type property but uses deriveType
* added test case for improved schemaMatches behavior


edgarmueller:
Looks good, thanks

edgarmueller:
BTW, we have more of those `schema.type === 'object'` checks in the codebase. It would probably make sense to remove them as well.

Issue #1275, opened 2019-02-26
Props should not be undefined when consuming via TS


edgarmueller:
**Describe the bug**
When accessing the `schema` prop in a TS environment with strict null checks enabled, the TS compiler complains about possible undefined access. This is annoying as one has to perform undefined checks everywhere. This is because we defined the `schema` property to be optional, I'm note sure why though.

**To Reproduce**
Consume core package in TS environment with strict null-checks enabled. Access `schema` prop.

**Expected behavior**
No TS compiler errors.

 - Version 2.2.1-alpha.4

**Used Setup (please complete the following information):**
 - Framework: core

edgarmueller:
Fixed via #1281

PR #1281, opened 2019-03-05
Improve typings

edgarmueller:
Addresses #1275 by improving typings:

* Split own & stated-based props (no optionals anymore)
* Change type of 1st parameter in `isVisible`/`isEnabled`
* Provide default renderer props
* Remove parentPath prop
* Add type annotation to `errorAt`
* Replace `resolvedRootData` with `data` in mui-tree
* Differentiate between array control and regular control in angular package

### DIFF
--- a/packages/core/src/util/index.ts
+++ b/packages/core/src/util/index.ts
@@ -25,12 +25,7 @@
 import isEmpty from 'lodash/isEmpty';
 import isArray from 'lodash/isArray';
 import head from 'lodash/head';
-import {
-  JsonFormsState,
-  JsonSchema,
-  Scopable,
-  StatePropsOfRenderer
-} from '../';
+import { JsonSchema, Scopable, UISchemaElement } from '../';
 import { resolveData, resolveSchema } from './resolvers';
 import {
   compose as composePaths,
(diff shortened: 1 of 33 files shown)


packages/core/src/util/index.ts
```diff
@@ -127,11 +122,11 @@ export { composePaths, composeWithUi, Paths, toDataPath };
 
 // Runtime --
 const Runtime = {
-  isEnabled(props: StatePropsOfRenderer, state: JsonFormsState): boolean {
-    return isEnabled(props, state);
+  isEnabled(uischema: UISchemaElement, state: JsonFormsState): boolean {
```
eneufeld: this is an api change, should we not keep backward compatibility?

packages/core/src/util/index.ts
```diff
@@ -127,11 +122,11 @@ export { composePaths, composeWithUi, Paths, toDataPath };
...
+    return isEnabled(uischema, state);
   },
-  isVisible(props: StatePropsOfRenderer, state: JsonFormsState): boolean {
-    return isVisible(props, state);
+  isVisible(uischema: UISchemaElement, state: JsonFormsState): boolean {
```
eneufeld: same as above

packages/vanilla/src/complex/LabelRenderer.tsx
```diff
@@ -1,19 +1,19 @@
 /*
   The MIT License
-  
+
   Copyright (c) 2018 EclipseSource Munich
```
eneufeld: year is old

packages/core/src/util/index.ts
edgarmueller: As discussed, I wouldn't, because I think the signature as it is right now does not make much sense. I'd rather just document it in our release section. I've also updated the signature as passing the whole state as the second parameter is weird as well.

PR #1284, opened 2019-03-06
[material] Adapt categorization renderer to support visibility rules for categories

### LINKED ISSUE #1278: Categories should support rules
**Describe the bug**
Categories should be hideable based on rules.

**To Reproduce**
1. Specify a rule on any category, e.g. in the context of the person example:
```js
{
      type: 'Category',
      label: 'Additional',
      elements: [
        {
          type: 'Control',
          scope: '#/properties/birthDate'
        },
        {
          type: 'Control',
          scope: '#/properties/vegetarian'
        }
      ],
      rule: {
        effect: 'HIDE',
        condition: {
          scope: '#/properties/vegetarian',
          schema: { const: true }
        }
      }
    }
```

**Expected behavior**
When setting the `vegetarian` property to true, the "Additional" tab should be hidden

 - Framework: react, angular]
 - RendererSet: material

edgarmueller:
Support visibility rules on categories, fixes #1278

### DIFF
--- a/packages/examples/src/categorization.ts
+++ b/packages/examples/src/categorization.ts
@@ -41,6 +41,10 @@ export const uischema = {
         {
           type: 'Control',
           scope: '#/properties/birthDate'
+        },
+        {
+          type: 'Control',
+          scope: '#/properties/vegetarian'
         }
       ]
     },
@@ -56,7 +60,14 @@ export const uischema = {
           type: 'Control',
           scope: '#/properties/vegetarian'
         }
-      ]
+      ],
+      rule: {
+        effect: 'HIDE',
+        condition: {
+          scope: '#/properties/vegetarian',
+          schema: { const: true }
+        }
+      }
     }
   ]
 };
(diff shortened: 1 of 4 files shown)


cliffschwab:
hello, we ran into the same issue as well. just want to get some idea  when the fix will be released. We currently see this issue in release 2.2.2

edgarmueller:
@cliffschwab What's the issue exactly? Could you post an example, so that we can reproduce it?

cliffschwab:
@edgarmueller please look at the ui config info example below, we want to control one tab's visibility based on some controlling fields we collected in previously. Currently the control will respond to the rules like below, but not the tab/Category. 

{
      "type": "Category",
      "label": "??? Plan",
      "rule": {
        "effect": "HIDE",
        "condition": {         
         "scope": "#/properties/controllerType",
          "expectedValue": "valueA"
        }
      },
      "elements": [
      {......}]

edgarmueller:
@cliffschwab Works for me. I've pushed the example I tested with to https://github.com/edgarmueller/jsonforms-react-seed/tree/examples/hide-category. So I guess, it must be something else that causes the misbehaviour? Could you post a reproducer?

cliffschwab:
@edgarmueller You r right,  we pull the 2.2.2 again and test the sample code and it is working now, thx for the follow up. great work!

PR #1287, opened 2019-03-07
Restrict minor versions of peer deps

edgarmueller:
I'll update the Wiki to take care of updating the minor versions of peer dependencies once this has been merged.


edgarmueller:
I think you mean 2.3 (such that the current 2.3 milestones actually becomes 2.4) and the version of the peer deps should not be bigger than the current one, I guess? Updating them will be part of the release cycle.

PR #1288, opened 2019-03-07
Remove defaultProps from signature, move defaults to mapStateToProps

edgarmueller:
Doc generation via `npm run doc` failed apparently, this change addresses this issue.

### DIFF
--- a/packages/vanilla/src/layouts/GroupLayout.tsx
+++ b/packages/vanilla/src/layouts/GroupLayout.tsx
@@ -1,19 +1,19 @@
 /*
   The MIT License
-  
+
   Copyright (c) 2018 EclipseSource Munich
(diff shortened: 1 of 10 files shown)


packages/vanilla/src/layouts/GroupLayout.tsx
```diff
@@ -1,19 +1,19 @@
 /*
   The MIT License
-  
+
   Copyright (c) 2018 EclipseSource Munich
```
eneufeld: year

edgarmueller:
I agree, actually we even have a script for that but I think it's not capable of updating headers, only adding them, I've created a follow-up https://github.com/eclipsesource/jsonforms/issues/1289

PR #1293, opened 2019-03-11
errorAt and subErrorAt now expect a schema

### LINKED ISSUE #1265: Duplicate validation warnings within oneOf
**Describe the bug**
Duplicate validation warnings within oneOf block. Warnings seem to be accumulated, so the number of duplicates can go up the more oneOf items present.

screenshot 2019-02-18 at 14 09 04

**To Reproduce**
Use the following schema and add item to array, observing the waring count go from 1 to 4. Selecting the error count shows a 3x error message duplication:
```
{
   "type":"object",
   "properties":{
      "coloursOrNumbers":{
         "oneOf":[
            {
               "$ref":"#/definitions/colours"
            },
            {
               "$ref":"#/definitions/numbers"
            },
            {
               "$ref":"#/definitions/shapes"
            }
         ]
      }
   },
   "definitions":{
      "colours":{
         "title":"Colours",
         "type":"array",
         "items":{
            "title":"Type",
            "type":"string",
            "enum":[
               "Red",
               "Green",
               "Blue"
            ],
            "minItems":1
         }
      },
      "numbers":{
         "title":"Numbers",
         "type":"array",
         "items":{
            "title":"Type",
            "type":"string",
            "enum":[
               "One",
               "Two",
               "Three"
            ],
            "minItems":1
         }
      },
      "shapes":{
         "title":"Shapes",
         "type":"array",
         "items":{
            "title":"Type",
            "type":"string",
            "enum":[
               "Circle",
               "Triangle",
               "Square"
            ],
            "minItems":1
         }
      }
   }
}
```

**Expected behavior**
No duplication of validation errors.

Safari v12.0.3

React/Material

**Additional context**
^2.2.0

eneufeld:
* AJV is now configured with `verbose` flag
* errorAt checks the parentSchema of error against provided schema
* subErrorAt checks the parentSchema of error against provided schema
* added test cases for both methods

Fixes #1265

### DIFF
--- a/packages/core/test/reducers/core.test.ts
+++ b/packages/core/test/reducers/core.test.ts
@@ -26,8 +26,14 @@ import test from 'ava';
 import { coreReducer } from '../../src/reducers';
 import { init } from '../../src/actions';
 import { JsonSchema } from '../../src/models/jsonSchema';
-import { errorAt, JsonFormsCore, subErrorsAt } from '../../src/reducers/core';
-import cloneDeep = require('lodash/cloneDeep');
+import {
+  errorAt,
+  JsonFormsCore,
+  sanitizeErrors,
+  subErrorsAt
+} from '../../src/reducers/core';
+
+import { createAjv } from '../../src';
 
 test('core reducer should support v7', t => {
   const schema: JsonSchema = {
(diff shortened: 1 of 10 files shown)


packages/core/test/reducers/core.test.ts
```diff
@@ -48,3 +50,199 @@ test('core reducer should support v7', t => {
...
+        schemaPath: '',
+        message: '',
+        params: {},
+        schema: undefined,
+        parentSchema: cloneDeep(schema.properties.foo)
+      },
+    ]
+  };
+  const filtered = subErrorsAt('foo', schema.properties.foo)(state);
+  t.is(filtered.length, 1);
```
edgarmueller: Maybe we should also add or reformulate a test where more than one item is matched as `subErrorsAt` is actually supposed to do that.

edgarmueller:
@eneufeld  Would it work if we conditionally check the `parentSchema` for combinator keywords? Something like in 
```js
if (error.dataPath === instancePath) {
  if (has(error.parentSchema, 'oneOf')) {
    return error.dataPath === instancePath && isEqual(error.parentSchema, schema);
  }
}
return error.dataPath === instancePath;
```

eneufeld:
I'll check this, by I think the oneOf was not part of the parentSchema

PR #1297, opened 2019-03-15
Topic/performance

eneufeld:


packages/examples/src/huge.ts
```diff
@@ -0,0 +1,916 @@
...
+          type: 'array',
+          items: { $ref: '#/definitions/identifier' }
+        },
+        created: { type: 'integer' },
+        modified: { type: 'integer' },
+        repository: { $ref: '#/definitions/resourceReference' }
+      },
+      required: ['citations']
+    },
+    agent: {
```
edgarmueller: This is example shows a couple of validation errors in the example for the agents property, but the error text is empty. Any idea why?

packages/examples/src/huge.ts
```diff
@@ -0,0 +1,916 @@
+import { registerExamples } from './register';
```
edgarmueller: This example breaks the layout of the example page. I think this is the case because of very long strings. Can we wrap them, perhaps?

packages/material/src/complex/MaterialAnyOfRenderer.tsx
```diff
@@ -18,7 +18,7 @@ interface MaterialAnyOfState {
   selectedAnyOf: number;
 }
 
-class MaterialAnyOfRenderer extends React.Component<ControlProps, MaterialAnyOfState> {
+class MaterialAnyOfRenderer extends React.Component<MaterialAllOfRendererProps, MaterialAnyOfState> {
```
edgarmueller: See previous comment about naming

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -38,21 +39,22 @@ import {
...
   Paths,
   Resolve
 } from '@jsonforms/core';
 import IconButton from '@material-ui/core/IconButton';
 import DeleteIcon from '@material-ui/icons/Delete';
 import { WithDeleteDialogSupport } from './DeleteDialog';
-import { ErrorObject } from 'ajv';
 import NoBorderTableCell from './NoBorderTableCell';
 import TableToolbar from './TableToolbar';
+import { connect } from 'react-redux';
```
edgarmueller: Move import above relative ones.

packages/material/src/controls/MaterialDateControl.tsx
```diff
@@ -140,21 +142,17 @@ export const addLabelProps =
...
             return {
                 ...stateProps,
-                labels: {
-                    // TODO cast
-                    default: stateProps.label as string,
-                    cancel: '%cancel',
-                    clear: '%clear'
-                },
+                // TODO cast
+                default_label: stateProps.label as string,
```
edgarmueller: Why use snake_case here?

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
-     findUISchema,
-     removeItems,
-     renderers
-   }: ArrayControlProps) => {
+const paperStyle = { padding: 10 };
+const iconStyle: any = { float: 'right' };
+export class MaterialArrayLayout extends React.Component<
+  ArrayLayoutProps,
+  { expanded: string | boolean }
+> {
```
edgarmueller: While this formatted correctly this is hard to read. I'd suggest to add some spacing, at least beween the class signature and its body.

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
               <Grid item>
-                <ValidationIcon id='tooltip-validation' errorMessages={errors}/>
+                <Tooltip
+                  id='tooltip-add'
+                  title={`Add to ${label}`}
+                  placement='bottom'
+                >
+                  <IconButton
+                    aria-label={`Add to ${label}`}
+                    onClick={addItem(path, createDefault())}
```
edgarmueller: The add button doesn't seem to work anymore (nothing happens)

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
+    this.setState({
+      expanded: expanded ? panel : false
+    });
+  };
+  isExpanded = (index: number) => this.state.expanded === composePaths(this.props.path, `${index}`);
+  render() {
+    const { data, path, schema, uischema, errors, addItem, renderers } = this.props;
+    const labelDescription = Helpers.createLabelDescriptionFrom(uischema);
+    const label = labelDescription.show ? labelDescription.text : '';
+    const arrayLayoutToolbarProps: ArrayLayoutToolbarProps = {
```
edgarmueller: Why not inline the props directly?

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
-              <Grid container>
+            </Grid>
+          </Grid>
+        </Grid>
+      </Toolbar>
+    );
+  }
+);
+
+class ExpandPandelRenderer extends React.Component<ExpandPanelProps, any> {
```
edgarmueller: Typo

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
+        return array;
+      })
+    );
+  }
+});
+export interface ExpandPanelProps
+  extends StatePropsOfExpandPanel,
+    DispatchPropsOfExpandPanel {}
+
+const ConnectedExpandPandelRenderer = connect(
```
edgarmueller: Typo

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
-     addItem,
-     findUISchema,
-     removeItems,
-     renderers
-   }: ArrayControlProps) => {
+const paperStyle = { padding: 10 };
+const iconStyle: any = { float: 'right' };
+export class MaterialArrayLayout extends React.Component<
+  ArrayLayoutProps,
+  { expanded: string | boolean }
```
edgarmueller: You can have multiple expanded panels, but according to the type signature it seems as if we can have only one.

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
+      </ExpansionPanel>
+    );
+  }
+}
 
-              const foundUISchema =
-                findUISchema(schema, uischema.scope, path, undefined, uischema);
-              const childPath = composePaths(path, `${index}`);
-              const childLabel = firstPrimitiveProp ? childData[firstPrimitiveProp] : '';
+interface OwnPropsOfExpendPanel {
```
edgarmueller: Typo

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -40,114 +56,266 @@ import Avatar from '@material-ui/core/Avatar';
...
-        prop.type === 'number' ||
-        prop.type === 'integer';
-    }) : undefined;
+    return (
+      <Paper style={paperStyle}>
+        <ArrayLayoutToolbar {...arrayLayoutToolbarProps} />
+        <div>
+          {data > 0 ?
+            map(range(data), index => {
+              const expansionPanelProps: OwnPropsOfExpendPanel = {
```
edgarmueller: Inline props

packages/material/src/layouts/MaterialGroupLayout.tsx
```diff
@@ -39,33 +39,40 @@ import { MaterialLayoutRenderer, MaterialLayoutRendererProps } from '../util/lay
...
     const groupLayout = uischema as GroupLayout;
-
     const childProps: MaterialLayoutRendererProps = {
-      elements: groupLayout.elements,
-      schema,
-      path,
-      direction: 'column',
-      visible,
-      renderers
+        elements: groupLayout.elements,
```
edgarmueller: Inline props

packages/core/src/reducers/core.ts
```diff
@@ -253,12 +253,17 @@ export const subErrorsAt = (instancePath: string, schema: JsonSchema) => (
   state: JsonFormsCore
 ): ErrorObject[] => {
   const path = `${instancePath}.`;
-
-  return filter(
+  const oneOfDataPaths = filter(
```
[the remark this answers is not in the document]
eneufeld: I suggest to create a separate issue for anyOf .
About extracting the filter, the tow methods are not the same. The datatpath for the result is checked differently.

packages/core/src/util/renderer.ts
```diff
@@ -537,9 +547,7 @@ export const mapStateToLayoutProps = (
   };
 };
 
-export interface RefResolver {
-  (schema: JsonSchema): Promise<any>;
-}
+// export type RefResolver = (schema: JsonSchema) => Promise<JsonSchema>;
```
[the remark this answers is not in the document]
eneufeld: it didn't compile for me before. I forgot to comment it in again.

packages/core/src/util/renderer.ts
```diff
@@ -589,3 +596,86 @@ export const controlDefaultProps = {
   ...layoutDefaultProps,
   errors: [] as string[]
 };
+
+export interface MaterialAllOfRendererProps {
```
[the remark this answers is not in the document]
eneufeld: I fixed the naming. 
I renamed it to be CombinatorRendererProps as it is used for AllOfRenderer and AnyOfRenderer. I didn't use it in the OneOfRenderer as it needs also the id. 
I would suggest to check the OneOfRenderer and an improved hierarchy after this is merged

packages/material/src/complex/MaterialAnyOfRenderer.tsx
eneufeld: fixed with the rename

packages/material/src/layouts/MaterialArrayLayout.tsx
eneufeld: now you can't ;-) i suggest to open a new issue for this if we want to support multiple open panels at once

packages/core/src/util/renderer.ts
```diff
@@ -589,3 +596,86 @@ export const controlDefaultProps = {
...
+  return {
+    path,
+    schema: resolvedSchema || rootSchema,
+    rootSchema,
+    visible
+  };
+};
+
+export interface StatePropsOfArrayLayout extends StatePropsOfControlWithDetail {
+  data: number;
```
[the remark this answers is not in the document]
eneufeld: I don't get your comment, could you explain in more detail?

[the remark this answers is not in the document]
eneufeld: Ok, so the reason for using a number here, is that the data property is mandatory. In order to have properties which can be compared easily, we need to have something that doesn't change if a value further down changes.

edgarmueller:
@eneufeld Could you rebase once more?

PR #1298, opened 2019-03-18
[material] Layout passes renderers props down to children

edgarmueller:


edgarmueller:
I've created a follow-up issue which is https://github.com/eclipsesource/jsonforms/issues/1299

Issue #1210, opened 2019-01-05
Support Schema reference to other Schema files


johnmorrell:
With complex Schemas it can be useful to define in multiple files and reference across files, eg:

```
"$ref":"definitions.json#/quantity"
```

Currently (v2.1.0-rc.0) such references seem to fail to resolve. It would be very useful if this was supported.

edgarmueller:
Yes, I agree. I think part of this issue is that `resolveSchema` does not correctly consider cases where the `schemaPath` points to a ref again. We'll have a look soon and will try to come up with a fix.

sdirix:
I took a closer look at this issue. JSONForms only resolves references within schemas by design.

JSONForms still supports schemas split over multiple files, however you have to take care of the resolving yourself. For this you need to create your own `Ajv` instance and execute `addSchema` on it with your schema dependencies. Afterwards you pass the `Ajv` instance to the the `init` action of JSONForms.

Example:
```typescript
import {
  Actions,
  createAjv
} from '@jsonforms/core';

// ...

const ajv = createAjv();
ajv.addSchema(
  definitionsSchema, // the schema you got from your disk, REST call, etc.
  'definitions.json' // the id which will be used to refer to this schema
);
store.dispatch(
  Actions.init(
    data,     // your data
    schema,   // your schema (which might reference 'definitions.json')
    uischema, // your uischema
    ajv
  )
);

```

This should already work in your application. You can also check #1241 where we adapted our resolve examples.

Please let us know if this answers your question or if you had another use case in mind.

johnmorrell:
Thanks.

I have tried this and it looks like the ajv is setup correctly with the primary schema referencing the definitions schema, but the app fails to launch with error:

`TypeError: undefined is not an object (evaluating 'schema.properties')`


index.js
```
import React from 'react';
import ReactDOM from 'react-dom';
import './index.css';
import App from './App';
import registerServiceWorker from './registerServiceWorker';
import { combineReducers, createStore } from 'redux';
import { Provider } from 'react-redux';
import { Actions, createAjv, jsonformsReducer } from '@jsonforms/core';
import { materialFields, materialRenderers } from '@jsonforms/material-renderers';
import schema from './schema/schema.json';
import definitionsSchema from './schema/definitions.json';
import uischema from './uischema.json';

const data = {};

const ajv = createAjv();
ajv.addSchema(definitionsSchema, 'definitions.json');

const store = createStore(
[...]

PR #1302, opened 2019-03-18
Use JSON Schema $Ref Parser instead of JsonRefs

sdirix:
JsonRefs isn't customizable enough to support all use cases concerning
references resolving. This commit removes JsonRefs and adds
JSON Schema $Ref Parser to the core library.

JSON Schema $Ref Parser configuration is exposed via the 'init' action
similar to Ajv.

 * Remove JsonRefs from all packages
 * Use JSON Schema $Ref Parser for resolving (react / material & webcomponent)
 * Add new options object to 'init' action
 * Add options fallback to not break existing applications
 * Copy 'findRefs' method of JsonRefs as JSON Schema $Ref Parser does
not provide a similar method
 * Adapt existing testcases using resolving
 * Remove resources and resource-set from material-tree-renderer as they
were unused

Implements #1210


edgarmueller:
LGTM, thanks

PR #1306, opened 2019-03-21
Register the radio control and use RankedTester to detect format: radio

iggh:
The radio controller was developed but wasn’t exported with a result I could use.

Export materialRadioGroupControlTester with rankedTester to be called by:

```
options {
    Format: "radio"
}
```

Register MaterialRadioGroupControl and  MaterialRadioGroupControlTester on `packages/material/src/controls/index.ts` and `packages/material/src/index.ts` in order to be globally available

Please let me know if any specific tests need to be added and/or point me in the right direction if this doesn't pass the test criteria.

### DIFF
--- a/packages/material/src/controls/MaterialRadioGroupControl.tsx
+++ b/packages/material/src/controls/MaterialRadioGroupControl.tsx
@@ -31,6 +31,9 @@ import {
   formatErrorMessage,
   isDescriptionHidden,
   isPlainLabel,
+  rankWith,
+  RankedTester,
+  optionIs,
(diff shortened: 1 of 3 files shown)


edgarmueller:
Thanks for the PR! Once the formatter issues are solved, we can merge this (run `npm run check-format` at the root and it'll tell which files are violating formatting guideliens).

iggh:
Thanks for the pointer. Hopefully this now covers it.

edgarmueller:
Yep, looks good. Thanks!

PR #1312, opened 2019-03-30
Add support for allOf to deriveType

eneufeld:
* add test cases for deriveType
* add support to handle all of case

### DIFF
--- a/packages/core/test/util/derivetype.test.ts
+++ b/packages/core/test/util/derivetype.test.ts
@@ -65,13 +65,6 @@ test('derive type with type - union', t => {
   const schema: JsonSchema = {
     type: ['string', 'number']
   };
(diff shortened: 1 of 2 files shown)


packages/core/test/util/derivetype.test.ts
```diff
@@ -0,0 +1,101 @@
...
+});
+
+test('derive type with type - union', t => {
+  const schema: JsonSchema = {
+    type: ['string', 'number']
+  };
+  t.is(deriveType(schema), 'string');
+});
+
+test('derive type with type - union', t => {
```
edgarmueller: What the difference to the previous test?

eneufeld: that is a wrong copy, thank you for finding this

PR #1326, opened 2019-04-10
Handle anyOf error matching

edgarmueller:
I've only added a single test case, I don't know whether it makes sense to duplicate all of them, but I could do it, if requested.


packages/core/test/util/renderer.test.ts
```diff
@@ -349,7 +349,7 @@ test('mapStateToControlProps - no duplicate error messages', t => {
     { uischema: coreUISchema }
   );
   // 'should be string' should only appear once
```
eneufeld: do we need this comment?

PR #1335, opened 2019-04-18
Use schema title attribute where appropriate

sdirix:
The label utility will now return the schema title if no other label was
defined via the ui schema.

Within the ui schema a label can be optionally defined for each control.
Previously if there was none, we fell back to auto-generating a label
from the given control scope. Now we will check the schema title
beforehand and return that if no label was defined for the control.

This also affects the ui schema generation which will no longer
auto-generate labels based on each control's scope.

Includes testcases.


packages/core/src/util/label.ts
```diff
@@ -25,9 +25,16 @@
...
-const deriveLabel = (controlElement: ControlElement): string => {
-  if (controlElement.scope !== undefined) {
+const deriveLabel = (
+  controlElement: ControlElement,
+  schemaElement?: JsonSchema
+): string => {
+  if (schemaElement && typeof schemaElement.title === 'string') {
+    return schemaElement.title;
+  }
+  if (typeof controlElement.scope === 'string') {
```
eneufeld: why do we need this check?

sdirix: I did it because the check was already in the previous code as `controlElement.scope !== undefined`. I can remove it if you want.

eneufeld: No, its fine. I missed the check that was there.

sdirix:
Rebased on `master` for a working Travis build

PR #1345, opened 2019-04-23
Fix list with detail selection when editing label

edgarmueller:
When updating the property that used for displaying the label of a selected item, the selection resets. This PR fixes this issue and also adds a logger the angular example app.


packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -209,7 +209,9 @@ export class MasterListComponent extends JsonFormsArrayControl {
     if (
       newSelectedItem !== undefined &&
       this.selectedItem !== undefined &&
-      newSelectedItem.label === this.selectedItem.label
+      (newSelectedItem.label === this.selectedItem.label ||
+        (newSelectedItem.label !== this.selectedItem.label &&
+          newSelectedItem.path === this.selectedItem.path))
```
eneufeld: why not only check for the path?

PR #1359, opened 2019-05-02
clone oldData when reducing core.data to avoid state mutation in updater

jeremy-vineti:
I've run across a scenario when editing an array of objects, the updater function is mutating a piece of the state object and causing undesired side-effects. Adding a cloneDeep here solves the problem.


edgarmueller:
Yep, that makes sense, we even clone the data when modifying the root, but not when modifying nested data. Thanks for pointing this out!

PR #1365, opened 2019-05-10
[Angular] Fix l10n support for number renderer

eneufeld:


packages/angular-material/src/controls/number.renderer.ts
```diff
@@ -97,7 +104,15 @@ export class NumberControlRenderer extends JsonFormsControl {
...
       this.max = this.scopedSchema.maximum;
       this.multipleOf = this.scopedSchema.multipleOf || defaultStep;
-      this.locale = getLocale(this.ngRedux.getState());
+      const currentLocale = getLocale(this.ngRedux.getState());
+      if (this.locale === undefined || this.locale !== currentLocale) {
+        this.locale = currentLocale;
+        this.numberFormat = new Intl.NumberFormat(this.locale, {
+          maximumFractionDigits: 20
+        });
+        const example = this.numberFormat.format(1.1);
```
edgarmueller: A bit confusing at the 1st glance. I'd suggest to extract the determination of the decimal separator into its own function and name it accordingly `determineDecimalSeparator` or similar.

packages/angular-material/src/controls/number.renderer.ts
```diff
@@ -97,7 +104,15 @@ export class NumberControlRenderer extends JsonFormsControl {
       this.min = this.scopedSchema.minimum;
       this.max = this.scopedSchema.maximum;
       this.multipleOf = this.scopedSchema.multipleOf || defaultStep;
-      this.locale = getLocale(this.ngRedux.getState());
+      const currentLocale = getLocale(this.ngRedux.getState());
+      if (this.locale === undefined || this.locale !== currentLocale) {
+        this.locale = currentLocale;
+        this.numberFormat = new Intl.NumberFormat(this.locale, {
+          maximumFractionDigits: 20
```
edgarmueller: Extract to constant

PR #1366, opened 2019-05-10
[Angular] Fix path passing

eneufeld:
* remember path as propsPath
* use propsPath when passing

### DIFF
--- a/packages/angular-material/test/object-control.spec.ts
+++ b/packages/angular-material/test/object-control.spec.ts
@@ -22,17 +22,18 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 */
-import { BrowserDynamicTestingModule } from '@angular/platform-browser-dynamic/testing';
 import { NgRedux, NgReduxModule } from '@angular-redux/store';
+import { MockNgRedux } from '@angular-redux/store/lib/testing';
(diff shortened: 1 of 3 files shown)


edgarmueller:
@eneufeld Build is failing

packages/angular-material/test/object-control.spec.ts
```diff
@@ -32,11 +32,13 @@ import {
   MatFormFieldModule,
   MatInputModule
 } from '@angular/material';
-import { BrowserDynamicTestingModule } from '@angular/platform-browser-dynamic/testing';
+import { MockNgRedux } from '@angular-redux/store/lib/testing';
```
edgarmueller: This should be imported from `@angular-redux/store/testing`

packages/angular-material/test/object-control.spec.ts
```diff
@@ -113,7 +117,11 @@ describe('Object Control', () => {
     })
       .overrideModule(BrowserDynamicTestingModule, {
         set: {
-          entryComponents: [TextControlRenderer, VerticalLayoutRenderer]
+          entryComponents: [
+            TextControlRenderer,
+            VerticalLayoutRenderer,
+            GroupLayoutRenderer
```
edgarmueller: Can we have an assert condition that is explicit about the change? It seems rather funny to me that adding the GroupLayoutRenderer fixes the issue. Maybe we could check whether the renderered result contains a Group/Card component?

PR #1368, opened 2019-05-13
Fix/uischema no object generation

eneufeld:


eneufeld:
I suggest to merge #1366 first.

edgarmueller:
@eneufeld Needs to be rebased now

PR #1369, opened 2019-05-14
[Angular] Initial Table

eneufeld:


packages/angular-material/src/other/table.renderer.ts
```diff
@@ -0,0 +1,154 @@
...
+import {
+  ArrayControlProps,
+  ControlElement,
+  isObjectArrayControl,
+  isPrimitiveArrayControl,
+  JsonFormsState,
+  JsonSchema,
+  or,
+  OwnPropsOfRenderer,
+  // Paths,
```
edgarmueller: Remove comment

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -0,0 +1,154 @@
...
+  Paths,
+  RankedTester,
+  rankWith,
+  UISchemaElement
+} from '@jsonforms/core';
+
+@Component({
+  selector: 'ObjectRenderer',
```
edgarmueller: This seems wrong

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -0,0 +1,154 @@
...
+      class="mat-elevation-z8"
+      [trackBy]="trackElement"
+    >
+      <ng-container
+        *ngFor="let item of items"
+        matColumnDef="{{ item.property }}"
+      >
+        <th mat-header-cell *matHeaderCellDef>{{ item.header }}</th>
+        <td mat-cell *matCellDef="let index = index">
+          <!--{{ element[item.property] }}-->
```
edgarmueller: Remove comment

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -0,0 +1,154 @@
...
+          path: rowPath
+        }
+      }
+    ];
+  };
+
+  getValidColumnProps = (scopedSchema: JsonSchema) => {
+    if (scopedSchema.type === 'object') {
+      return Object.keys(scopedSchema.properties).filter(
+        prop => scopedSchema.properties[prop].type !== 'array'
```
edgarmueller: Do we only care about arrays being filtered out? Shouldn't we filter objects as well?

packages/angular-material/test/table-control.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+        }
+      }
+    });
+    mockSubStore.complete();
+    fixture.detectChanges();
+    component.ngOnInit();
+    fixture.whenStable().then(() => {
+      // 2 columns
+      expect(fixture.nativeElement.querySelectorAll('th').length).toBe(2);
+      // 1 head column and 2 data columns
```
edgarmueller: The comments seems off, I think you mean rows. It's the same for the other test cases.

packages/angular-material/test/table-control.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+      // 1 columns
+      expect(fixture.nativeElement.querySelectorAll('th').length).toBe(1);
+      // 1 head column and 2 data columns
+      expect(fixture.nativeElement.querySelectorAll('tr').length).toBe(1 + 2);
+      // 2 data entries
+      expect(fixture.nativeElement.querySelectorAll('td').length).toBe(2);
+    });
+  }));
+
+  xit('can be disabled', async(() => {
```
edgarmueller: Are these two tests disabled on purpose?

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -0,0 +1,154 @@
...
+          header: startCase(prop),
+          props: {
+            schema: schema,
+            uischema: controlWithoutLabel(`#/properties/${prop}`),
+            path: rowPath
+          }
+        };
+      });
+    }
+    return [
```
edgarmueller: Maybe add comment why it's necessary to add a dummy property.

packages/angular-material/src/other/table.renderer.ts
eneufeld: I copied from the react renderer. Yes it would make sense to filter out objects, too.

packages/angular-material/test/table-control.spec.ts
eneufeld: I think they are correct, as `th` is a `tableheader`.
Rows would be `tr`.

eneufeld:
I didn't manage to add a padding, the examples don't show an example with padding, I will try to add a custom css see https://material.angular.io/components/table/overview

packages/angular-material/test/table-control.spec.ts
edgarmueller: Yes, that was my point, the comment is referring to https://github.com/eclipsesource/jsonforms/pull/1369/files/b4c01e46cabaeefad3b18fa40a208d5795078c3f#diff-bd6faf34e11b664459093ead7e1ebfecR164 isn't it?

eneufeld: I see, I mixed up the loc, you are right. I will update it

packages/angular-material/test/table-control.spec.ts
eneufeld: I activated the enabled by default test case, the disabled one could be enabled after we merge #1371

edgarmueller: #1371 has been merged, we could update now.

PR #1371, opened 2019-05-15
Add Readonly utility function

eneufeld:
* Add support for angular renderers
* Add button to set/unset readonly in example

### DIFF
--- a/packages/core/src/models/uischema.ts
+++ b/packages/core/src/models/uischema.ts
@@ -246,5 +246,5 @@ export interface Categorization extends UISchemaElement {
 export const isGroup = (layout: Layout): layout is GroupLayout =>
   layout.type === 'Group';
 
-export const isLayout = (element: UISchemaElement): element is Layout =>
-  element.hasOwnProperty('elements');
+export const isLayout = (uischema: UISchemaElement): uischema is Layout =>
+  (uischema as Layout).elements !== undefined;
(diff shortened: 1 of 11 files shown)


packages/core/src/util/uischema.ts
```diff
@@ -0,0 +1,56 @@
...
+): void => {
+  if (!child.options) {
+    child.options = {};
+  }
+  child.options.readonly = value;
+};
+export const setReadonly = (uischema: UISchemaElement): void => {
+  iterateSchema(uischema, setReadonlyPropertyValue(true));
+};
+export const removeReadonly = (uischema: UISchemaElement): void => {
```
edgarmueller: I think `unsetReadonly` would be a better name (you also used it for the labeling of the button in the app.component)

packages/core/src/models/uischema.ts
```diff
@@ -245,3 +245,6 @@ export interface Categorization extends UISchemaElement {
 
 export const isGroup = (layout: Layout): layout is GroupLayout =>
   layout.type === 'Group';
+
+export const isLayout = (element: UISchemaElement): element is Layout =>
```
edgarmueller: We already have an `isLayout` function in `generators/`uischema`. Maybe you could unify both, such that the generator is then using the one exposed here.

Issue #1375, opened 2019-05-20
MaterialArrayLayout throws error if childData is undefined.


shivgarg5676:
I have a very nested schema. 
```
{
   "type":"object",
   "properties":{
      "componentTitle":{
         "type":"string"
      },
      "showOnNavigation":{
         "type":"boolean"
      },
      "domIdentifier":{
         "type":"string"
      },
      "description":{
         "type":"string"
      },
      "modules":{
         "type":"array",
         "items":{
            "type":"object",
            "properties":{
               "title":{
                  "type":"string"
               },
               "meta":{
                  "type":"array",
                  "items":{
                     "type":"object",
                     "properties":{
                        "count":{
                           "type":"number"
                        },
                        "description":{
                           "type":"string"
                        },
                        "iconSlug":{
                           "type":"string"
                        }
                     },
                     "required":[
                        "count",
                        "description",
                        "iconSlug"
                     ]
                  }
               },
               "instructors":{
                  "type":"array",
                  "items":{
                     "type":"object",
                     "properties":{
                        "name":{
                           "type":"string"
                        },
                        "imgURL":{
                           "type":"string"
                        }
                     },
                     "required":[
                        "name",
                        "imgURL"
                     ]
                  }
               },
               "subModules":{
                  "type":"array",
                  "items":{
                     "type":"object",
                     "properties":{
                        "title":{
                           "type":"string"
                        },
                        "type":{
                           "type":"string"
                        },
                        "duration":{
                           "type":"number"
                        }
                     },
                     "required":[
                        "title",
                        "type",
                        "duration"
                     ]
                  }
               },
               "tools":{
                  "type":"array",
                  "items":{
                     "type":"object",
                     "properties":{
                        "name":{
                           "type":"string"
                        },
                        "iconSlug":{
                           "type":"string"
                        }
                     },
                     "required":[
                        "name",
                        "iconSlug"
                     ]
                  }
               }
            },
            "required":[
               "title",
               "meta",
               "instructors"
            ]
         }
      },
      "summary":{
         "type":"object",
         "properties":{
            "items":{
               "type":"array",
               "items":{
                  "type":"object",
                  "properties":{
                     "count":{
                        "type":"string"
                     },
                     "description":{
                        "type":"string"
                     }
                  },
                  "required":[
                     "count",
                     "description"
                  ]
               }
            }
         },
         "required":[
            "items"
         ]
      }
   },
   "required":[
      "componentTitle",
      "showOnNavigation",
[...]

PR #1377, opened 2019-05-21
undefined check in material array layout

shivgarg5676:
Added an undefined check at line `232`.
All tests are green.


edgarmueller:
Thanks for the PR! I'm not yet sure why this error actually occurs as the renderer checks for empty data and displays a `No data` text, but we'll have a look

edgarmueller:
Closing this, see https://github.com/eclipsesource/jsonforms/issues/1375#issuecomment-498978887.

PR #1382, opened 2019-05-23
Provide context-based API

edgarmueller:
Add context-based mapping functions to react package.
Disconnect material renderer set and provide redux based context.

This is a WIP, there's are still some things to do:
- dispatch currently only handles core actions
- typings (react dispatch vs. redux dispatch)

### DIFF
--- a/packages/material/src/additional/MaterialListWithDetailRenderer.tsx
+++ b/packages/material/src/additional/MaterialListWithDetailRenderer.tsx
@@ -29,211 +29,96 @@ import {
   computeLabel,
   createDefaultValue,
   findUISchema,
-  getData,
   isObjectArray,
   isPlainLabel,
-  JsonFormsState,
-  JsonSchema,
-  mapDispatchToArrayControlProps,
-  mapStateToArrayLayoutProps,
   RankedTester,
   rankWith,
-  Resolve,
   uiTypeIs
 } from '@jsonforms/core';
-import { ResolvedJsonForms } from '@jsonforms/react';
+import { JsonFormsDispatch, withJsonFormsArrayLayoutProps } from '@jsonforms/react';
 import {
-  Avatar,
   Grid,
   Hidden,
-  IconButton,
   List,
-  ListItem,
-  ListItemAvatar,
-  ListItemSecondaryAction,
-  ListItemText,
   Typography
 } from '@material-ui/core';
-import DeleteIcon from '@material-ui/icons/Delete';
-import find from 'lodash/find';
 import map from 'lodash/map';
 import range from 'lodash/range';
-import React from 'react';
-import { connect } from 'react-redux';
+import React, { useCallback, useState } from 'react';
 import { ArrayLayoutToolbar } from '../layouts/ArrayToolbar';
+import ListWithDetailMasterItem from './ListWithDetailMasterItem';
 
-interface MaterialListWithDetailRendererState {
-  selectedIndex: number;
-}
-export class MaterialListWithDetailRenderer extends React.Component<
-  ArrayLayoutProps,
-  MaterialListWithDetailRendererState
-> {
-  state: MaterialListWithDetailRendererState = {
-    selectedIndex: undefined
-  };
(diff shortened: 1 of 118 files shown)


packages/material/src/additional/MaterialListWithDetailRenderer.tsx
```diff
@@ -24,213 +24,115 @@
...
-    if (this.state.selectedIndex === value) {
-      this.state.selectedIndex = undefined;
-    } else if (this.state.selectedIndex > value) {
-      this.state.selectedIndex--;
+export const MaterialListWithDetailRenderer = (props: OwnPropsOfControl) => {
+  const [selectedIndex, setSelectedIndex] = useState(undefined);
+  const ctx = useJsonForms();
+  const { required, visible, errors, path, data, schema, label, uischema, uischemas } = ctxToArrayLayoutProps(ctx, props);
+  const { addItem, removeItems } = ctxDispatchToArrayControlProps(ctx.dispatch);
+  const handleRemoveItem = (p: string, value: any) => () => {
```
eneufeld: If I understand the life cycle correctly, then the handleRemoveItem will be recreated every time we run through here and thus the ListWithDetailMasterItem will be re-rendered.

packages/material/src/controls/MaterialBooleanControl.tsx
```diff
@@ -24,47 +24,47 @@
...
-          visible={visible}
-          {...props}
-        />
-      }
-    />
-  </Hidden>
-);
+export const MaterialBooleanControl = (props: OwnPropsOfControl) => {
+  const ctx = useJsonForms();
+  const { data, enabled, uischema, path, visible, label, id, errors, schema, rootSchema } = ctxToControlProps(ctx, props);
```
eneufeld: Maybe use a HigherOrderComponent to not duplicate code and have less changes on the renderers?

packages/react/src/JsonForms.tsx
```diff
@@ -159,7 +156,46 @@ export class JsonFormsDispatchRenderer extends ResolvedJsonFormsDispatchRenderer
     }
 }
 
-export const JsonForms = connect<StatePropsOfJsonFormsRenderer, {}, OwnPropsOfJsonFormsRenderer>(
-    mapStateToJsonFormsRendererProps,
-    null
-)(JsonFormsDispatchRenderer);
+export const JsonFormsDispatch = (props: OwnPropsOfJsonFormsRenderer) => {
+    const ctx = useJsonForms();
+    const { refResolver } = mapStateToJsonFormsRendererProps(
```
eneufeld: why not use ctxToJsonFormsDispatchProps?

packages/material/src/additional/MaterialListWithDetailRenderer.tsx
edgarmueller: Yes, that's the case currently, but we should be able to memoize it with `useCallback`, so that we can avoid any re-render

packages/react/src/JsonFormsContext.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+export const ctxToLayoutProps = (ctx: JsonFormsStateContext, props: OwnPropsOfJsonFormsRenderer) =>
+  mapStateToLayoutProps({ jsonforms: { ...ctx } }, props);
+
+export const ctxToControlProps = (ctx: JsonFormsStateContext, props: OwnPropsOfControl) =>
+  mapStateToControlProps({ jsonforms: { ...ctx } }, props);
+
+export const ctxToControlWithDetailProps = (
+  ctx: JsonFormsStateContext,
+  props: OwnPropsOfControl
```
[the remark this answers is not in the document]
edgarmueller: Hm, I'm not sure what's inconsistent here?

edgarmueller:
@eneufeld I've addressed your review comments and removed the draft status.

eneufeld:
@edgarmueller  could you fix the commit message?

PR #1384, opened 2019-05-24
Completed sort buttons for material array layout

shivgarg5676:


shivgarg5676:
@edgarmueller  Any idea why is build failing. Why aren't ArrowUpWard and ArrowDownWard getting resolved in ci environment?

edgarmueller:
@shivgarg5676 I think you are using the wrong casing on the ArrowDownward import, it should read `import ArrowDownward from '@material-ui/icons/ArrowDownward'; ` (note lower cased `w`). I don't have much time today, but I'll do a review tomorrow.

shivgarg5676:
Thanks, @edgarmueller. I work on the mac which has a case insensitive file system, that's why the builds were working on my local system. Anyways solved the problem.

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -202,60 +209,115 @@ interface NonEmptyRowProps {
...
+    isLast,
+    sortButtons
+  }: NonEmptyRowProps & WithDeleteDialogSupport) => {
+    return (
+      <TableRow key={childPath} hover>
+        {generateCells(NonEmptyCell, schema, childPath)}
+        <NoBorderTableCell
+          style={sortButtons ? styles.fixedCell : styles.fixedCellSmall}
+        >
+          <div style={{ display: 'flex', justifyContent: 'center' }}>
```
eneufeld: I think the div should be removed as the Grid below is handling the correct item placement now

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -202,60 +209,115 @@ interface NonEmptyRowProps {
...
+              justify="center"
+              alignItems="center"
+            >
+              {sortButtons ? (
+                <Fragment>
+                  <Grid item>
+                    <IconButton
+                      aria-label={`move up`}
+                      onClick={moveUp}
+                      disabled={rowIndex == 0}
```
eneufeld: this is a bit weird as we handle the disabled states of the sort buttons differently, maybe provide two props: `enableUp` and `enableDown` or similar?

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -202,60 +209,115 @@ interface NonEmptyRowProps {
...
+                    >
+                      <ArrowUpward />
+                    </IconButton>
+                  </Grid>
+                  <Grid item>
+                    <IconButton
+                      aria-label={`move down`}
```
eneufeld: the aria labels are inkonsistent with the upper case/lower case naming, eg `move down` vs `Delete`, I don't know what the recommended way for the aria labels is though

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -202,60 +209,115 @@ interface NonEmptyRowProps {
...
           <NonEmptyRow
             key={childPath}
             childPath={childPath}
             rowIndex={index}
             schema={schema}
             openDeleteDialog={openDeleteDialog}
+            moveUp={moveUp(path, index)}
+            moveDown={moveDown(path, index)}
+            isLast={index === data - 1}
+            sortButtons={uischema.options && uischema.options.sortButtons}
```
eneufeld: for the showing of the buttons we should also take the enablement state of the whole table into account, as we probably don't want to show the buttons if the table is not editable

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -148,25 +160,60 @@ class ExpandPanelRenderer extends React.Component<ExpandPanelProps, any> {
...
+                          style={iconStyle}
+                          disabled={index == 0}
+                        >
+                          <ArrowUpward />
+                        </IconButton>
+                      </Grid>
+                    ) : (
+                      ''
+                    )}
+                    {uischema.options && uischema.options.sortButtons ? (
```
eneufeld: why do we check the up/down buttons individually? I think it would be easier to simply reuse the props from the table

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -257,6 +310,34 @@ export const mapDispatchToExpandPanelProps: (dispatch: Dispatch<AnyAction>) => D
         return array;
       })
     );
+  },
+  moveUp: (path: string, toMove: number) => (event: any): void => {
```
eneufeld: here we could ideally reuse the move method from above to have less code

shivgarg5676:
Thanks, @eneufeld for the comments. I will try to solve them today if I am able to find time after work.

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -114,7 +124,7 @@ export interface EmptyTableProps {
 const EmptyTable = ({ numColumns }: EmptyTableProps) => (
   <TableRow>
     <NoBorderTableCell colSpan={numColumns}>
-      <Typography align='center'>No data</Typography>
+      <Typography align="center">No data</Typography>
```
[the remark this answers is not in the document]
shivgarg5676: I have ts lint and prettier configured, but the thing is prettier is automatically changing it to double quote when I am trying to save. Am I missing on any config? I can't see any problem reported by ts-lint.

packages/material/src/layouts/MaterialArrayLayout.tsx
shivgarg5676: From the above means? Are you referring to the helper method? or methods I created in packages/core/src/util/renderer.ts

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -148,25 +163,58 @@ class ExpandPanelRenderer extends React.Component<ExpandPanelProps, any> {
...
-                    style={iconStyle}
+                  <Grid
+                    container
+                    direction="row"
+                    justify="center"
+                    alignItems="center"
                   >
-                    <DeleteIcon />
-                  </IconButton>
+                    {uischema.options && uischema.options.showSortButtons ? (
```
shivgarg5676: @eneufeld `why do we check the up/down buttons individually? I think it would be easier to simply reuse the props from the table`
Can you elaborate this more. I am not able to follow over here.

eneufeld: You can ignore it now, in the first version you had the `uischema.options && uischema.options.showSortButtons` check before each button. Now it is fine!

packages/material/src/layouts/MaterialArrayLayout.tsx
eneufeld: I meant the helper method from `packages/core/src/util/array.ts`

eneufeld:
Thank you for the changes. I will try to review the new code later today.

shivgarg5676:
@eneufeld I haven't done changes for the disabled state(when the whole table is disabled). I will do those changes and let you know to review.

packages/material/src/complex/MaterialTableControl.tsx
shivgarg5676: @eneufeld  here you mean to have a check based on the `edited` prop? 
Why I am asking this is, because I could not see any similar check for the display of delete button. Moreover, there is no logic in this component based on that particular prop.

eneufeld: Yes, this is missing which is a bug as we cannot disable the table atm

shivgarg5676: can we skip the particular suggestion for now? This can be solved when we solve the bug.

packages/core/test/util/array.test.ts
```diff
@@ -0,0 +1,26 @@
+import { moveUp, moveDown } from '../../src/util/';
+import test from 'ava';
+let array: Array<number> = [];
+test.beforeEach(() => {
+  array = [1, 2, 3, 4, 5];
```
shivgarg5676: Haven't used ava before. Don't know if reassigning a common variable is fine or it will lead to race conditions. Please suggest here.

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -202,60 +209,113 @@ interface NonEmptyRowProps {
   childPath: string;
   schema: JsonSchema;
   rowIndex: number;
+  moveUp: () => any;
```
eneufeld: do we need the return value?

packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -202,60 +209,113 @@ interface NonEmptyRowProps {
...
+  moveDown: () => any;
```
eneufeld: do we need the return value?

packages/material/src/layouts/MaterialArrayLayout.tsx
```diff
@@ -148,25 +162,58 @@ class ExpandPanelRenderer extends React.Component<ExpandPanelProps, any> {
...
+                            disabled={index == 0}
+                          >
+                            <ArrowUpward />
+                          </IconButton>
+                        </Grid>
+                        <Grid item>
+                          <IconButton
+                            onClick={moveDown(path, index)}
+                            style={iconStyle}
+                            disabled={isLast}
```
eneufeld: I would suggest to unify the disable logic of Up/Down like in the MaterialLabelControl

packages/material/test/renderers/MaterialArrayControl.test.tsx
```diff
@@ -309,4 +351,118 @@ describe('Material array control', () => {
...
+      <Provider store={store}>
+        <MaterialArrayControlRenderer
+          schema={fixture2.schema}
+          uischema={fixture2.uischema}
+        />
+      </Provider>
+    );
+    //first row is header in table
+    // first buttton is up arrow, second button is down arrow
+    const downButton = wrapper
```
eneufeld: can't we find the button using the aria label? this would make the test more robust against layout changes

packages/material/src/complex/MaterialTableControl.tsx
shivgarg5676: changed return type to void

packages/core/src/util/array.ts
```diff
@@ -0,0 +1,16 @@
+const move = function(array: Array<any>, index: number, delta: number) {
+  const newIndex: number = index + delta;
+  if (newIndex < 0 || newIndex >= array.length) return; //Already at the top or bottom.
+  var indexes: number[] = [index, newIndex].sort(); //Sort the indixes
```
eneufeld: did you check for the bug described here: https://gist.github.com/albertein/4496103 ?

packages/core/src/util/array.ts
```diff
@@ -0,0 +1,16 @@
...
+  array.splice(indexes[0], 2, array[indexes[1]], array[indexes[0]]);
+};
+
+var moveUp = function(array: Array<any>, toMove: number) {
```
eneufeld: I suggest to use `const` instead of `var` and Fat Arrows instead of function. 
ES Lint should complain, but something is not working as expected there

packages/example/src/index.tsx
```diff
@@ -41,6 +41,11 @@ import { getExamples } from '@jsonforms/examples';
 import { AdditionalStoreParams, exampleReducer } from './reduxUtil';
 import { enhanceExample, ReactExampleDescription } from './util';
 
+declare global {
+  interface Window {
+    __REDUX_DEVTOOLS_EXTENSION_COMPOSE__?: typeof compose;
```
eneufeld: I think this was not intentional?

shivgarg5676: yes, removed

eneufeld:
@shivgarg5676 Thank you very much!
@edgarmueller could you double check that I didn't miss anything?

shivgarg5676:
Thanks, @eneufeld, I will make sure my next prs have conventions in place so that I consume less time of yours. Sorry for reiterations on same things.

eneufeld:
@shivgarg5676 No need to apologize. We are very happy to have more contributors. We need to improve our tooling so that checks don't need to be done manually for things like linting and formatting. Thank you for the contribution!

Issue #1347, opened 2019-04-25
Support async validation with ajv


edgarmueller:
**Describe the bug**
If ajv is configured to use asynchronous validation, error reporting is not working correctly, see  https://github.com/epoberezkin/ajv#asynchronous-validation for more info.

**To Reproduce**
Use any example with $async set

**Expected behavior**
Error reporting should not break.

mlandry8:
Are there any updates on this issue? Have a hacky workaround going but it's not ideal. I see there is a solution ready in your PRs!

edgarmueller:
@mlandry8 Except the open draft PR there isn't unfortunately, we simply didn't find the time to finish it. But we'll try to do so for the next minor release (which would be 2.4.0). If you already have something and you think it's a valuable addition to the framework, we always welcome contributions! Maybe you could compare it to the existing PR and check which approach would be better.

sdirix:
Update: While JSON Forms does not support async validation directly, there is a straightforward workaround by managing the errors outside of JSON Forms:

 - Hand over `validationMode="NoValidation"` to turn off the built-in validation
 - Create an AJV instance outside of JSON Forms, e.g. via `createAjv()` exported by `@jsonforms/core`
 - Compile your schema with the ajv instance, e.g. `const validator = myAjv.compile(schema)` (this is costly, so the validator should be memoized to not compile the schema again with every change)
 - Whenever data is received from JSON Forms, use the validator to validate and store the errors in some state
 - Handover the determined errors to AJV via `additionalProps={myErrors}`

PR #1386, opened 2019-05-28
Async Validation with Ajv

sdirix:
This (draft) PR introduces the concept of [async validation with Ajv](https://github.com/epoberezkin/ajv#asynchronous-validation) for #1347.

This concept is introduced next to (instead of replacing) synchronous validation. To implement the functionality we use [redux-thunk](https://github.com/reduxjs/redux-thunk) to allow for chained async actions, however redux-thunk is only necessary when the user wants to use async-marked schemas.

This (draft) PR contains the thunk for JSON Forms initialization with an async Ajv validation run, see `Actions.initAsyncValidation`. The (currently) last commit enables the async behavior for all examples, however only the person example is actually marked async (the remaining examples use a fallback).

Open tasks:
 - Implement same behavior for the remaining core actions which perform validation: `SET_AJV` and `UPDATE_DATA`.
 - Check whether we can offer dispatching-functions which either use the normal action or the preconfigured thunk depending on whether the current schema is marked asynchronous.
 - Make sure the correct update data action / thunk is called depending on the circumstances
 - Write testcases.

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -33,15 +33,19 @@ import RefParser from 'json-schema-ref-parser';
 import {
   INIT,
   InitAction,
+  InitActionWithoutValidation,
   InitActionOptions,
   SET_AJV,
+  SET_ERRORS,
   SET_SCHEMA,
   SET_UISCHEMA,
   SetAjvAction,
+  SetErrorsAction,
   SetSchemaAction,
   SetUISchemaAction,
   UPDATE_DATA,
-  UpdateAction
+  UpdateAction,
+  INIT_WITHOUT_VALIDATION
 } from '../actions';
 import { createAjv } from '../util/validator';
 import { JsonSchema, UISchemaElement } from '..';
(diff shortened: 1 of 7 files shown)


edgarmueller:
Thank you, Stefan!

packages/core/src/reducers/core.ts
```diff
@@ -51,16 +55,35 @@ const validate = (validator: ValidateFunction, data: any): ErrorObject[] => {
...
-  validate(validator, data).map(error => {
-    error.dataPath = error.dataPath.replace(/\//g, '.').substr(1);
+const sanitizeError = (error: ErrorObject): ErrorObject => {
+  error.dataPath = error.dataPath.replace(/\//g, '.').substr(1);
+  return error;
+};
 
-    return error;
-  });
+export const sanitizeErrors = (validator: ValidateFunction, data: any) => {
```
eneufeld: It would be better to have a return type

packages/core/src/reducers/core.ts
```diff
@@ -51,16 +55,35 @@ const validate = (validator: ValidateFunction, data: any): ErrorObject[] => {
...
+  if (validator.$async) {
+    // sync method called with async schema, fallback
+    return [];
+  } else {
+    return validate(validator, data).map(sanitizeError);
+  }
+};
+
+export const asyncSanitizeErrors = (validator: ValidateFunction, data: any) => {
```
eneufeld: It would be better to have a return type

PR #1392, opened 2019-06-04
Configure linter and fix any linting errors

eneufeld:


edgarmueller:
@eneufeld Any idea why this is failing now?

eneufeld:
The reason is that the typing seam to be outdated:
`Property 'text' is incompatible with index signature.
lerna ERR!     Type '{ encoding: string; }' is not assignable to type 'boolean | ParserOptions'.`

eneufeld:
I see the same error in vs code but not when I lint the code, I will double check

Issue #1254, opened 2019-02-08
Add support for JSONSchema constants


johnmorrell:
**Is your feature request related to a problem? Please describe.**
Where a JSONSchema contains a constant value this does not get reflected in the output model. If this was a required property the output JSON would never validate against the Schema.

**Describe the solution you'd like**
Using JSONSchema constant:
```
{
   "type":"object",
   "properties":{
      "version":{
         "const":"1.0"
      }
   },
   "required":[
      "version"
   ]
}
```
Constants to be added to the model, via either a hidden or readonly form element.

**Describe alternatives you've considered**
Currently forcing constants by using an required Enum with a single and default value:
```
{
   "type":"object",
   "properties":{
      "version":{
         "type":"string",
         "enum":[
            "1.0"
         ],
         "default":"1.0"
      }
   },
   "required":[
      "version"
   ]
}
```
**Describe for which setup you like to have the improvement**
Framework: React
RendererSet: Material

**Additional context**
JSONForms: v2.2.0

edgarmueller:
Thanks for the report, we'll try to have a look soon.

johnmorrell:
Just for reference,  I have seen good results with populating oneOf data in v2.2.3 since #1273. I did hit a couple of cases where the intended option was not selected, due to partially complete data or where the oneOf structures are similar with optional fields. I think these case could be rectified with support for constants. Each of the oneOf types could be given a required const identifier that would have to be valid preventing early validation passes.

edgarmueller:
Alright, thanks for letting us know. I hope we'll be able to tackle this for 2.3.0.

edgarmueller:
https://github.com/eclipsesource/jsonforms/pull/1393 added support for `const`. I'm closing this therefore now. Any additional features related to `const` should be opened in a new issue. Thank you!

PR #1393, opened 2019-06-04
#1254 - Add support for JSONSchema constants

Lily418:
The JSON reference for constant values in JSONSchema says
"It should be noted that const is merely syntactic sugar for an enum
with a single element"
https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

So this is the way I've implemented them.

### DIFF
--- a/packages/core/test/testers.test.ts
+++ b/packages/core/test/testers.test.ts
@@ -518,8 +518,7 @@ test('test isEnumControl', t => {
   t.true(
     isEnumControl(t.context.uischema, {
       type: 'object',
-      properties: { foo: { const: '1.0' } },
-      required: ['version']
+      properties: { foo: { const: '1.0' } }
     })
   );
 });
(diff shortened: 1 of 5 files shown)


packages/examples/src/1254.ts
```diff
@@ -0,0 +1,67 @@
...
+  },
+  required: ['version']
+};
+
+export const uischema = {
+  type: 'HorizontalLayout',
+  elements: [
+    {
+      type: 'Control',
+      label: 'Value',
```
edgarmueller: Just a minor that I noticed, this should be probably be labelled as 'Name'.

Lily418:
@edgarmueller have updated ✨

packages/core/test/testers.test.ts
```diff
@@ -515,6 +515,13 @@ test('test isEnumControl', t => {
       properties: { foo: { type: 'number', enum: [1, 2] } }
     })
   );
+  t.true(
+    isEnumControl(t.context.uischema, {
+      type: 'object',
+      properties: { foo: { const: '1.0' } },
+      required: ['version']
```
eneufeld: the property `version` does not exist in this schema, I think this is a left over from the example, I would suggest to remove it

Lily418:
@eneufeld thanks, I've removed that now

PR #1404, opened 2019-06-13
Add config for layout direction of group

Lily418:
I have a custom renderer which looks like this

```
import * as React from 'react'
import { connect } from 'react-redux';
import { MaterialHorizontalLayout } from '@jsonforms/material-renderers'
import { withJsonFormsLayoutProps } from '@jsonforms/react';


const WithUnit = (props) => {
  let uiSchema = props.uischema
  uiSchema.type = "HorizontalLayout"
  return  (<MaterialHorizontalLayout {...props}/>)
}


export default withJsonFormsLayoutProps(WithUnit);
```

I use this as the default layout is vertical but in some cases I'd like to make it horizontal. I'd like to do this without writing UISchema for every part of my Schema. So instead I use a tester to write the rule for when I'd like to render horizontally. 

I'd also like to be able to do something similar for GroupLayouts so I've proposed this change which would allow overriding of the default layout direction.


eneufeld:
Could you explain how you switch from vertical to horizontal layout without touching the schema?

Lily418:
I have a tester

```js
import {
  rankWith
} from '@jsonforms/core'

export default rankWith(
  Number.MAX_VALUE,
  (x) => {
    if (!x.elements) {
      return false
    }

    const lastElement = x.elements[x.elements.length - 1]

    return lastElement.scope && lastElement.scope.endsWith('unit')
  }
)
```

and a renderer

```js
import * as React from 'react'
import { connect } from 'react-redux';
import { MaterialHorizontalLayout } from '@jsonforms/material-renderers'
import { withJsonFormsLayoutProps } from '@jsonforms/react';


const WithUnit = (props) => {
  let uiSchema = props.uischema
  uiSchema.type = "HorizontalLayout"
  return  (<MaterialHorizontalLayout {...props}/>)
}


export default withJsonFormsLayoutProps(WithUnit);
```


My UI Schema lists properties like
```json
      {
         "type":"Category",
         "label":"Doses",
         "elements":[
            {
               "type":"Control",
               "scope":"#/properties/suggestedDose"
            }
         ]
      }
```

but suggestedDose is actually a very complex schema, ~300 lines with references to a defintions.json file which is ~1500 lines.

So the thing I'm trying to do is to avoid needing to write lots of UISchema to go along with the Schema. Instead I use the custom renderer shown above.

edgarmueller:
@Lily418 So if I got this right you want to change the default orientation of the group renderer to be horizontal based on the `config` state (which you seem to maintain yourself), is that correct?

Lily418:
I'd like to write a tester which matches places where MaterializedGroupLayoutRenderer would be used,  then a custom renderer which uses MaterializedGroupLayoutRenderer but adds the config option. 

So a tester like 
```js
import {
  rankWith
} from '@jsonforms/core'

export default rankWith(
  Number.MAX_VALUE,
  (x) => {

    if (!x.elements || type !== "Group" ) {
      return false
    }

    // Some custom logic to only select some of the groups
    const lastElement = x.elements[x.elements.length - 1]

    return lastElement.scope && lastElement.scope.endsWith('unit')
  }
)
```

and a renderer which reuses  MaterializedGroupLayoutRenderer
```js
import * as React from 'react'
import { MaterializedGroupLayoutRenderer } from '@jsonforms/material-renderers'
import { withJsonFormsLayoutProps } from '@jsonforms/react'

// Wrote this quickly for demo purposes so not sure correct 
// but here I inject the configuration 
const GroupLayoutHorizontalLayout = (props) => {
  return (<MaterializedGroupLayoutRenderer {...props} config={Object.assign({}, props.config, {'direction': 'row'})} />)
}

export default withJsonFormsLayoutProps(GroupLayoutHorizontalLayout)

```

Lily418:
Apologies, I destroyed this PR while trying to make another one

Lily418:
Okay I'll try not to do that again 😆

edgarmueller:
We got a quick discussion and we think it would make sense to introduce a `direction` property to `LayoutProps`. So the group renderer would receive its direction not via config, but directly via `direction`. This will likely involve a bit mork as we would also need adapt other layout renderers, but it seems cleaner. What do you think?

Lily418:
Sounds cool! I can update this PR

Lily418:
Hi @edgarmueller is this what you were thinking about?

packages/core/src/util/renderer.ts
```diff
@@ -321,7 +325,9 @@ export interface StatePropsOfLayout extends StatePropsOfRenderer {
   renderers?: any[];
 }
 
-export interface LayoutProps extends StatePropsOfLayout {}
+export interface LayoutProps extends StatePropsOfLayout {
+  direction: 'row' | 'column';
```
edgarmueller: Move `direction` to `StatePropsOfLayout`.

packages/core/src/util/renderer.ts
```diff
@@ -640,7 +646,8 @@ export const mapStateToLayoutProps = (
     enabled,
     path: ownProps.path,
     uischema: ownProps.uischema,
-    schema: ownProps.schema
+    schema: ownProps.schema,
+    direction: ownProps.direction || 'column'
```
edgarmueller: The default value should be moved to `layoutDefaultProps` and used here

Lily418:
@edgarmueller Thanks, have updated 🦄

PR #1406, opened 2019-06-17
[react] Improve performance related to context

edgarmueller:
Split usage of context and mapping of properties by using additional HOCs which allows usage of React.memo as only the properties can be checked via an `areEqual` function.

Only perform call to `hasRefs` within `ResolvedJsonFormsDispatchRenderer` if schemas are not equal.

Provide CombinatorProps intersection type.

### DIFF
--- a/packages/react/src/JsonFormsContext.tsx
+++ b/packages/react/src/JsonFormsContext.tsx
@@ -23,11 +23,16 @@
   THE SOFTWARE.
 */
 
-import React, { ComponentType, Dispatch, ReducerAction, useContext, useEffect, useReducer } from 'react';
+import React, { ComponentType, Dispatch, ReducerAction, useContext, useEffect, useReducer, useCallback } from 'react';
+import isEqual from 'lodash/isEqual';
+import omit from 'lodash/omit';
+import get from 'lodash/get';
 import {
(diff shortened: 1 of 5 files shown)


packages/react/src/JsonFormsContext.tsx
```diff
@@ -205,107 +210,219 @@ export const ctxToCellProps = (
...
+    };
+
+// --
+
+type JsonFormsPropTypes = ControlProps | CombinatorProps | LayoutProps | CellProps;
+
+export const areEqual = (prevProps: JsonFormsPropTypes, nextProps: JsonFormsPropTypes) => {
+  const prev = omit(prevProps, ['handleChange', 'renderers', 'cells', 'uischemas']);
+  const next = omit(nextProps, ['handleChange', 'renderers', 'cells', 'uischemas']);
+  return isEqual(prev, next)
```
eneufeld: shouldn't this be:
`isEqual && renderers.length==renderers.length && cells.length === cells.length && uischemas.length === uischemas.length`

edgarmueller: Right I somehow managed to switch the logic in between, thanks for noticing

PR #1413, opened 2019-06-21
Export MaterialInputControl

Lily418:
I'd like to make a custom renderer that uses MaterialInputControl. This is so I can override the onChange so I can create a TextField which automatically converts the input to Sentence Case.

I haven't tested this yet but I imagine something like
```jsx
import * as React from "react";
import { Control } from "@jsonforms/react";
import { connect } from "react-redux";
import {
  MuiInputText,
  MaterialInputControl
} from "@jsonforms/material-renderers";

export class SentenceCaseRender extends Control {
  render() {
    const { path, handleChange } = this.props;
    return (
      <MaterialInputControl
        input={MuiInputText}
        {...this.props}
        onChange={e => handleChange(path, sentenceCase(ev.target.value))}
      />
    );
  }
}

export default SentenceCaseRender;
```

### DIFF
--- a/packages/material/src/controls/index.ts
+++ b/packages/material/src/controls/index.ts
@@ -53,6 +53,8 @@ import MaterialTextControl, {
   materialTextControlTester
 } from './MaterialTextControl';
 
+import { MaterialInputControl } from './MaterialInputControl';
+
 import MaterialAnyOfStringOrEnumControl, {
   materialAnyOfStringOrEnumControlTester
 } from './MaterialAnyOfStringOrEnumControl';
@@ -78,6 +80,7 @@ export {
   materialNumberControlTester,
   MaterialTextControl,
   materialTextControlTester,
+  MaterialInputControl,
   MaterialAnyOfStringOrEnumControl,
   materialAnyOfStringOrEnumControlTester
 };


eneufeld:
I see your use case. I'm wondering though whether you really want to reuse the MaterialInputControl or actually the MaterialTextControl?
The MaterialInputControl is used for integers, numbers and so on. 
I could imagine such a code using the MaterialTextControl:
```typescript
import { MaterialTextControl } from '../../material/src/controls/MaterialTextControl';
import * as React from 'react';
import {
  ControlProps,
  RankedTester,
  rankWith,
  isStringControl
} from '../../core/lib';

import { withJsonFormsControlProps } from '@jsonforms/react';

export const SentenceCaseRender = (props: ControlProps) => {
  const { handleChange } = props;
  return (
    <MaterialTextControl
      {...props}
      handleChange={(path: string, value: string) =>
        handleChange(path, sentenceCase(value))
      }
    />
  );
};
const sentenceCase = (value: string) => {
  console.log(value);
  return value;
};

export const myControlTester: RankedTester = rankWith(2, isStringControl);
export default withJsonFormsControlProps(SentenceCaseRender);
```

So the export must be adapted anyways, the question is which control to export.

Lily418:
Hi @eneufeld  thank you for your help! 🎉

I'm using v2.2.3 currently

```jsx
import * as React from "react";
import { Control } from "@jsonforms/react";
import { connect } from "react-redux";
import {
  mapDispatchToControlProps,
  mapStateToControlProps
} from '@jsonforms/core';
import {
  MaterialTextControl
} from "@jsonforms/material-renderers";

export class SentenceCaseRender extends Control {

  sentenceCaseString(s) {
    return s.length > 0 ? s[0].toUpperCase() + s.slice(1).toLowerCase() : s
  }

  render() {
    const { handleChange, path } = this.props;
    return (
      <MaterialTextControl.WrappedComponent
        {...this.props}
        handleChange={(path, targetValue) => {
          handleChange(path, this.sentenceCaseString(targetValue))
        }}
      />
    );
  }
}
export default connect(
  mapStateToControlProps,
  mapDispatchToControlProps
)(SentenceCaseRender);
```
So I used your suggestion but took the WrappedComponent in order to override handleChange. 

Going to close this PR as I didn't make any changes. Maybe there is a change in v2.3.0 to support this but will cross that bridge when I come to it.

Issue #1203, opened 2018-12-21
Update docs


edgarmueller:
**Describe the bug**
This issue will act as a bucket for missing/outdated documentation. Some things off the top of my head include:

* default data reducer
* detail options in UI schema
* object renderer (website example?)
* empty string to undefined conversion
* ListWithDetail element & renderer (website example?)
* anyOf/allOf/oneOf renderers (website examples?)

sdirix:
Detail options for controls referring to arrays (only react/material) are documented in #1198

edgarmueller:
Issues have been re-created on the website repo: https://github.com/eclipsesource/jsonforms2-website/issues

eneufeld:
Reopen to easier track it

edgarmueller:
Here's more:
* locale related actions (these are probably not supported yet by the context-based API at the time of writing)
* Usage of `cells`
* Handling of default values
* Move up/down buttons in array layout renderer
* ListWithDetail usage
* Angular related docs
* Link to examples
* `config` options

sdirix:
This issue will be closed as it has been open for a long time with no activity. Feel free to reopen this issue if needed.

PR #1414, opened 2019-06-21
Show unfocused description config

Lily418:
I was doing user testing yesterday and found the user didn't know what unit one of the inputs required, this was it the description but this only showed when they focused the field.

I'd like the option to show unfocused helper text.
Screenshot 2019-06-21 at 15 38 16

### DIFF
--- a/packages/core/src/configDefault.ts
+++ b/packages/core/src/configDefault.ts
@@ -38,5 +38,5 @@ export const configDefault = {
   /*
    * [text] if input descriptions should hide when not focused
    */
-   showUnfocusedDescription: false
+  showUnfocusedDescription: false
 };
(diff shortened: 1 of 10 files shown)


eneufeld:
I agree the new option would be useful. I would not mix in commit 1b8d51fe53a0b6322b4bb528a547e853405a5b48 (which is pr #1413 ) into this though.

Lily418:
@eneufeld opps, that wasn't meant to be there. Thanks have updated now.

eneufeld:
@edgarmueller what is your take on this? I think this is a nice extension.

edgarmueller:
I agree, actually we have the same requirement in another project, so I'm all in favor for it!

edgarmueller:
On a sidenote: this adds to the long list of TODOs about documentation, I've updated https://github.com/eclipsesource/jsonforms/issues/1203 accordingly

PR #1415, opened 2019-06-21
Seperate FormHelperText into description and errors

Lily418:
I'd like to be able to see the description even when there are validation errors. I've seperated them into two seperate FormHelperText.

https://giphy.com/gifs/UQD8EYWLpnqMxiRp0f/fullscreen

### DIFF
--- a/packages/material/src/controls/MaterialInputControl.tsx
+++ b/packages/material/src/controls/MaterialInputControl.tsx
@@ -69,6 +69,14 @@ export abstract class MaterialInputControl extends Control<ControlProps & WithIn
       description,
       this.state.isFocused
     );
+
+    const firstFormHelperText = showDescription
+      ? description
+      : !isValid
+      ? errors
+      : null;
+    const secondFormHelperText = showDescription && !isValid ? errors : null;
+
     const InnerComponent = input;
     return (
       <Hidden xsUp={!visible}>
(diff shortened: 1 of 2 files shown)


eneufeld:
I think that makes sense. @edgarmueller what do you think?

packages/material/src/controls/MaterialInputControl.tsx
```diff
@@ -91,12 +91,13 @@ export abstract class MaterialInputControl extends Control<ControlProps & WithIn
             isValid={isValid}
             visible={visible}
           />
+          {description !== undefined ? (
+            <FormHelperText>
+              {showDescription || !isValid ? description : null}
```
eneufeld: why check for an invalid state here? I think it makes sense to only check for showDescription

edgarmueller: I kinda agree. Both `FormHelperText`s should behave the same way: ither include them both and let their children be `null` or let check them condtionaly

edgarmueller:
@eneufeld @Lily418 I agree, I'm in favor of this change as well. If the comments are addressed, we are good to go!

Lily418:
Thanks for feedback @edgarmueller @eneufeld 

The reason I wanted to show description when there is an error is to avoid this case where there appears an error but also an empty space above it.
Screenshot 2019-06-24 at 09 46 27

So I've updated the logic to use the top FormHelperText when there is an error but showDescription is false.

PR #1416, opened 2019-06-25
Add ajv-i18n

Lily418:
I would like to be able to localize the error messages from ajv.

I've added the ajv-i18n package and localize based on navigator.languages. 

I've asked a question on spectrum about how best to get state from other reducers as I'd like to take the list of languages from the i18n reducer.  


Lily418:
Closing in favour of  #1457

PR #1418, opened 2019-06-25
Export unwrapped material controls

Lily418:
Last week was talking to @eneufeld about building a custom control which reuses the MaterialTextControl. 

https://github.com/eclipsesource/jsonforms/pull/1413#issuecomment-504598408

This is a PR to export the control components not wrapped with `withJsonFormsControlProps` so they can be used in custom renderers.

### DIFF
--- a/packages/material/src/controls/index.ts
+++ b/packages/material/src/controls/index.ts
@@ -80,7 +80,7 @@ export const Unwrapped = {
   MaterialNumberControl: MaterialNumberControlUnwrapped,
   MaterialTextControl: MaterialTextControlUnwrapped,
   MaterialAnyOfStringOrEnumControl: MaterialAnyOfStringOrEnumControlUnwrapped
-}
+};
 
 export {
   MaterialBooleanControl,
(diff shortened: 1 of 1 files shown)


edgarmueller:
Thanks for the PR! Just a quick suggestion: could we have an object which is providing all the unwrapped components, so usage would be something like:

```js
import { Unwrapped } from '@jsonforms/material-renderers';

const MyRenderer = () => (
  <Unwrapped.MaterialTextControl ... />
)
```

edgarmueller:
@Lily418 Something went wrong here. Would you like to update once more?

Lily418:
Hi Edgar, I switched laptops and didn't update my email for commits yet. Going to update this PR with suggestions and re-write the email on the commits.

edgarmueller:
Great, thank you!

Lily418:
@edgarmueller  updated 🚡

edgarmueller:
It's still complaining about the formatting, but other than that this looks good

PR #1430, opened 2019-07-04
Add allowing additional inputProps

Lily418:
I would like to turn off spellcheck for a TextControl. I have my custom renderer 

```
import * as React from "react";
import { withJsonFormsControlProps } from "@jsonforms/react";

import { Unwrapped } from "@jsonforms/material-renderers";

export const SentenceCaseRender = props => {
  const sentenceCaseString = s => {
    return s.length > 0 ? s[0].toUpperCase() + s.slice(1).toLowerCase() : s;
  };

  const { handleChange } = props;
  return (
    <Unwrapped.MaterialTextControl
      {...props}
      inputProps={{ spellcheck: "false" }}
      handleChange={(path, targetValue) => {
        handleChange(path, sentenceCaseString(targetValue));
      }}
    />
  );
};

export default withJsonFormsControlProps(SentenceCaseRender);
```

This PR allows adding inputProps which get passed down to the MuiInputText

### DIFF
--- a/packages/material/src/mui-controls/MuiInputText.tsx
+++ b/packages/material/src/mui-controls/MuiInputText.tsx
@@ -36,19 +36,7 @@ interface MuiInputTextStatus {
 }
 
 interface MuiTextInputProps {
-  inputProps?: {
-    autocomplete?: string;
-    maxLength?: number;
-    minLength?: number;
-    pattern?: string | object;
-    placeholder?: string;
-    readonly?: boolean;
-    required?: boolean;
-    size?: number;
-    spellcheck?: boolean;
-    autocorrect?: 'on' | 'off';
-    mozactionhint?: 'go' | 'done' | 'next' | 'search' | 'send';
-  };
+  muiInputProps? : React.HTMLAttributes<HTMLInputElement>
 }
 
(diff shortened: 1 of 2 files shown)


packages/material/src/mui-controls/MuiInputText.tsx
```diff
@@ -51,18 +67,22 @@ export class MuiInputText extends React.Component<
       isValid,
       path,
       handleChange,
-      schema
+      schema,
+      inputProps
```
edgarmueller: Maybe we could rename `inputProps` to `muiInputProps` and keep `inputProps` in the body?

packages/material/src/mui-controls/MuiInputText.tsx
```diff
@@ -35,8 +35,24 @@ interface MuiInputTextStatus {
   showAdornment: boolean;
 }
 
+interface MuiTextInputProps {
+  inputProps?: {
```
edgarmueller: Do we have to re-declare all the props or is there a type exported by MUI that we could re-use?

Lily418:
Thanks for feedback @edgarmueller have updated 🛸

PR #1432, opened 2019-07-04
No alert when oneOf is empty object

Lily418:
When switching tabs of oneOf there can be a lot of alerts. When there is no data entered yet I think the alert should not display as there is no data to lose.


sdirix:
I'll take a look soon

sdirix:
I squash merged the change as the commits contained commits which added and removed logging information as well changes to the file header. If you prefer us merging in your commits without squashing, please clean your history to not contain redundant changes.

Thanks a lot again for contributing!

PR #1435, opened 2019-07-09
Add enum to structuralKeywords

Lily418:
There are cases where AJV error will be keyword enum.

Here is a such a case:
Schema:
```
{
  "title": "Injection",
  "type": "object",
  "properties": {
    "method": {
      "title": "Method",
      "type": "string",
      "enum": [
        "Injection"
      ],
      "default": "Injection"
    }
  },
  "required": [
    "method"
  ]
}
```

Data
```
{
  "method": "Infusion"
}
```

(We're using these enums to force the oneOf to select a specific tab when it could otherwise be ambigious) 

The error is 
```
  {
    keyword: 'enum',
    dataPath: '.method',
    schemaPath: '#/properties/method/enum',
    params: { allowedValues: [Array] },
    message: 'should be equal to one of the allowed values'
  }
```

### DIFF
--- a/packages/core/src/util/renderer.ts
+++ b/packages/core/src/util/renderer.ts
@@ -743,7 +743,7 @@ const mapStateToCombinatorRendererProps = (
   const ajv = state.jsonforms.core.ajv;
   const schema = resolvedSchema || rootSchema;
   const _schema = resolveSubSchemas(schema, rootSchema, keyword);
-  const structuralKeywords = ['required', 'additionalProperties', 'type'];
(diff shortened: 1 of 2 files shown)


sdirix:
Thanks for your PR. We'll take a look soon!

edgarmueller:
Thanks for the PR! Would you mind adding your example as a test case? That'd be great!

Lily418:
@edgarmueller thank you for the feedback, have updated 🐣

PR #1436, opened 2019-07-10
Use mapStateToEnumControlProps

Lily418:
mapStateToEnumControlProps was an unused function, this broke support for constants #1393


sdirix:
Thank you very much! I'll take a look soon

PR #1445, opened 2019-07-18
Add hideRequiredAsterisk option

Lily418:
I'd like the option to not have asterisk in labels of required properties. 

This can be useful if
- all properties are required so asterisk is not necessary
- If you want to build custom renderers using Unwrapped Controls which have some other way to show the required state

For me as almost all properties are required the asterisk is adding visual noise.


edgarmueller:
Think build is failing, could you have a look again?

Lily418:
Thanks @edgarmueller have updated

Issue #1271, opened 2019-02-22
Can we get an onChange prop for the JsonForms component?


edgarmueller:
**Describe the bug**
I'd be great if the the `JsonForms` could support an `onChange` prop that gets called whenever the UPDATE_DATA action is fired. This would enable easier integration within an existing redux app infrastructure as one would not need to take care of writing custom reducers anymore which sync the state between the app reducer and the jsonforms one.

**Used Setup (please complete the following information):**
 - Framework: react
 - RendererSet: n/A

sunilb0575:
Hi @edgarmueller this feature will be very much helpful. I would like to make sure that this fix will also address the following issue
[request for onDelete props when used schema of type='array'](https://github.com/eclipsesource/jsonforms/issues/1286)

PR #1446, opened 2019-07-18
Add onChange prop to JsonForms react component

Lily418:
This is for #1271

# Possible Examples
I intend to update the example for something which shows a use of this feature. 
## Updating Error
Changing the validation logic as discussed here

or localising the error as discussed

Both these cases require an additional action for updating errors

TODO:
- [ ] Consider API, should onChange parameter be JsonFormsCore? I think perhaps there's only use cases for errors and data.
- [ ] Fix issue where swapping between examples doesn't remove the onChange handler
- [ ] Refactor to avoid repeating `state.onChange && state.onChange(state)`
- [ ] Refactor to allow any InitActionOptions to be provided by example
- [ ] Update example to be sensible use case
- [ ] Testing

And because this is the longest Pull Request description I've written, here is a gif related to this feature
The more things change, the more they stay the same

### DIFF
--- a/packages/core/src/actions/index.ts
+++ b/packages/core/src/actions/index.ts
@@ -25,7 +25,7 @@
 import AJV from 'ajv';
 import RefParser from 'json-schema-ref-parser';
 import { RankedTester } from '../testers';
-import { JsonSchema, UISchemaElement } from '../';
+import { JsonSchema, UISchemaElement, JsonFormsCore } from '../';
 import { generateDefaultUISchema, generateJsonSchema } from '../generators';
 import { UISchemaTester } from '../reducers/uischemas';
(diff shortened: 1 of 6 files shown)


edgarmueller:
@Lily418 That gif is great :laughing:
I had a quick look over the changes you did. One thing I'm note so sure of is whether we want the `onChange` property to leak into redux, i.e. I'd not make it part of the reducer of any action.
Instead we could only let the `JsonForms` component handle the logic. The component would need to check via an `useEffect`/`componentDidUpdate` hook whether the root data changed and if so, call the `onChange` property. At least this is what I hand in mind. What do you think about that approach?

Lily418:
Sounds great, closing this PR in favour of  #1448

PR #1448, opened 2019-07-19
Add onChange props to React forms

Lily418:
This is for issue #1271 and an alternative to #1446

Opening a new PR as I've removed any changes to core.

# Possible Examples
I intend to update the example for something which shows a use of this feature. 
## Updating Error
Changing the validation logic as discussed here

or localising the error as discussed

Both these cases require an additional action for updating errors

### DIFF
--- a/packages/react/src/JsonForms.tsx
+++ b/packages/react/src/JsonForms.tsx
@@ -24,7 +24,7 @@
 */
 import isEqual from 'lodash/isEqual';
 import maxBy from 'lodash/maxBy';
-import React from 'react';
+import React, { useEffect } from 'react';
 import AJV from 'ajv';
 import RefParser from 'json-schema-ref-parser';
 import { UnknownRenderer } from './UnknownRenderer';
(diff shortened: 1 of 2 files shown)


Lily418:
Hi @edgarmueller I've followed your suggestion for onChange handler. 

However at the moment I have an issue, the onChange handler in the onChange example dispatches the UpdateErrors actions asynchronous. This means as you are typing the error flashes up each key press before it is removed. 

I'm unsure of the solution to this, I wonder if rather than dispatching an action the onChange handler should return the data and errors to render. This way the core state in unalterered by gives the user a hook to change what gets renderered.

edgarmueller:
Right, there seems to be an issue when updating the errors, but anyways: I'd suggest to split these changes into two PRs:
1. one which is only adding the `onChange` property to the `JsonForms` component as this seems to work fine already :+1:  (you can update this PR)
2. a second one would provide the action for updating the errors state. Also, I think we should not return an action from `onChange` but instead updating the errors on the client side, i.e. it's up to the consumer to trigger the action within his/her implementation of `onChange` (our current example structure might not properly support this use-case). We can deal with any issues we might face then.

Lily418:
@edgarmueller sounds good, have updated this PR

packages/react/src/JsonForms.tsx
```diff
@@ -48,6 +49,10 @@ interface JsonFormsRendererState {
     resolvedSchema: JsonSchema;
 }
 
+interface JsonFormsReactProps {
+  onChange?(state: {data: any,   errors?: ErrorObject[]}): void;
```
edgarmueller: We can refine this type via `Pick<JsonFormsCore, 'data' | 'errors'>`

Lily418:
@edgarmueller Thanks for the feedback, Pick type is very nice. have updated.

PR #1449, opened 2019-07-21
Use Memo and PureComponent where it was missing

eneufeld:
* Update functional components  to use React.memo
* Update components to extends React.PureComponent

### DIFF
--- a/packages/material/src/util/layout.tsx
+++ b/packages/material/src/util/layout.tsx
@@ -23,9 +23,6 @@
   THE SOFTWARE.
 */
 import isEmpty from 'lodash/isEmpty';
-import omit from 'lodash/omit';
-import isEqual from 'lodash/isEqual';
-import get from 'lodash/get';
 import React from 'react';
 import {
   JsonFormsRendererRegistryEntry,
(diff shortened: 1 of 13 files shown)


packages/material/src/util/layout.tsx
```diff
@@ -51,32 +54,56 @@ export const renderLayoutElements = (
   ));
 };
 
+const areEqual = (
```
edgarmueller: Why is this re-defined here?

eneufeld: yes, the types didn't match. But it would make sense to find a common type between MaterialLayoutType and JsonFormsPropTypes, or maybe extend the JsonFormsPropTypes, I can have a look at that

edgarmueller: Right, now that you mention it, I think this is also the case for the `ExpandPanelRenderer`. I'd be great if we could unify all of these.

edgarmueller: I'll merge this one then, let's create a follow-up. Could you take care of that?

edgarmueller: Also this needs to be rebased before I can merge it.

eneufeld: rebased and removed the unnecessary areEqual

eneufeld:
I rebased and removed the unnecessary areEqual methods

PR #1450, opened 2019-07-22
Add error updating action

Lily418:
This PR builds on #1448 to add an example use of validation which keeps track if a field has been touched as discussed here.

It also uses use useLayoutEffect rather than useEffect to call the onChange handler. This avoid visual glitches where the old error displays for a moment before re-render.


edgarmueller:
Looks good! Could you rebase once more? :pray:

Lily418:
Thanks @edgarmueller have rebased 🐙

edgarmueller:
Thanks @Lily418

PR #1452, opened 2019-07-23
Fix circular resolving issues

sdirix:
Avoid returning objects with circular references.

ToDo:
- [x] Avoid circular references
- [ ] Handle non-circular references in `mapStateToCombinatorRendererProps` in a different way than catching resolving errors
- [ ] Handle unresolvable subschemas in `JsonForms` component 
- [ ] Verify whether `cloneDeep` is necessary to avoid unwanted changes by `json-schema-ref-parser`


sdirix:
WIth the removal of `json-schema-ref-parser` for JSON Forms v3.0 this PR is no longer relevant.

PR #1466, opened 2019-08-01
Export JsonFormsReactProps interface

Lily418:
I'd like to create a function in my code like
```
const initalJsonFormsProps = (): JsonFormsInitStateProps & JsonFormsReactProps => {
// Some logic for creating inital props
}
```

but I noticed JsonFormsReactProps is not currently exported.


sdirix:
Thanks for the PR! We'll have a look at this for the next release

PR #1471, opened 2019-08-12
Improve layouting of buttons in ExpandPanelRenderer

shivgarg5676:
The expansion panel is breaking in the forms which are relatively smaller(tablet size) when showSortButtons is true. 
Screenshot 2019-08-12 at 10 53 47 AM

With this pr, the expansion panel will work fine for the smaller screen sizes(tablet)

Screenshot 2019-08-12 at 10 52 35 AM


iluuu1994:
I still have this issue when the page-width is small. 

image

The table has to be at least `948px` wide for this to work. When I make the left column `grid-md-9` and the right one `grid-md-3` it works. Would that be an acceptable change?

shivgarg5676:
May we specify sm={9} for left and sm={3} for right as 948px screen comes in sm category. Currently, this can be breaking for any screen with sm size( 768-991 pixels)

PR #1474, opened 2019-08-15
Memoize hasRefs

Lily418:
This is a performance optimisation for large schemas. 

I found with a large schema with a combination of 
- `MaterialArrayLayout`
- `MaterialOneOfRenderer`, 
- `MaterializedGroupLayoutRenderer` etc. 

There can be many instances of`JsonFormsDispatch`, but many will have identical schemas.

hasRefs is an expensive operation for a large schema so I found repeated calls were causing a slow down. 

I noticed a significant improvement using memoize

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -56,9 +56,11 @@
     ]
   },
   "dependencies": {
+    "@types/object-hash": "^1.3.0",
     "@types/react": "^16.4.14",
     "@types/react-redux": "^7.0.6",
-    "lodash": "^4.17.4"
+    "lodash": "^4.17.4",
+    "object-hash": "^1.3.1"
   },
   "peerDependencies": {
     "@jsonforms/core": "^2.3.0",
--- a/packages/react/src/JsonForms.tsx
+++ b/packages/react/src/JsonForms.tsx
@@ -59,12 +59,15 @@ export interface JsonFormsReactProps {
   onChange?(state: Pick<JsonFormsCore, 'data' | 'errors'>): void;
 }
 
-const hasRefs = memoize((schema: JsonSchema): boolean => {
-  if (schema !== undefined) {
-    return Object.keys(findRefs(schema)).length > 0;
-  }
-  return false;
-}, hash);
+const hasRefs = memoize(
+  (schema: JsonSchema): boolean => {
+    if (schema !== undefined) {
+      return Object.keys(findRefs(schema)).length > 0;
+    }
+    return false;
+  },
+  (schema: JsonSchema) => (schema ? hash(schema) : false)
+);
 
 export class ResolvedJsonFormsDispatchRenderer extends React.Component<
   JsonFormsProps,
--- a/packages/react/src/JsonForms.tsx
+++ b/packages/react/src/JsonForms.tsx
@@ -24,6 +24,7 @@
 */
 import isEqual from 'lodash/isEqual';
 import maxBy from 'lodash/maxBy';
+import memoize from 'lodash/memoize';
 import React, { useLayoutEffect } from 'react';
 import AJV from 'ajv';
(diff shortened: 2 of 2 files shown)


edgarmueller:
Thanks for the PR! I'm in favour of the change as we are aware of the performance issue and this seems like a good solutions, but unfortunately the `oneOf recursive` example breaks. I'm not sure why this is the case. Would you like to have another look at this?

Lily418:
Thanks @edgarmueller 

The issue was Map which [lodash memoize](https://github.com/lodash/lodash/blob/master/memoize.js#L62) uses seem to hate cyclic objects. I wasn't able to find any infomation about this limitation of Map but seemed to be the case with WeakMap too. 

The issue is trying to find an alternative hashing method or dropping memoization in this case in a way which doesn't cost as much or more performance than the benefit of the optimisation.

Have updated the PR to use [object-hash](https://www.npmjs.com/package/object-hash)

This is the first time in many years I find myself missing a Java feature [(hashcode)](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode())

edgarmueller:
Thanks for the update! Unfortunately tests are still failing though because of wrong typings, e.g. in `resolvers.tests.ts`.

sdirix:
Some comments about cyclic objects:
 - Ajv also doesn't support cyclic objects. Cyclic refs are fine but not when they are actually set in the object. So having cyclic schemas around will at the moment sometimes crash JSONForms.
 - json schema ref parser will produce cyclic objects when it's not turned off in the options.

Consequences for the memoize / hash use case:
 - If we say we just don't support cyclic schemas and make sure to configure json schema ref parser in a way that it doesn't create cyclic schemas anymore, we will also not hit that case in memoize and can therefore skip the hashing
 - If hashing on the other side has no real downside (assuming it is fast and won't introduce additional errors) I'm fine using it anyway

What do you think?

Lily418:
I think there's no extra that can be acheived with cyclic objects that cannot be done with cyclic refs. 

It would be good to avoid the additional dependancy on object-hash

I'll have a look at using ref parser config to avoid the issue with the recursive example.

Lily418:
I've hit a problem updating the json schema ref options.
https://apidevtools.org/json-schema-ref-parser/docs/options.html

Using option "ignore" the reference is still in the schema, the MaterialOneOfRenderer then creates this subschema.

```
{
  "name": {
    "type": "string"
  },
  "children": {
    "type": "array",
    "items": {
      "$ref": "#/definitions/fileOrFolder"
    }
  }
}
```

This is an invalid schema because it doesn't contain the definitions property.

From the [JSON Schema Docs on reuse](https://json-schema.org/understanding-json-schema/structuring.html#reuse)
"it is customary (but not required) to put it in the parent schema under a key called definitions"

As use of definitions key is not required we don't know which properties from the parent the subschema could reference. I'm not sure there is a clean way to ensure the subschema remain valid.

edgarmueller:
@Lily418 @sdirix  Yes, I think this is related to https://github.com/eclipsesource/jsonforms/pull/1452: We can end up with schemas which are not resolvable because referenced schemas get out of scope (as we are narrowing the schemas passed to sub renderers). This is a somewhat bigger issue and I don't think we'll be able to fix it for 2.3.1 release. I'm still in favor of the change though because it's a good quickfix for the performance issues without breaking anything: we could create an issue to remove the object-hash dependency as soon as the handling recursive references has been fixed. Do you agree?

Lily418:
This works for me and seems like a good approach.

sdirix:
I was not aware that additional stuff breaks when changing the behavior of the json schema ref parser. I therefore agree with both of you that using the object-hash is probably the best solution for now until we have a proper resolution strategy.

Issue #1464, opened 2019-07-29
Support @material-ui/core v4.3.0 remove dependency to @material-ui/lab


Schloemicher:
**Description**
Support @material-ui/core v4.3.0
Remove dependency @material-ui/lab

Slider is now in the core package 

**Describe for which setup you like to have the improvement**
Framework: [react]
RendererSet: [material-renderers]

Related Pull request: #1025

sdirix:
Thanks for creating the issue. We'd also like to remove `@material-ui/lab` (see #1028).

woodreamz:
Good idea! I use jsonforms and material-ui 4.3.0. As jsonforms uses material-ui ~4.0.1, I have two instances of material-ui which is bad :(

What do you think about make material-ui a peer dependency (and externalized by webpack)?

edgarmueller:
Sounds good! We'll take care of that.

edgarmueller:
The peer dependencies have been configured via d0c756ec181ffefb319abe10960a8ad41af82fca

PR #1481, opened 2019-08-20
Update Material UI

Lily418:
This is for #1464


sdirix:
Hi @Lily418, I already started working on this. Sadly I ran into the same issue as you do: The material tests start to fail. You can check my version here: https://github.com/sdirix/jsonforms/tree/update-material-ui

Lily418:
Ah amazing @sdirix , looks like WithWidth isn't rendering its children

sdirix:
@Lily418 I'll make sure to always open draft PRs in the future so we can avoid this situation

PR #1485, opened 2019-08-23
Add disabled option for tables - Fix #1389

### LINKED ISSUE #1389: React Table renderer can not be disabled.
Table renderer has does not handle enabled state. 
UI schema
``` 
{
  type: 'VerticalLayout',
  elements: [
    {
      type: 'Control',
      scope: '#/properties/abc'
    },
    {
      type: 'Control',
      scope: '#/properties/comments',
      options: {
        showSortButtons: true
      },
      rule: {
        effect: 'DISABLE',
        condition: {
          scope: '#/properties/abc',
          schema: { const: 'developer' }
        }
      }
    }
  ]
};
```
Schema
```
{
  type: 'object',
  properties: {
    abc: {
      type: 'string'
    },
    comments: {
      type: 'array',
      items: {
        type: 'object',
        properties: {
          date: {
            type: 'string',
            format: 'date'
          },
          message: {
            type: 'string',
            maxLength: 5
          }
        }
      }
    }
  }
};
```
Even if condition is true, table is still active. Because `enabled` prop has no effect on the table as the tablerenderer does not handle this at the moment

https://spectrum.chat/jsonforms/general/disable-rule-is-not-working~b256812a-d1a0-42fd-9781-6c395222dedd

TheZoker:


TheZoker:
Good catch, thanks!

PR #1495, opened 2019-09-03
Fix jumping cell width caused by delete button

TheZoker:


sdirix:
We now used a different approach: https://github.com/eclipsesource/jsonforms/pull/1513

PR #1502, opened 2019-09-20
Fix time field not being disabled

### LINKED ISSUE #1500: Disabling a control with format time has no effect
**Describe the bug**

A rule with effect DISABLE used on a time control has no effect. It's still possible to enter a time and use the up/down arrows etc.
Using DISABLE on date controls however work fine.

**To Reproduce**

Create a schema such as

```
type: object
properties:
  foo:
    type: boolean
  bar:
    type: string
    format: time
```

and a UI schema such as

```
type: VerticalLayout
elements:
  - type: Control
    scope: "#/properties/foo"
  - type: Control
    scope: "#/properties/bar"
    rule:
      effect: ENABLE
      condition:
        scope: "#/properties/foo"
        schema: 
          const: true
```

**Expected behavior**

When the time control `bar` is disabled by triggering the rule (foo: false), it should not be possible to edit the value in the time control.

**Screenshots**

 - Browser: Chrome
 - Version: 76.0.3809.132

Standard seed test app with only schema/ui schema replaced.

**Additional context**

TheZoker:
Fixes #1500

@sdirix

### DIFF
--- a/packages/material/src/controls/MaterialNativeControl.tsx
+++ b/packages/material/src/controls/MaterialNativeControl.tsx
@@ -48,6 +48,7 @@ export class MaterialNativeControl extends Control<ControlProps, ControlState> {
       label,
       schema,
       description,
+      enabled,
       visible,
       required,
       path,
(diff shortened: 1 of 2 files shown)


sdirix:
Please add a testcase :)

PR #1505, opened 2019-10-01
Async ref resolving

edgarmueller:
Some thing we need to clarify from the top of my head:
- manual ref resolution in context of arrays via `ResolveRef` (clear resolving design) + when to use `ScopedRenderer`
- caching of refs (this could be part of the store state)
- duplicate code (`refResovler` callback)
- usage of `resolveLocalSchema`
- generation of UI schema for object controls
- typings


sdirix:
I'll close this PR as it's open for a long time. There is a [copy of the contribution branch as part of the repository](https://github.com/eclipsesource/jsonforms/tree/fix/ref-handling-refactoring) which could prove useful should we go into this direction at some point in the future.

PR #1506, opened 2019-10-02
Add layout enable/disable feature

### LINKED ISSUE #1499: Disabling entire layout doesn't work
**Describe the bug**

If applying a rule to an entire layout, with ENABLE or DISABLE effect, nothing happens when triggering the rule. SHOW/HIDE works fine, but ENABLE/DISABLE shows no visible difference, and the controls within the layout are still selectable etc.

**To Reproduce**

Create a schema such as

```
type: object
properties:
  foo:
    type: boolean
  bar:
    type: string
  baz:
    type: string
```

And a UI schema such as

```
type: VerticalLayout
elements:
  - type: Control
    scope: "#/properties/foo"
  - type: VerticalLayout
    rule:
      effect: ENABLE
      condition:
        scope: "#/properties/foo"
        schema: 
          const: true
    elements:
      - type: Control
        scope: "#/properties/bar"
      - type: Control
        scope: "#/properties/baz"
```

**Expected behavior**

When `foo` is `false`, the vertical layout containing `bar` and `baz` should be disabled, reasonably meaning that each child in it should recursively be disabled.

**Screenshots**

 - Browser Chrome
 - Version 76.0.3809.132

Default seed test app with only schema and UI schema altered.

**Additional context**

TheZoker:
Fix #1499


packages/core/src/util/renderer.ts
```diff
@@ -375,7 +375,7 @@ export const mapStateToControlProps = (
   const visible = has(ownProps, 'visible')
     ? ownProps.visible
     : isVisible(uischema, rootData, ownProps.path);
-  const enabled = has(ownProps, 'enabled')
+  const enabled = (has(ownProps, 'enabled') && ownProps.enabled !== undefined)
```
sdirix: It seems to me that the code was written with the expectation that `has` also checks for `undefined`. As this is not the case we should use the `foo !== undefined` check everywhere instead of `has`. Please replace all usages of `has` with an undefined check in all files in a separate commit.

packages/core/src/util/renderer.ts
```diff
@@ -656,9 +656,7 @@ export const mapStateToLayoutProps = (
   const visible: boolean = has(ownProps, 'visible')
     ? ownProps.visible
     : isVisible(ownProps.uischema, rootData, ownProps.path);
-  const enabled: boolean = has(ownProps, 'enabled')
-    ? ownProps.enabled
-    : isEnabled(ownProps.uischema, rootData, ownProps.path);
+  const enabled: boolean = isEnabled(ownProps.uischema, rootData, ownProps.path);
```
sdirix: `isEnabled` only checks locally for enablement, e.g. whether the given ui schema has a rule which enables / disables it. This will lead to unexpected behavior in case of nested layouts where only the outer most layout is disabled via a rule. In general the most specific rule should win.

Issue #1504, opened 2019-09-24
uischema prop is required


hayk94:
**Describe the bug**
In the docs https://jsonforms.io/examples/gen-uischema it says 

> If you provide no UI schema to JSON Forms it'll generate one.

However I am using typescript with react and TS complains if I don't provide `uischema`  prop.

![image](https://user-images.githubusercontent.com/9801183/65507561-3cc30d80-dedf-11e9-9cc2-aa029e512307.png)

This is probably just type problems.

edgarmueller:
Hi, thanks for the report! The standalone component currently does not generate a UI schema by default, so we should either fix this or update the docs. We'll try to address it for the next release. You can get the behaviour by generating a UI schema yourself by calling `Generate.uiSchema` and passing in the schema.

sdirix:
Implemented with #1515

PR #1515, opened 2019-10-14
Generate uischema in JsonForms component

sdirix:
Implements: #1504

### DIFF
--- a/packages/react/src/JsonForms.tsx
+++ b/packages/react/src/JsonForms.tsx
@@ -40,7 +40,8 @@ import {
   OwnPropsOfJsonFormsRenderer,
   removeId,
   UISchemaElement,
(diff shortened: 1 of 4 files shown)


packages/react/src/JsonForms.tsx
```diff
@@ -217,6 +218,8 @@ export const JsonForms = (
     refParserOptions,
     onChange
   } = props;
+  const uischemaToUse =
+    uischema !== undefined ? uischema : Generate.uiSchema(schema);
```
edgarmueller: Minor possible improvement: I think we could make this check stricter by switching the logic and test for `typeof usichema === 'object'`.

sdirix: done

PR #1522, opened 2019-11-18
Allow configuring array child label

iluuu1994:


packages/material/test/renderers/MaterialArrayLayout.test.tsx
```diff
@@ -431,4 +441,54 @@ describe('Material array layout', () => {
...
+          <JsonFormsReduxContext>
+            <MaterialArrayLayout
+                schema={schema}
+                uischema={uischemaWithSortOption}
+            />
+          </JsonFormsReduxContext>
+        </Provider>
+    );
+    const childLabel1 = wrapper
+        .find('.MuiGrid-root.MuiGrid-item.MuiGrid-grid-xs-10.MuiGrid-grid-md-11')
```
iluuu1994: Should we add a class to that element so it can be selected easier?

sdirix:
Thank you very much for the contribution. Much appreciated! We'll take a closer look at it soon.

packages/material/test/renderers/MaterialArrayLayout.test.tsx
sdirix: This question made me realize that we have no proper semantics defined for this label. I now introduced `aria-labelledby` to add respective semantics.

iluuu1994:
Looks good, thanks!

sdirix:
Thanks again for the contribution!

iluuu1994:
Cool, thank you @sdirix! 😊

PR #1531, opened 2019-12-13
Remove redux from angular

eneufeld:
* Introduce `JSONFormsAngularService` which replaces redux
* Adapt renderers
* Update tests


ruimarques:
Oh yes this would be awesome!

Anyway I can help?

To provide some context I am considering to add jsonforms to an angular application product but I will need:

- uncoupling from redux as we don't use it and I don't want to carry that dependency (we use other state mgmt lib)
- custom renderer which we will develop for our needs, we have our own components we want to use. (jsonforms already supports this I believe)

sdirix:
@ruimarques Thanks for your interest! Rendering customization is a staple of JSON Forms on which the whole architecture is based. Actually all our provided renderers use the very same mechanisms we expect users of JSON Forms to use too.

If you'd like to help you could test this PR in your setup and see if we missed something when removing redux or you'd expect a different replacement interface (or implementation). The more eyes look at this, the better.

packages/angular-material/test/autocomplete-control.spec.ts
```diff
@@ -99,104 +101,95 @@ describe('Autocomplete control Base Tests', () => {
   });
 
   it('should render', fakeAsync(() => {
-    const mockSubStore = setupMockStore(fixture, { uischema, schema, data });
-    mockSubStore.complete();
-    fixture.detectChanges();
+    setupMockStore(fixture, { uischema, schema, data });
+    getJsonFormsService(component).updateCore(
```
sdirix: Why do we need to do this? Doesn't `setupMockStore` not already setup the `JsonFormsService`?

packages/core/src/actions/index.ts
```diff
@@ -56,6 +56,14 @@ export const SET_LOCALIZED_UISCHEMAS: 'jsonforms/SET_LOCALIZED_UISCHEMAS' =
 export const ADD_DEFAULT_DATA: 'jsonforms/ADD_DEFAULT_DATA' = `jsonforms/ADD_DEFAULT_DATA`;
 export const REMOVE_DEFAULT_DATA: 'jsonforms/REMOVE_DEFAULT_DATA' = `jsonforms/REMOVE_DEFAULT_DATA`;
 
+export type ValidCoreActions =
```
sdirix: Shouldn't we just name them `CoreActions`?
```suggestion
export type CoreActions =
```

packages/core/src/actions/index.ts
```diff
@@ -200,6 +208,10 @@ export const setConfig = (config: any) => (dispatch: Dispatch<AnyAction>) => {
   });
 };
 
+export type ValidUISchemaReducerActions =
```
sdirix:
```suggestion
export type UISchemaActions =
```

packages/core/src/actions/index.ts
```diff
@@ -231,6 +243,11 @@ export const unregisterUISchema = (
   };
 };
 
+export type ValidLocaleActions =
```
sdirix:
```suggestion
export type LocaleActions =
```

packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -241,10 +242,9 @@ export class MasterListComponent extends JsonFormsArrayControl {
     this.removeItems(this.propsPath, [item])();
   }
 
-  protected mapToProps(state: JsonFormsState): ArrayControlProps {
+  protected mapToProps(state: JsonFormsState): StatePropsOfArrayControl {
```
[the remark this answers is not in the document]
sdirix: If we change this here then we should change all occurrences of this pattern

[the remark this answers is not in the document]
eneufeld: we can't as this is a method that we need to override

PR #1552, opened 2020-02-27
Extract tree renderer and webcomponent packages

sdirix:
Both the `material-tree-renderer` as well as the `webcomponent` packages will no longer be actively maintained. Therefore both are extracted to their own repositories for now:

- https://github.com/eclipsesource/jsonforms-react-material-tree-renderer
- https://github.com/eclipsesource/jsonforms-react-webcomponents


packages/material/package.json
```diff
@@ -81,11 +81,11 @@
   "devDependencies": {
     "@jsonforms/core": "^2.3.2",
     "@jsonforms/react": "^2.3.2",
-    "@jsonforms/test": "^2.3.2",
```
AlexandraBuzila: Are these changes meeded for the removal?

sdirix: Yes as we are also removing the test package

PR #1559, opened 2020-03-12
Fix performance of rendering lots of fields

nowells:
I introduced a fix in https://github.com/eclipsesource/jsonforms/pull/1214 to fix performance of rendering lots of fields. Since that time the code has shifted and that fix has been lost. The new manifestation appears to stems from
`JsonFormsDispatchRenderer` calling `hasRefs` while `ResolvedJsonFormsDispatchRenderer` does not, as only the first JsonForm needs to handle processing resolving refs.

Right now we are doing a lot of wasted `object-hash` comparisons in the `hasRefs` memoized check due to using the wrong component in our `material` and `vanilla` components.

### DIFF
--- a/packages/material/src/additional/MaterialListWithDetailRenderer.tsx
+++ b/packages/material/src/additional/MaterialListWithDetailRenderer.tsx
@@ -36,7 +36,7 @@ import {
   uiTypeIs
 } from '@jsonforms/core';
 import {
-  JsonFormsDispatch,
+  ResolvedJsonFormsDispatch,
   withJsonFormsArrayLayoutProps
 } from '@jsonforms/react';
 import { Grid, Hidden, List, Typography } from '@material-ui/core';
@@ -128,7 +128,7 @@ export const MaterialListWithDetailRenderer = ({
         </Grid>
         <Grid item xs>
           {selectedIndex !== undefined ? (
-            <JsonFormsDispatch
+            <ResolvedJsonFormsDispatch
               renderers={renderers}
               cells={cells}
               visible={visible}
--- a/packages/material/src/complex/CombinatorProperties.tsx
+++ b/packages/material/src/complex/CombinatorProperties.tsx
@@ -25,7 +25,7 @@
 import React from 'react';
 import _ from 'lodash';
 import { Generate, JsonSchema, Layout, UISchemaElement } from '@jsonforms/core';
-import { JsonFormsDispatch } from '@jsonforms/react';
+import { ResolvedJsonFormsDispatch } from '@jsonforms/react';
 
 interface CombinatorPropertiesProps {
   schema: JsonSchema;
@@ -51,7 +51,7 @@ export class CombinatorProperties extends React.Component<CombinatorPropertiesPr
 
     if (isLayoutWithElements) {
       return (
-        <JsonFormsDispatch
+        <ResolvedJsonFormsDispatch
           schema={otherProps}
           path={path}
           uischema={foundUISchema}
--- a/packages/material/src/complex/MaterialAllOfRenderer.tsx
+++ b/packages/material/src/complex/MaterialAllOfRenderer.tsx
@@ -35,7 +35,7 @@ import {
   resolveSubSchemas,
   StatePropsOfCombinator
 } from '@jsonforms/core';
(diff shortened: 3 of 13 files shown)


sdirix:
Thanks @nowells, I'm sorry that your fix got lost!

We :heart: contributions, so we'll take a look soon!

nowells:
@sdirix question for you, with this change, is is now safe for me to add `schema` to the `omit` list in the `areEqual` comparison in https://github.com/eclipsesource/jsonforms/blob/master/packages/react/src/JsonFormsContext.tsx#L362-L363 because now any form should have a fully resolved schema where as before they still potentially had resolving schemas? That is also a performance hit for the `React.memo` when the schema is sufficiently large (which we have forms with hundreds of fields in the schema). I can tackle trying to improve that performance in a follow on PR, but wanted to get guidance on the rationale why `uischema` is already excluded and `schema` is not.

nowells:
Side question, would it be helpful for me to add (in a different PR) a performance test suite to help catch performance regressions automatically in the future? If that type of PR would be interesting to you let me know, and I or someone on my team will likely be happy to contribute something like that. Thanks for the great library, and always being so responsive and welcoming to contributions!

nowells:
I have locally tested this, and in a form with 444 inputs that are wrapped in `MaterialLayoutRenderer`, the initial render time goes from Chrome prompting with a Wait or Kill script response, to < 2s. There is more we can do to improve, but this is a big deal.

sdirix:
> @sdirix question for you, with this change, is is now safe for me to add schema to the omit list in the areEqual comparison in https://github.com/eclipsesource/jsonforms/blob/1f3f92a/packages/react/src/JsonFormsContext.tsx#L362-L363 because now any form should have a fully resolved schema where as before they still potentially had resolving schemas? That is also a performance hit for the React.memo when the schema is sufficiently large (which we have forms with hundreds of fields in the schema). I can tackle trying to improve that performance in a follow on PR, but wanted to get guidance on the rationale why uischema is already excluded and schema is not.

It's not `uischema` but `uischemas` which are excluded ;) `uischemas` is for a very niche feature where you can register multiple uischemas with the store which can then be used as a detail ui schema in some renderers. This feature might get removed at some point. Case in point the non-redux version of JSON Forms doesn't support it yet and nobody seems to miss it.

We'll reevaluate with your fix whether we can add additional items to the omit list :+1: 

> Side question, would it be helpful for me to add (in a different PR) a performance test suite to help catch performance regressions automatically in the future? If that type of PR would be interesting to you let me know, and I or someone on my team will likely be happy to contribute something like that. Thanks for the great library, and always being so responsive and welcoming to contributions!

This would be amazing!

> I have locally tested this, and in a form with 444 inputs that are wrapped in MaterialLayoutRenderer, the initial render time goes from Chrome prompting with a Wait or Kill script response, to < 2s. There is more we can do to improve, but this is a big deal.

Awesome :tada:

nowells:
Thanks for the answers. Can the `uischema` and `schema` both be removed from comparison? Or do that memo comparison once at the top level, and then pass down a hash to all children to memoize on? Because having every component do the same expensive comparison for those two big props could absolutely be done once at the top level. If this makes sense, then let me know if you want me to tackle something like that in a future PR.

Thanks for the context on the `uischemas`, good call, totally missed that trailing `s`.

I'll work on getting someone (or myself) to draft up a performance test.

nowells:
Let me know if there is anything I can to to support or change this PR.

sdirix:
I tested this change and I couldn't find any regression :tada: 

We also looked again at the `areEqual` comparison. We can't remove schema and uischema in general as there are use cases where they could change. However the deep comparison we do at the moment is very likely counterproductive and especially so for large schemas. A shallow compare is more than enough. I don't think we even have a use case where we change the schema or uischema object but retain its content. Thanks a lot for pointing us as this problem!

nowells:
Thanks @sdirix!!!

sdirix:
@nowells fyi, we released a `next`-stream version `2.4.0-alpha.2` which contains these changes. So if you'd like you can already consume them. Thanks again :smile:

nowells:
thanks @sdirix!!! I just linked it up and it works like a dream. I still owe you a performance test contribution, might take a bit, but I'll get it up here for you. Thanks for the support!

PR #1563, opened 2020-03-20
Fix crash in MaterialTableControl

sdirix:
When the item object schema doesn't specify properties the control
crashed. We now check the type of the properties before (not)
accessing them.

### DIFF
--- a/packages/material/src/complex/MaterialTableControl.tsx
+++ b/packages/material/src/complex/MaterialTableControl.tsx
@@ -116,7 +116,7 @@ const generateCells = (
 };
 
 const getValidColumnProps = (scopedSchema: JsonSchema) => {
-  if (scopedSchema.type === 'object') {
+  if (scopedSchema.type === 'object' && typeof scopedSchema.properties === 'object') {
     return Object.keys(scopedSchema.properties).filter(
       prop => scopedSchema.properties[prop].type !== 'array'
     );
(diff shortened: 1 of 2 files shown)


packages/material/src/complex/MaterialTableControl.tsx
```diff
@@ -116,7 +116,7 @@ const generateCells = (
 };
 
 const getValidColumnProps = (scopedSchema: JsonSchema) => {
-  if (scopedSchema.type === 'object') {
+  if (scopedSchema.type === 'object' && typeof scopedSchema.properties === 'object') {
```
eneufeld: shouldn't a check for not undefined be enough?

packages/material/test/renderers/MaterialArrayControl.test.tsx
```diff
@@ -167,6 +167,41 @@ describe('Material array control', () => {
...
+  it('should render even without properties', () => {
+    const store = initJsonFormsStore();
+    // re-init
+    const data: any = { test: [] };
+    const schema: JsonSchema = {
+      type: 'object',
+      properties: {
+        test: {
+          type: 'array',
+          items: { type: 'object' }
```
eneufeld: shouldn't the items object contain a `properties` property instead of `type:'object'` ?

sdirix: The crash happens exactly in this case: Items are defined as `object` without defining further properties. If there was a `properties`key, the crash didn't happen.

packages/material/src/complex/MaterialTableControl.tsx
sdirix: We're using `Object.keys(scopedSchema.properties)` later which expects an object-type parameter. Depending on the JS version it will either throw a `TypeError` or does a meaningless coercion on non-object types. So we really should check for `object`.

PR #1569, opened 2020-04-01
Angular enhancements

zRosenthal:
Expose state as an observable to give more flexibility (leverage pipes)

Remove dependency on BrowserAnimationsModule as it should be treated the same as the BrowserModule and be required in the root of the angular app.

### DIFF
--- a/packages/angular/src/jsonforms.service.ts
+++ b/packages/angular/src/jsonforms.service.ts
@@ -1,10 +1,21 @@
-import { JsonFormsState, RankedTester, UISchemaElement, JsonFormsSubStates, coreReducer, JsonFormsRendererRegistryEntry } from "@jsonforms/core";
-import { BehaviorSubject, Subscription, PartialObserver } from "rxjs";
-import { JsonFormsBaseRenderer } from "./base.renderer";
-import { CoreActions, i18nReducer, LocaleActions, UISchemaActions, uischemaRegistryReducer } from "@jsonforms/core";
+import {
+  CoreActions,
+  coreReducer,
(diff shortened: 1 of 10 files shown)


sdirix:
Thanks for the contribution! :heart: We'll take a look soon!

packages/angular/src/jsonforms.service.ts
```diff
@@ -19,11 +30,11 @@ export class JsonFormsAngularService {
         this.state = new BehaviorSubject({ jsonforms: this._state });
     }
 
-    subscribe(observer: PartialObserver<JsonFormsState>): Subscription {
+    get $state(): Observable<JsonFormsState> {
```
eneufeld, line 33: I suggest to keep the subscribe method but mark it as deprecated.
@sdirix what do you think?

sdirix: Besides some alpha releases this was not yet made public so I would be fine with changing this now without deprecation.

eneufeld: agreed,

zRosenthal:
@sdirix @eneufeld Thanks for taking a look at this!
Any idea when this might get merged/released?

sdirix:
We'll release this (as an alpha release) at latest at some point next week. The regular release probably comes at some point next month.

sdirix:
Thanks again for the PR :heart:

sdirix:
@zRosenthal We managed to get the alpha release out today. `2.4.0-alpha.3` is ready to be consumed :+1:

PR #1572, opened 2020-04-20
[Angular] Add hint to show description on controls

eneufeld:
Add hint to angular controls. This way the hint is always visible.


eneufeld:
I can adopt the angular code base to behave the same as the react renderers.

PR #1573, opened 2020-04-21
[Angular] Fix rule visibility for enum renderer

eneufeld:
The autocomplete renderer which is also used for enum,
is now checking the hidden field so that rules are fully supported.


sdirix:
Looking over the codebase it seems that the object and table renderer also don't have visbility support. Can you confirm whether that's the case?

eneufeld:
> Looking over the codebase it seems that the object and table renderer also don't have visbility support. Can you confirm whether that's the case?

Yes you are right, those two are missing the rule support completely

sdirix:
Created an issue for that: https://github.com/eclipsesource/jsonforms/issues/1578

Issue #1454, opened 2019-07-24
ability to define uischema scope by $id


gilbdev:
**Is your feature request related to a problem? Please describe.**

currently i have to reference :
`
{
        "type": "Control",
        "scope": "#/definitions/AnyElement/anyOf/0/allOf/1"
}
`
this is not really safe in case i add more sections in the anyOf or allOf, i will ahve to verify my uischema each time.

**Describe the solution you'd like**
i would like to be able to define this way:
`
{
        "type": "Control",
        "scope": "#myUniqueId"
}
`

where #myUniqueId is the "$id" json field of the element i want.

bonus:
being able to go further down the scope:
"scope": "#myUniqueId/properties/myproperty"

Lily418:
Hi, are you able to update the JSON Schema you are using? 
An alternative to this deeply nested reference would be to use `$ref` this way your UISchema can reference something that's used in any place in your schema. 

There's some [documentation about the use of $ref](https://json-schema.org/understanding-json-schema/structuring.html)

buryndin:
@Lily418 @sdirix 
Are there any ways to avoid using such scopes in uischema now?
(like `"#/definitions/AnyElement/anyOf/0/allOf/1"`)

sdirix:
In general yes as anything in JSON Forms is customizable, however there is no specific support.

We use the scope in two ways:
 - Resolving the corresponding part in the JSON Schema, currently using the [JSON Schema ref parser](https://github.com/APIDevTools/json-schema-ref-parser). So for this part you can use any functionality this library supports.
 - Resolving the path within the data object to retrieve / modify. Currently this is manually implemented, so whatever scope you give will be transformed and used (e.g. removing `properties` out of the path etc.)

So you can just specify different scopes when they work with these two use cases.

The other approach would be to handle these two use cases in your own code instead of relying on JSON Forms. For this you need to register a whole renderer set (or at least all renderers which shall handle the custom scopes) and do the resolving on your own. The code in JSON Forms lives in the [`mapStateToControlProps` function](https://github.com/eclipsesource/jsonforms/blob/5b751d2c0aa44475b8d4d85630a0606cbbedf215/packages/core/src/util/renderer.ts#L379-L429) which is called by `withJsonFormsControlProps`. So you can "just" wrap all your or existing renderers with a customized `withJsonFormsControlProps` which can handle your custom scopes. This way any scope can be supported.

buryndin:
Stefan,

[...]

PR #1579, opened 2020-04-29
Using $ref in scope in uischema (POC)

buryndin:


sdirix:
I'll close this for now until we come up with a better solution as discussed in the [associated issue](https://github.com/eclipsesource/jsonforms/issues/1454).

PR #1581, opened 2020-05-01
attempt at adding value and label support for select control

chavalipraveen:
As per the discussion in the following thread, trying to add support for having different key and label for select options (MaterialEnumControl).

https://spectrum.chat/jsonforms/general/for-enums-i-need-to-be-able-to-put-labels-for-the-enum-values~5b8b5153-033f-43a9-8f5e-bc27c37a2781

I am still working on adding the test for this. MaterialEnumControl doesn't seem to have existing tests (please correct me if I am wrong). So, I am writing tests for the base control and also the modification I am proposing.

### DIFF
--- a/packages/material/src/mui-controls/MuiSelect.tsx
+++ b/packages/material/src/mui-controls/MuiSelect.tsx
@@ -30,6 +30,12 @@ import { MenuItem } from '@material-ui/core';
 import { areEqual } from '@jsonforms/react';
 import merge from 'lodash/merge';
 
+interface Option {
+  key: string;
+  value: string;
+  label: string;
+}
+
 export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
   const {
     data,
@@ -43,6 +49,17 @@ export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
     config
   } = props;
   const appliedUiSchemaOptions = merge({}, config, uischema.options);
+  const delimiter: string = appliedUiSchemaOptions.keyValueDelimiter;
+  const keyValueLabels: Option[] = (delimiter === undefined) ?
+  options.map(option => {
+    return { key: option, value: option, label: option };
+  }) :
+  options.map(option => {
+    return option.split(delimiter, 2).map((parts: void[]) => {
+      return { key: parts[0], value: option, label: parts[1] };
+    }
+    );
+  });
 
   return (
     <Select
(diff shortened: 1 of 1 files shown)


packages/material/src/mui-controls/MuiSelect.tsx
```diff
@@ -55,9 +72,9 @@ export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
       fullWidth={true}
     >
       {[<MenuItem value='' key={'empty'} />].concat(
-        options.map(optionValue => (
-          <MenuItem value={optionValue} key={optionValue}>
-            {optionValue}
+        keyValueLabels.map(option => (
+          <MenuItem value={option.value} key={option.key}>
```
sdirix, line 76: Keys are uses to [uniquely identify](https://reactjs.org/docs/lists-and-keys.html#keys) elements in React. Therefore there is no need to customize it. Just using the value is good enough, as the value [should be unique](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.2).

packages/material/src/mui-controls/MuiSelect.tsx
```diff
@@ -43,6 +49,17 @@ export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
...
   } = props;
   const appliedUiSchemaOptions = merge({}, config, uischema.options);
+  const delimiter: string = appliedUiSchemaOptions.keyValueDelimiter;
+  const keyValueLabels: Option[] = (delimiter === undefined) ?
+  options.map(option => {
+    return { key: option, value: option, label: option };
+  }) :
+  options.map(option => {
+    return option.split(delimiter, 2).map((parts: void[]) => {
+      return { key: parts[0], value: option, label: parts[1] };
```
sdirix, line 59: This needs some sort of sanity check whether there are actually two parts.

sdirix, line 59: I usually avoid function bodies which immediately start with a `return` as this just adds visual clutter.

chavalipraveen:
Hi @sdirix 

Thanks for the detailed review comments. I will explore both approaches you mentioned and get back. 

- using "enumLabels" in the uischema
- using a custom renderer using oneOf and const

Regards,
Praveen.

packages/material/src/mui-controls/MuiSelect.tsx
chavalipraveen: Just using value without the label fails at ajv validation as it expects the value to be one of the allowed enums.

sdirix: I'm not sure we're talking about the same thing. What I meant is that the separate `option.key` is not needed, e.g.
```suggestion
          <MenuItem value={option.value} key={option.value}>
```
would also be fine.

sdirix:
The oneOf-const approach is now implemented with https://github.com/eclipsesource/jsonforms/pull/1591. Please let me know if you have further questions or feedback ;)

Issue #1576, opened 2020-04-24
Hide Angular renderer hints by default


sdirix:
Currently the Angular renderers behave like the React renderers with the `showUnfocusedDescription:true` option set, i.e. rendering the description of an input as a hint at all times (except for when there is a validation error). It would be useful to extend the Angular renderers to also support `showUnfocusedDescription:false` and to set this as the default.

PR #1584, opened 2020-05-12
Hide Angular renderer hints by default #1576

AlexandraBuzila:
- set and update config
- use default config when not set
- show/hide renderer hints based on config

### DIFF
--- a/packages/angular/src/abstract-control.ts
+++ b/packages/angular/src/abstract-control.ts
@@ -86,6 +86,14 @@ export abstract class JsonFormsAbstractControl<
     this.triggerValidation();
   }
 
+  shouldShowUnfocusedDescription(): boolean {
+    const config = this.jsonFormsService.getState().jsonforms.config;
+    if (config) {
+      return config.showUnfocusedDescription;
+    }
+    return false;
+  };
+
   ngOnInit() {
     this.jsonFormsService.$state.subscribe({
       next: (state: JsonFormsState) => {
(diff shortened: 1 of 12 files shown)


packages/angular/src/abstract-control.ts
```diff
@@ -86,6 +86,14 @@ export abstract class JsonFormsAbstractControl<
     this.triggerValidation();
   }
 
+  shouldShowUnfocusedDescription(): boolean {
+    const config = this.jsonFormsService.getState().jsonforms.config;
```
sdirix, line 90: The `config` is just an easy way to set default options for ui schema elements. If the same option exists on a ui schema element it has a higher precedence then the one from the config.

sdirix:
I'll take care of the fix.

AlexandraBuzila:
I rebased the changes the fix the conflicts with master.

packages/angular/src/abstract-control.ts
```diff
@@ -86,6 +87,15 @@ export abstract class JsonFormsAbstractControl<
...
+    const appliedUiSchemaOptions = merge({}, config, this.uischema.options);
+    if (appliedUiSchemaOptions) {
+      return !!appliedUiSchemaOptions.showUnfocusedDescription;
+    }
+    return false;
```
sdirix, line 96: Here we hide away the default value for `showUnfocusedDescription` in this method. I would prefer when the service would return the default `config` when the user did not set their own. This way we also don't need the `undefined` check and can just immediately return the value of the config.

AlexandraBuzila: The config here already contains the default values for undefined properties (the default values are set by the configReducer https://github.com/eclipsesource/jsonforms/pull/1584/files#diff-42788974ca6b5be621ff5d539f3155e8R83). I guess I added the undefined checks out of habit, would you prefer if I removed them?

sdirix: Maybe I'm missing something but isn't the config `undefined` by default and will only be set when the user actually calls `updateConfig`? Ideally it should always set to its default state when the user did not call `updateConfig`.

AlexandraBuzila: You're right, I see your point now. The last comit fixes this.

PR #1585, opened 2020-05-25
Remove label cutoff

sdirix:
## Don't cut off labels in MaterialInputControl 
Adjust the custom classes for MaterialInputControl's InputLabel to not
cut off characters.

We use a special style to increase the InputLabel's length when
shrinked as well as to avoid line breaks by rendering ellipses instead.
With an update of material ui the custom styles collided with the
default ones and lead to characters like 'ypq' etc. to be
cut off. This cutoff is now avoided. Also the width of the InputLabel is
adjusted to respect the input adornment when not shrinked.

## Hide 'clear' adornment when data is undefined 
Adds a check in MuiInputText to hide the clear button when its data
is already undefined.

### DIFF
--- a/packages/material/src/controls/MaterialInputControl.tsx
+++ b/packages/material/src/controls/MaterialInputControl.tsx
@@ -32,13 +32,27 @@ import {
 } from '@jsonforms/core';
 import { Control } from '@jsonforms/react';
 
-import { Hidden, InputLabel } from '@material-ui/core';
+import { Hidden, InputLabel, withStyles } from '@material-ui/core';
 import { FormControl, FormHelperText } from '@material-ui/core';
 import merge from 'lodash/merge';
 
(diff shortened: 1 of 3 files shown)


packages/material/src/controls/MaterialInputControl.tsx
```diff
@@ -1,3 +1,14 @@
+import {
```
eneufeld, line 1: The import above the license header looks wrong

packages/material/src/mui-controls/MuiInputText.tsx
```diff
@@ -1,3 +1,9 @@
+import { CellProps, WithClassname } from '@jsonforms/core';
```
eneufeld, line 1: import section looks to be at the wrong place again

PR #1591, opened 2020-06-17
Add support for enum using OneOf constructs

eneufeld:
- Add an example for enums
- Add a tester for oneOf enums
- Add a control to render oneOf enums based on MUISelect

### DIFF
--- a/packages/core/src/util/cell.ts
+++ b/packages/core/src/util/cell.ts
@@ -162,7 +162,7 @@ export const defaultMapStateToEnumCellProps = (
 ): StatePropsOfEnumCell => {
   const props: StatePropsOfCell = mapStateToCellProps(state, ownProps);
   const options: EnumOption[] =
-    ownProps.options !== undefined ? ownProps.options :
+    ownProps.options ||
     props.schema.enum?.map(enumToEnumOptionMapper) ||
     props.schema.const && [enumToEnumOptionMapper(props.schema.const)];
   return {
--- a/packages/core/src/util/cell.ts
+++ b/packages/core/src/util/cell.ts
@@ -34,7 +34,12 @@ import {
   Resolve,
   StatePropsOfScopedRenderer
 } from '.';
-import { DispatchPropsOfControl, mapDispatchToControlProps } from './renderer';
+import {
+  DispatchPropsOfControl,
+  EnumOption,
+  enumToEnumOptionMapper,
+  mapDispatchToControlProps,
+} from './renderer';
 import { JsonFormsState } from '../store';
 import { AnyAction, Dispatch } from 'redux';
 import { JsonFormsCellRendererRegistryEntry } from '../reducers/cells';
@@ -156,8 +161,8 @@ export const defaultMapStateToEnumCellProps = (
   ownProps: OwnPropsOfEnumCell
 ): StatePropsOfEnumCell => {
   const props: StatePropsOfCell = mapStateToCellProps(state, ownProps);
-  const options =
-    ownProps.options !== undefined ? ownProps.options : props.schema.enum;
+  const options: EnumOption[] =
+    ownProps.options !== undefined ? ownProps.options : props.schema.enum.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
   return {
     ...props,
     options
--- a/packages/core/src/util/cell.ts
+++ b/packages/core/src/util/cell.ts
@@ -163,7 +163,8 @@ export const defaultMapStateToEnumCellProps = (
   const props: StatePropsOfCell = mapStateToCellProps(state, ownProps);
   const options: EnumOption[] =
     ownProps.options !== undefined ? ownProps.options :
-    props.schema.enum?.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
+    props.schema.enum?.map(enumToEnumOptionMapper) ||
+    props.schema.const && [enumToEnumOptionMapper(props.schema.const)];
   return {
     ...props,
     options
--- a/packages/core/src/util/cell.ts
+++ b/packages/core/src/util/cell.ts
@@ -162,7 +162,8 @@ export const defaultMapStateToEnumCellProps = (
 ): StatePropsOfEnumCell => {
   const props: StatePropsOfCell = mapStateToCellProps(state, ownProps);
   const options: EnumOption[] =
-    ownProps.options !== undefined ? ownProps.options : props.schema.enum.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
+    ownProps.options !== undefined ? ownProps.options :
+    props.schema.enum?.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
   return {
     ...props,
     options
--- a/packages/core/src/util/renderer.ts
+++ b/packages/core/src/util/renderer.ts
@@ -172,6 +172,14 @@ export interface WithClassname {
   className?: string;
 }
 
+export interface EnumOption {
+  label: string;
+  description: string;
+  value: any;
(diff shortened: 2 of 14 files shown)


packages/core/src/util/cell.ts
```diff
@@ -156,8 +161,8 @@ export const defaultMapStateToEnumCellProps = (
   ownProps: OwnPropsOfEnumCell
 ): StatePropsOfEnumCell => {
   const props: StatePropsOfCell = mapStateToCellProps(state, ownProps);
-  const options =
-    ownProps.options !== undefined ? ownProps.options : props.schema.enum;
+  const options: EnumOption[] =
+    ownProps.options !== undefined ? ownProps.options : props.schema.enum.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
```
sdirix, line 165: This will cause runtime errors when `props.schema.enum` is undefined.

packages/core/src/util/renderer.ts
```diff
@@ -209,7 +217,7 @@ export interface OwnPropsOfControl extends OwnPropsOfRenderer {
 }
 
 export interface OwnPropsOfEnum {
-  options?: any[];
+  options?: EnumOption[];
```
sdirix, line 220: This change fails some unit tests as they simply assign strings. They should be adapted.

packages/core/src/util/renderer.ts
```diff
@@ -172,6 +172,14 @@ export interface WithClassname {
   className?: string;
 }
 
+export interface EnumOption {
+  label: string;
+  description: string;
```
sdirix, line 177: Not sure what `description` is intended for. If we want to keep it we should probably make it optional.

packages/core/src/util/renderer.ts
```diff
@@ -172,6 +172,14 @@ export interface WithClassname {
...
+  value: any;
+}
+
+export const enumToEnumOptionMapper = (e: any): EnumOption => ({ label: e, description: e, value: e });
```
sdirix, line 181: we should `stringify` `e` for label and description as we typed them as `string`

packages/core/src/util/renderer.ts
```diff
@@ -454,10 +462,31 @@ export const mapStateToEnumControlProps = (
   ownProps: OwnPropsOfControl & OwnPropsOfEnum
 ): StatePropsOfControl & OwnPropsOfEnum => {
   const props: StatePropsOfControl = mapStateToControlProps(state, ownProps);
-  const options =
+  const options: EnumOption[] =
+    ownProps.options !== undefined
+      ? ownProps.options
+      : props.schema.enum.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
```
sdirix, line 468: This fails when `props.schema.enum` is undefined.

packages/core/src/util/renderer.ts
```diff
@@ -454,10 +462,31 @@ export const mapStateToEnumControlProps = (
...
+export const mapStateToOneOfEnumControlProps = (
+  state: JsonFormsState,
+  ownProps: OwnPropsOfControl & OwnPropsOfEnum
+): StatePropsOfControl & OwnPropsOfEnum => {
+  const props: StatePropsOfControl = mapStateToControlProps(state, ownProps);
+  const options: EnumOption[] =
     ownProps.options !== undefined
       ? ownProps.options
-      : props.schema.enum || [props.schema.const];
+      : (props.schema.oneOf as JsonSchema[]).map(e => ({label: e.title, value: e.const, description: e.description}));
```
sdirix, line 489: When there is no `title` we should stringify the `const`.

packages/vanilla/src/cells/EnumCell.tsx
```diff
@@ -50,8 +50,8 @@ export const EnumCell = (props: EnumCellProps & VanillaRendererProps) => {
           .concat(
             options.map(optionValue =>
               (
-                <option value={optionValue} label={optionValue} key={optionValue}>
-                  {optionValue}
+                <option value={optionValue.value} label={optionValue.label} key={optionValue.value}>
+                  {optionValue.label}
```
sdirix, line 54: Not really a part of this PR but is `{optionValue.label}` and `label={optionValue.label}` not redundant? Also can we rename `optionValue` to `option`?

packages/core/src/util/cell.ts
sdirix: Formatting is currently off (working on it), so would be great to format this line manually.

packages/core/src/util/cell.ts
```diff
@@ -162,7 +162,8 @@ export const defaultMapStateToEnumCellProps = (
 ): StatePropsOfEnumCell => {
   const props: StatePropsOfCell = mapStateToCellProps(state, ownProps);
   const options: EnumOption[] =
-    ownProps.options !== undefined ? ownProps.options : props.schema.enum.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
+    ownProps.options !== undefined ? ownProps.options :
+    props.schema.enum?.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
```
sdirix, line 166: I would prefer `undefined` when there is no `const` instead of an "empty" option object. I think this should work:
```suggestion
    ownProps.options ||
    props.schema.enum?.map(enumToEnumOptionMapper) || 
    props.schema.const && [enumToEnumOptionMapper(props.schema.const)];
```

packages/core/src/util/renderer.ts
```diff
@@ -464,8 +466,8 @@ export const mapStateToEnumControlProps = (
   const props: StatePropsOfControl = mapStateToControlProps(state, ownProps);
   const options: EnumOption[] =
     ownProps.options !== undefined
-      ? ownProps.options
-      : props.schema.enum.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
+      ? ownProps.options :
+      props.schema.enum?.map(enumToEnumOptionMapper) || [enumToEnumOptionMapper(props.schema.const)];
```
sdirix: same here:
```
    ownProps.options ||
    props.schema.enum?.map(enumToEnumOptionMapper) || 
    props.schema.const && [enumToEnumOptionMapper(props.schema.const)];
```

packages/core/src/util/renderer.ts
```diff
@@ -486,7 +488,8 @@ export const mapStateToOneOfEnumControlProps = (
   const options: EnumOption[] =
     ownProps.options !== undefined
       ? ownProps.options
-      : (props.schema.oneOf as JsonSchema[]).map(e => ({label: e.title, value: e.const, description: e.description}));
+      : (props.schema.oneOf as JsonSchema[])?.
+        map(e => ({label: e.title || JSON.stringify(e.const), value: e.const}));
```
sdirix, line 492: Don't we need the `typeof` check also here?

packages/core/src/util/renderer.ts
```diff
@@ -454,10 +464,33 @@ export const mapStateToEnumControlProps = (
...
+export const mapStateToOneOfEnumControlProps = (
+  state: JsonFormsState,
+  ownProps: OwnPropsOfControl & OwnPropsOfEnum
+): StatePropsOfControl & OwnPropsOfEnum => {
+  const props: StatePropsOfControl = mapStateToControlProps(state, ownProps);
+  const options: EnumOption[] =
+    ownProps.options ||
+    (props.schema.oneOf as JsonSchema[])?.map(e => ({
+      value: e.const,
+      label: e.title || typeof e.const === 'string' ? e.const : JSON.stringify(e.const)
```
sdirix, line 492: We need parenthesis here ;)
```suggestion
      label: e.title || (typeof e.const === 'string' ? e.const : JSON.stringify(e.const))
```

PR #1592, opened 2020-06-17
Fix peer dependencies versions of JSON Forms

sdirix:
The peer dependency versions were incorrectly set to an older version
leading to warnings when installing JSON Forms. They are now adjusted to
be the current version.


sdirix:
Will be fixed with the next release "automatically"

PR #1593, opened 2020-06-17
Feat/autocomplete

eneufeld:


sdirix:
Updated the PR to latest `master`

packages/material/src/mui-controls/MuiAutocomplete.tsx
```diff
@@ -0,0 +1,81 @@
...
+    handleChange,
+    options,
+    config,
+    getOptionLabel,
+    renderOptionLabel
+  } = props;
+  const appliedUiSchemaOptions = merge({}, config, uischema.options);
+  const [inputValue, setInputValue] = React.useState('');
+
+  const findOption = data ? options.find(o => o.value === data) : null;
```
sdirix, line 55: Why specifically using `null`?
```suggestion
  const findOption = options.find(o => o.value === data);
```

packages/material/src/controls/MaterialOneOfEnumControl.tsx
```diff
@@ -31,13 +31,13 @@ import {
...
 import { withJsonFormsOneOfEnumProps } from '@jsonforms/react';
-import { MuiSelect } from '../mui-controls/MuiSelect';
+import { MuiAutocomplete } from '../mui-controls/MuiAutocomplete';
 import { MaterialInputControl } from './MaterialInputControl';
 
 export const MaterialOneOfEnumControl = (props: ControlProps & OwnPropsOfEnum) => (
   <MaterialInputControl
     {...props}
-    input={MuiSelect}
+    input={MuiAutocomplete}
```
sdirix, line 40: I think I would prefer making `MuiAutocomplete` the default here and in the normal enum but making it configurable via `options: { autoComplete: false }`. Don't forget to check the `config` too.

PR #1600, opened 2020-07-02
Extend Autocomplete Control to allow custom filterOptions

eneufeld:
* extend autocomplete control
* extend enum example
* add custom autocomplete renderer


packages/material/example/index.ts
```diff
@@ -25,5 +25,6 @@
 import { renderExample } from '../../example/src/index';
 import { materialCells } from '../src';
 import { extendedMaterialRenderers } from '../src/extended';
+import MyAutocompleteControl, { myAutocompleteTester } from './customAutocomplete';
 
-renderExample(extendedMaterialRenderers, materialCells);
+renderExample([{renderer: MyAutocompleteControl, tester: myAutocompleteTester}, ...extendedMaterialRenderers], materialCells);
```
sdirix, line 30: I would prefer to only register the renderer for the enum example so we have a chance to see the normal renderer break.

PR #1606, opened 2020-07-09
Add show/hide data to rule example

TheZoker:


packages/examples/src/rule.ts
```diff
@@ -66,14 +73,35 @@ export const uischema = {
...
   ]
 };
 
 export const data = {
   name: 'John Doe',
   alive: true,
-  kindOfDead: 'Zombie'
+  kindOfDead: 'Zombie',
+  vegetables: true,
+  kindOfVegetables: 'Only potatos'
```
sdirix: This does not match the enum `Only potato` so there is a validation error.

PR #1615, opened 2020-07-29
Angular renderer for object arrays

AlexandraBuzila:
Add array layout renderer for objects.

### DIFF
--- a/packages/angular-material/src/layouts/array-layout.renderer.ts
+++ b/packages/angular-material/src/layouts/array-layout.renderer.ts
@@ -30,7 +30,7 @@ import {
 import {
   ArrayControlProps,
   createDefaultValue,
-  generateDefaultUISchema,
+  findUISchema,
   isObjectArray,
   mapDispatchToArrayControlProps,
   OwnPropsOfRenderer,
@@ -39,7 +39,8 @@ import {
   rankWith,
   setReadonly,
   UISchemaElement,
-  unsetReadonly
+  UISchemaTester,
+  unsetReadonly,
 } from '@jsonforms/core';
 
 @Component({
@@ -84,10 +85,13 @@ import {
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class ArrayLayoutRenderer extends JsonFormsArrayControl {
-  private _uischema: UISchemaElement;
   noData: boolean;
   addItem: (path: string, value: any) => () => void;
   removeItems: (path: string, toDelete: number[]) => () => void;
+  uischemas: {
+    tester: UISchemaTester;
+    uischema: UISchemaElement;
+  }[];
 
   constructor(private jsonFormAngularService: JsonFormsAngularService) {
     super(jsonFormAngularService);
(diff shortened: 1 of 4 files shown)


packages/angular-material/src/layouts/array-layout.renderer.ts
```diff
@@ -0,0 +1,165 @@
...
+                Remove
+              </button>
+            </mat-card-actions>
+          </mat-card>
+        </div>
+      </div>
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush
+  })
+  export class ArrayLayoutRenderer extends JsonFormsBaseRenderer<ControlElement>
```
eneufeld, line 92: why not extend the JsonFormsArrayControl?

packages/angular-material/src/layouts/array-layout.renderer.ts
```diff
@@ -0,0 +1,165 @@
...
+    template: `
+      <div fxLayout="column" fxLayoutGap="16px" [fxHide]="hidden">
+        <mat-toolbar>
+          <h2>{{ this.label }}</h2>
+          <span fxFlex></span>
+          <button
+            mat-icon-button
+            [disabled]="readonly"
+            (click)="add()"
+            attr.aria-label="{{ 'Add to ' + this.label + 'button' }}"
```
eneufeld, line 60: we will need to find a localization solution

packages/angular-material/src/layouts/array-layout.renderer.ts
AlexandraBuzila: Good point! I changed the renderer so it extends `JsonFormsArrayControl`

packages/angular-material/src/layouts/array-layout.renderer.ts
```diff
@@ -0,0 +1,145 @@
...
+  get defaultUISchema() {
+    if (!this._uischema) {
+      this._uischema = generateDefaultUISchema(this.scopedSchema);
+    }
+    return this._uischema;
+  }
+
+  getProps(index: number): OwnPropsOfRenderer {
+    const uischema =
+      (this.uischema.options?.detailUISchema as UISchemaElement) ??
```
eneufeld, line 124: we should unify this with the react renderer and use `detail`

eneufeld: or even better use the findschema like here: https://github.com/eclipsesource/jsonforms/blob/6cf39efbb82fdf77b173c38ca4b706d526920f9b/packages/material/src/layouts/ExpandPanelRenderer.tsx#L101

packages/angular-material/src/layouts/array-layout.renderer.ts
```diff
@@ -0,0 +1,148 @@
...
+    };
+  }
+  trackByFn(index: number) {
+    return index;
+  }
+}
+
+export const ArrayLayoutRendererTester: RankedTester = rankWith(
+  4,
+  isObjectArray
```
eneufeld, line 147: shouldn't we check for `isObjectArrayWithNesting` instead?

AlexandraBuzila: Isn't that tester too restrictive? If we don't have a nested array or don't provide a detail schema in the options, then the new renderer won't be used.

packages/angular-material/src/layouts/array-layout.renderer.ts
```diff
@@ -0,0 +1,148 @@
...
+export class ArrayLayoutRenderer extends JsonFormsArrayControl {
+  noData: boolean;
+  addItem: (path: string, value: any) => () => void;
+  removeItems: (path: string, toDelete: number[]) => () => void;
+  uischemas: {
+    tester: UISchemaTester;
+    uischema: UISchemaElement;
+  }[];
+
+  constructor(private jsonFormAngularService: JsonFormsAngularService) {
```
eneufeld, line 96: we don't need to make this private as we can access it from the parent

eneufeld: If not, it should be accessible via the parent

eneufeld:
Do you know why neither the buttons nor the toolbar is using the default styling colors of material?

packages/angular-material/src/layouts/array-layout.renderer.ts
AlexandraBuzila: Fixed

AlexandraBuzila:
I did the small refactoring the the array renderer.

Regarding the styling:
- the button style was not consistent with the other renderers, I fixed that
- the toolbar styling was a little different because we were not using the Roboto font.  I added it to the Angular Material examples, but I can revert this change if you don't agree with it.
The toolbar should look now like in the Angular Material documentation: https://material.angular.io/components/toolbar/overview

AlexandraBuzila:
Ah, and the toolbar color is not set to 'primary'. I'll update my commit to use it.

Issue #1498, opened 2019-09-09
Ajv loading requires `unsafe-eval` policy for `script-src` Content-Security-Policy


clayroach:
**Is your feature request related to a problem? Please describe.**
For secure environments, the content security policy requires the `unsafe-eval` policy directive for the `script-src` policy.  This requirement is primarily due to the use of Ajv's use of the `Function` constructor per https://github.com/epoberezkin/ajv/issues/406

**Describe the solution you'd like**
The best resolution to this is to enable pre-compilation of the JSON schemas so that they don't have to be compiled at runtime.  This can be accomplished using `ajv-pack` https://github.com/epoberezkin/ajv-pack

**Describe alternatives you've considered**
Alternatives or workarounds is to lazy load the Ajv validator so that it only loads when JsonForms components are rendered.  This would allow compiling a version that disables jsonforms for environment with strict CSP's.

**Describe for which setup you like to have the improvement**
Framework: Material-UI (or any of the others, this should apply)

**Additional context**
See screen shots below for some examples of this error
<img width="2209" alt="Screen Shot 2019-09-09 at 9 13 50 AM" src="https://user-images.githubusercontent.com/1972937/64538474-4f83f280-d2e2-11e9-9b3b-40082005e447.png">

<img width="1295" alt="Screen Shot 2019-09-09 at 8 57 42 AM" src="https://user-images.githubusercontent.com/1972937/64538559-77735600-d2e2-11e9-83c7-2e62436dccc9.png">

Lily418:
I like the look of ajv-pack, this is also related to a potential performance enhancement discussed here. Pre-complilation is great because avoids a penalty for doing this at runtime which is non-trivial for large schemas.
https://spectrum.chat/jsonforms/general/performance-enhancements-for-ref-resolving~4fb7b288-ddad-43b4-9697-07e923a1e667

  
Although [ajv-pack does not support recursive references](https://github.com/epoberezkin/ajv-pack#limitations) which is a feature the eclipsesource team are currently looking to support. https://github.com/eclipsesource/jsonforms/issues/1476

I wonder if its possible to use ajv-pack already by passing an object to init which implements the methods of ajv that JSON Forms uses but uses ajv-pack instead
Take a look a the init action 
https://github.com/eclipsesource/jsonforms/blob/master/packages/core/src/actions/index.ts#L83

and example of use
https://github.com/eclipsesource/jsonforms-react-seed/blob/master/src/index.tsx

Something like 
```
{
[...]

PR #1624, opened 2020-08-19
always reuse AJV instance from store

DBosley:
This PR changes `packages/core/src/util/runtime.ts` so it no longer creates an instance of AJV on file load. It will only run `createAjv()` if an instance of AJV is not supplied to any of the exported utilities. This moves the project closer to full abstraction of AJV to allow it as a peer dependency. 

This change will allow users to inject their own instances of AJV and that instance will be used throughout the library. It allows a user to create mocks or delegates for core AJV functions which can be leveraged to void unsafe-eval errors in projects with CSP enabled.

#1498 #1557


packages/angular-material/src/other/label.renderer.ts
```diff
@@ -46,7 +47,7 @@ const mapStateToProps = (
   const visible =
     ownProps.visible !== undefined
       ? ownProps.visible
-      : isVisible(ownProps.uischema, getData(state));
+      : isVisible(ownProps.uischema, getData(state), undefined, getOrCreateAjv(state));
```
sdirix, line 50: We can assume that Ajv exists

packages/core/src/reducers/core.ts
```diff
@@ -283,6 +283,7 @@ export const coreReducer = (
 export const extractData = (state: JsonFormsCore) => get(state, 'data');
 export const extractSchema = (state: JsonFormsCore) => get(state, 'schema');
 export const extractUiSchema = (state: JsonFormsCore) => get(state, 'uischema');
+export const extractOrCreateAjv = (state: JsonFormsCore) => getOrCreateAjv(state);
```
sdirix, line 286: Please change to `extractAjv`

packages/core/src/util/cell.ts
```diff
@@ -107,12 +107,12 @@ export const mapStateToCellProps = (
   const visible =
     ownProps.visible !== undefined
       ? ownProps.visible
-      : isVisible(uischema, rootData);
+      : isVisible(uischema, rootData, undefined, getOrCreateAjv(state));
```
sdirix, line 110: Please change to `getAjv(state)`

packages/core/src/reducers/index.ts
```diff
@@ -98,6 +100,9 @@ export const getUiSchema = (state: JsonFormsState): UISchemaElement =>
   extractUiSchema(get(state, 'jsonforms.core'));
 export const getRefParserOptions = (state: JsonFormsState): RefParser.Options =>
   extractRefParserOptions(get(state, 'jsonforms.core'));
+export const getOrCreateAjv = (
```
sdirix, line 103: Please change to `getAjv`

packages/core/src/util/index.ts
```diff
@@ -132,11 +134,11 @@ export { composePaths, composeWithUi, Paths, toDataPath };
 
 // Runtime --
 const Runtime = {
-  isEnabled(uischema: UISchemaElement, data: any): boolean {
-    return isEnabled(uischema, data);
+  isEnabled(uischema: UISchemaElement, data: any, ajv: Ajv = createAjv()): boolean {
```
sdirix, line 137: Please remove the default for ajv

packages/core/src/util/renderer.ts
```diff
@@ -274,6 +275,12 @@ export interface StatePropsOfRenderer {
    */
 
   cells?: JsonFormsCellRendererRegistryEntry[];
+
+  /**
+   * AJV instance from core state.
+   */
+
+   ajv?: Ajv;
```
sdirix, line 283: Please remove the ajv prop

packages/core/src/util/renderer.ts
```diff
@@ -725,7 +733,8 @@ export const mapStateToLayoutProps = (
     data,
     uischema: ownProps.uischema,
     schema: ownProps.schema,
-    direction: ownProps.direction || layoutDefaultProps.direction
+    direction: ownProps.direction || layoutDefaultProps.direction,
+    ajv
```
sdirix, line 737: Please remove ajv here

packages/core/src/util/renderer.ts
```diff
@@ -702,15 +709,16 @@ export const mapStateToLayoutProps = (
   ownProps: OwnPropsOfLayout
 ): LayoutProps => {
   const rootData = getData(state);
+  const ajv = getOrCreateAjv(state);
```
sdirix, line 712: This should be `get` and all the functions should just use this variable.

packages/material/src/layouts/MaterialCategorizationLayout.tsx
```diff
@@ -92,7 +92,8 @@ export class MaterialCategorizationLayoutRenderer extends RendererComponent<
       uischema,
       visible,
       enabled,
-      selected
+      selected,
+      ajv
```
sdirix, line 96: Please hand in this ajv with an extra HOC

sdirix (changes requested):
Thanks for this PR!

 * We can assume that Ajv will always be in the state as we require that the init action was executed before rendering. Therefore we can just always extract Ajv from the state without creating a new instance when it doesn't exist (as this has already happened if needed in the init action).
 * I would also prefer not giving Ajv as a prop to all renderers as no renderer besides `MaterialCategorizationLayout` and `MaterialCategorizationStepperLayout` care about it. Instead we could add a `withAjv` HOC which can be used by any renderer additionally to their `withJsonFormsXXXProps` HOC, e.g. `withAjv(withJsonFormsLayoutProps(MaterialCategorizationLayoutRenderer))`.

packages/material/src/layouts/MaterialCategorizationLayout.tsx
DBosley: Should this extra HOC come from /core/src/util/renderer.ts?

How should it be named? I'm trying to figure out the pattern to use but struggling.

sdirix: It's fine to just place it in the `MaterialCategorizationLayoutRenderer` itself for now. You can take a look at `withVanillaControlProps` as a guide. It should look something like this (untested):
```
export const withAjv = (Component) => (props) => {
  const ctx = useJsonForms();
  return <Component {...props} {ajv: ctx.core.ajv} />;
}
```

DBosley: Oh! I can get it with `withJsonFormsContext`, then just do a little `getAjv({ jsonforms: { ...ctx } })`

packages/angular-material/src/other/label.renderer.ts
DBosley: Fixed

packages/core/src/reducers/core.ts
DBosley: Fixed

packages/core/src/util/cell.ts
DBosley: Fixed

packages/core/src/reducers/index.ts
DBosley: Fixed

packages/core/src/util/renderer.ts
DBosley: Fixed

packages/core/src/util/renderer.ts
DBosley: Fixed

packages/core/src/util/renderer.ts
DBosley: Fixed

packages/material/src/layouts/MaterialCategorizationLayout.tsx
DBosley: I didn't see this comment until now. The HOC is needed in 2 files in the material package, so I put it in the `util/layout.tsx` file. If you'd like it somewhere else I can move it.

Also, I set up it's types so it's able to correctly return the prop types for the outputted component. This pattern might be helpful in the `withJsonFormsContext` HOC as well if you want to try it out there.

DBosley: Oh, and just a heads up: `withAjv(withJsonFormsLayoutProps(MaterialCategorizationLayoutRenderer))` doesn't work because `withJsonFormsLayoutProps` returns a component with type: `React.ComponentType<OwnPropsOfLayout>`. It would have to be a generic which returns a component of type `React.ComponentType<OwnPropsOfLayout & T>` where T are any props that are additional to the OwnPropsOfLayout + any props that are injected and abstracted away.

This is a simple issue to work around, however. It just means you have to do it in this order:

`withJsonFormsLayoutProps(withAjvProps(MaterialCategorizationStepperLayoutRenderer));`

PR #1636, opened 2020-09-15
[Angular] Array Layout renderer fixes

AlexandraBuzila:
- reduce colors and space
- show error marker in array header
- use ArrayLayoutProps for easier error handling
- add tooltips
- allow labels and tooltips to be overwritten in subclasses

Selection_15-09-2020-14:05:21


eneufeld:
Could you double check the tooltip:
image

eneufeld:
Can you also check the `Array with detail and rule` example? It should be quite simple to fix

eneufeld (approved):

PR #1644, opened 2020-09-30
Cache generated schema and ui schema in JsonForms

sdirix:
The JsonForms React component now caches generated schema and ui schemas
to avoid triggering endless rendering loops. These could occur when the
outside app also rerenders JsonForms whenever an "onChange" event is
emitted.


AlexandraBuzila (approved):
Nice fix! Works well!

PR #1646, opened 2020-09-30
[Angular] Fix array-layout-renderer Component

AlexandraBuzila:
Move component definition inside component annotation


sdirix (approved):
LGTM

Issue #1627, opened 2020-08-20
Don't cut off text with the "InputAdornment"


sdirix:
Some material inputs contain an `InputAdornment` to render a "Clear" button. Even when the button is not visible it still cuts off the text. This should be improved in some way, for example by layering the text (but still looking good) or making the button smaller and potentially even outside of the input. Please experiment and check what looks / feels the best.

![Cutoff](https://user-images.githubusercontent.com/8998368/90773594-6d314880-e2f6-11ea-9180-3fd13cccc07c.png)
![CutoffReason](https://user-images.githubusercontent.com/8998368/90773600-6f93a280-e2f6-11ea-854f-e04b60d7a2f3.png)

PR #1648, opened 2020-10-07
Fix adornment placement with display instead of visibility

TheZoker:
The input clear icon overlayed the input text, so
part of the text was not visible. 
This commit fixes that by replacing the 'visibility' 
with the 'display' css property. 

Fixes: #1627


sdirix:
I adjusted the styling a bit. What do you think?

sdirix (changes requested):

sdirix (approved):
LGTM!

Issue #1537, opened 2020-01-22
Remove redux dependencies from JSON Forms


sdirix:
# Summary
 - [x] Investigate different approaches and effort
 - [ ] Remove redux from all packages

## Investigation

 - [x] Investigate removal of redux from `core` and `react`. Also check the repercussion in `vanilla`, `material`
 - [x] Define a rough approach / outline of the steps needed to remove redux from `core`, `react` and have `vanilla` and `material` renderers in a working state
 - [x] Check which functionality is missing when redux is removed. This can be done by listing all available actions and check whether their functionality can be handled with the standalone version of JSON Forms in some way.
 - [x] Convert a testcase relying on redux to instead test the standalone version and estimate the effort of converting all tests
- [ ] Investigate and estimate the effort of adapting the provided examples

sdirix:
This was completed with the release of JSON Forms 2.5.

PR #1649, opened 2020-10-07
Remove redux dependencies from JSON Forms

TheZoker:
Removes redux as a direct dependency. 
- Redux is removed from the material package (not needed anyway)
- The missing types have been replaced with own types
- `jsonformsRecducer` was removed from the core package (not used anyway)
- All material tests have been rewritten to run without redux
- `react-redux` was moved from the `peerDepedencies` to the `optionalPeerDependencies`
- The `jsonFormsReduxContext` is now available under `@jsonforms/react/redux`
```
import { jsonFormsReduxContext } from `@jsonforms/react/lib/redux`;
```

Related issue: https://github.com/eclipsesource/jsonforms/issues/1537


packages/core/src/reducers/index.ts
```diff
@@ -78,20 +76,6 @@ export {
...
-    core: coreReducer,
-    renderers: rendererReducer,
-    cells: cellReducer,
-    config: configReducer,
-    uischemas: uischemaRegistryReducer,
-    defaultData: defaultDataReducer,
-    i18n: i18nReducer,
-    ...additionalReducers
-  });
-
```
sdirix, line 94: We need to keep the `jsonformsReducer` as they are needed when using JSON Forms with Redux, see for example the React seed or the example apps.

sdirix (changes requested):
I tried it with the React seed and was successfully able to compile and run JSON Forms without installing Redux :tada: 

However I was not able to run the fallback Redux variant as the `jsonformsReducer` is now missing. Please check whether we can restore this export.

Please also check:
 * All tests should run. Currently the core tests (and maybe more) are failing.
 * All example apps should. Currently at least the React-Material example app doesn't start.

packages/core/src/util/type.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+
```
sdirix, line 25: Please add a line from where the types were copied.

packages/core/src/util/type.ts
```diff
@@ -0,0 +1,146 @@
...
+  [Symbol.observable](): Observable<T>
+};
+
+/**
+ * An Observer is used to receive data from an Observable, and is supplied as
+ * an argument to subscribe.
+ */
+export type Observer<T> = {
+  next?(value: T): void
+};
```
sdirix, line 146: We don't need these types, right? Please remove them when not needed.

packages/core/src/reducers/index.ts
```diff
@@ -80,8 +79,7 @@ export { JsonFormsCore, ValidationMode };
 
 export const jsonformsReducer = (
```
sdirix, line 80: Please rename as discussed.

sdirix (changes requested):
Looks good! Some minor comments.

packages/core/src/util/type.ts
TheZoker: `Observer` is used by `Observable` and `Observable` is used by `Store`, so I think we need these types here, right?

sdirix: Do we use the Observable of the Store somewhere?

sdirix (changes requested):
Build is failing

sdirix (changes requested):
The typings for the `jsonFormsReducerConfig` fail when trying to use a built version. It seems each of the reducers should declare an appropriate `Reducer<State,Action>` type instead of relying on automatic type conversion.

sdirix (approved):
LGTM

PR #1651, opened 2020-10-20
yargs-parser peer dep upgrade for cve 1500

drkstr101:
This still needs to be thoroughly tested for major API version bump, but I thought I get the review process started.

See: https://www.npmjs.com/advisories/1500

Signed-off-by: Aaron R Miller <amiller@watheia.dev>


drkstr101:
ah bummer,I need to add my new email address to Eclipse legal first.

drkstr101:
Alright, it's good now.

Cheers!

sdirix:
In general I have no problem with updating `ava` to a later version, especially to fix a potential security issue in the dev tooling.

Out of curiosity: Is there a specific reason to fix exactly this vulnerability?

package.json
```diff
@@ -18,7 +18,7 @@
     "@types/jest": "^24.0.23",
     "@types/lodash": "^4.14.149",
     "ajv": "^6.10.2",
-    "ava": "~2.4.0",
+    "ava": "^3.13.0",
```
sdirix, line 21: Please also update all other `ava` dependencies in each package.  We hoist all dependencies so ideally they should all confirm to the same version.

sdirix (changes requested):

package.json
drkstr101: > Out of curiosity: Is there a specific reason to fix exactly this vulnerability?

No there is not. The risk profile is practically nill. I just had some audit warnings crop up that we'rent there the day before, and so I attempted to find the root cause and submit PR as a courtesy. I would not be heartbroken if you decide it's not worth the hassle.

That said, I'll go ahead and make those updates to `ava` dependencies, and anything else you think we need as a result.

drkstr101:
@sdirix I've run the full upgrade, then an `audit fix`, and this time npm took care of the 2 that cropped up automatically.

Please let me know if there was anything else I can do to help.

Cheers!

drkstr101:
PS: One thing I didn't mention is that there is a slight value add to the upgrade beyond the audit warnings, and that is the fact that downstream consumers of this package will not have to `--force` upgrade their own installs to resolve the issue, with the slim but non-zero chance this will break things. I didn't mean to imply I was simply being pedantic about the audit warnings.

sdirix (changes requested):
It seems that the Angular test setup is now broken:
> lerna ERR! npm run test-cov stderr:
  ✖ Enhancement compilation must be configured in AVA’s Babel options

drkstr101:
Indeed it did. I just saw the `travis-ci` report. Not sure why it seemed to work ok for me. I probably didn't execute the tests correctly. Thanks for bringing it to my attention. It's on my TODO list for tomorrow. Cheers!

PR #1652, opened 2020-10-28
Add style context

TheZoker:
- Add style context for vanilla renderers
- Replaces core styles with the new styles context 

Related issue: https://github.com/eclipsesource/jsonforms/issues/1537

### DIFF
--- a/packages/vanilla/src/styleContext.ts
+++ b/packages/vanilla/src/styleContext.ts
@@ -1,0 +1,52 @@
+/*
+  The MIT License
+  
+  Copyright (c) 2017-2019 EclipseSource Munich
+  https://github.com/eclipsesource/jsonforms
+  
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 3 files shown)


sdirix (changes requested):
Looks pretty good! However we need two changes:
 - We should export the new context via the index so it can be easily used ;)
 - `useStyles` should contain a fallback to consume the regular context styles if there aren't any in the `StyleContext`

sdirix (changes requested):
The testcases currently fail. Can you have a look?

sdirix (approved):
LGTM

packages/vanilla/src/styleContext.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export interface StyleContext {
+  styles: StyleDef[];
+}
+
+const defaultContext: any = {
+  styles: []
+};
+
+export const StyleContextInstance = React.createContext<StyleContext>(
```
sdirix, line 38:
```suggestion
export const JsonFormsStyleContext = React.createContext<StyleContext>(
```

packages/vanilla/src/styleContext.ts
```diff
@@ -0,0 +1,52 @@
...
+  defaultContext
+);
+
+export const useStyleContext = (): StyleContext =>
+  useContext(StyleContextInstance);
```
sdirix, line 43:
```suggestion
  useContext(JsonFormsStyleContext);
```

PR #1654, opened 2020-10-29
Add RadioGroupControl support for "oneOf" enums

AlexandraBuzila:
- the radio group control can now handle 'oneOf' schemas as well. If the
oneOf item has a title, it will be shown as the label for the option
- update Radio Group example with a oneOf demo
- add test

### DIFF
--- a/packages/examples/src/radioGroup.ts
+++ b/packages/examples/src/radioGroup.ts
@@ -33,17 +33,49 @@ const schema = {
       exampleRadioEnum: {
         type: 'string',
         enum: ['One', 'Two', 'Three']
-      }
+      },
+      exampleRadioOneOfEnum: {
+        type: 'string',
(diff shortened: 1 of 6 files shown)


sdirix (changes requested):
Thanks for the PR!

I would prefer when we could use the same approach as with the regular enum renderers, i.e. use `withJsonFormsEnumProps` for the `MaterialRadioGroupControl` and `withJsonFormsOneOfEnumProps` for the `MaterialOneOfRadioGroupControl`, both of them using the same control.

AlexandraBuzila:
I rebased the branch and added a dedicated control `MaterialOneOfRadioGroupControl` for the oneof enums. The code looks a lot cleaner now, thank you for the review!

sdirix (approved):
LGTM

PR #1655, opened 2020-10-30
Improve performance

AlexandraBuzila:
### Avoid unnecessary initialization calls in React
Use the initialization function of 'useReducer' to avoid unnecessary dispatching of init actions whose result is ignored anyway.

### Only update state with changed properties of JsonFormsCore
Instead of calling the INIT action when one of the core properties changes, introduce a new action (UPDATE_CORE) which updates only the changed properties in the state.


sdirix (approved):
LGTM!

PR #1656, opened 2020-11-13
Fix useLayoutEffect warning for SSR with React

JovanShandro:
We tried using the JsonForms component when doing server side rendering with react using
ReactDomServer and because of the use of the useLayoutEffect hook, we would get a  warning of 'Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format...' all the time.

This pull request removes the warning by using useEffect instead if we are not in the browser.

Thank you!

### DIFF
--- a/packages/react/src/JsonForms.tsx
+++ b/packages/react/src/JsonForms.tsx
@@ -25,7 +25,8 @@
 import isEqual from 'lodash/isEqual';
 import maxBy from 'lodash/maxBy';
 import memoize from 'lodash/memoize';
-import React, { useLayoutEffect, useMemo } from 'react';
+import React, { useMemo } from 'react';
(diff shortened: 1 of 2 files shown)


sdirix:
Thanks for this contribution! We also want to remove the `useLayoutEffect` hook and [are already working on it](https://github.com/eclipsesource/jsonforms/pull/1655). We want to merge soon so the next (alpha) release should already have these changes in it.

JovanShandro:
> Thanks for this contribution! We also want to remove the `useLayoutEffect` hook and [are already working on it](https://github.com/eclipsesource/jsonforms/pull/1655). We want to merge soon so the next (alpha) release should already have these changes in it.

Thank you for quick reply!  Glad to hear that.

sdirix:
#1655 is now merged. The next release of JSON Forms will no longer contain the `useLayoutEffect` hook :+1:

PR #1657, opened 2020-11-24
Make Angular component consumable multiple times

eneufeld:
The JSONForms Angular component needs the jsonforms service
to set the right data, schema, uischema etc.
The problem is, that the service is a singleton,
as it is registering as a root service.

This commit introduces a new root component `jsonforms`
which allows to pass all parameters into the component.
The component is now initializing the service internally.
Furthermore, the service is not registered as a root service anymore.
Now multiple jsonforms components can be used next to each other.

### DIFF
--- a/packages/angular-material/example/app/app.component.ts
+++ b/packages/angular-material/example/app/app.component.ts
@@ -1,7 +1,7 @@
 /*
   The MIT License
 
-  Copyright (c) 2017-2019 EclipseSource Munich
+  Copyright (c) 2017-2020 EclipseSource Munich
   https://github.com/eclipsesource/jsonforms
 
   Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -23,19 +23,28 @@
   THE SOFTWARE.
 */
 import { Component } from '@angular/core';
-import { JsonFormsAngularService } from '@jsonforms/angular';
-import {
-  Actions,
-  getData,
-  getUiSchema,
-  setLocale,
-  setReadonly,
-  unsetReadonly,
-  setConfig
-} from '@jsonforms/core';
-import { getExamples } from '@jsonforms/examples';
-import { of } from 'rxjs';
-import { map } from 'rxjs/operators';
+import { ExampleDescription, getExamples } from '@jsonforms/examples';
+import { UISchemaElement, UISchemaTester } from '@jsonforms/core';
+import { angularMaterialRenderers } from '../../src/index';
+const uiSchema = {
+  type: 'HorizontalLayout',
+  elements: [
+    {
+      type: 'Control',
+      scope: '#/properties/buyer/properties/email'
+    },
+    {
+      type: 'Control',
+      scope: '#/properties/status'
+    }
+  ]
+};
+const itemTester: UISchemaTester = (_schema, schemaPath, _path) => {
+  if (schemaPath === '#/properties/warehouseitems/items') {
+    return 10;
+  }
+  return -1;
+};
 @Component({
   selector: 'app-root',
   template: `
@@ -45,7 +54,7 @@ import { map } from 'rxjs/operators';
       Example:
       <select (change)="onChange($event)">
         <option
-          *ngFor="let example of exampleData$ | async"
+          *ngFor="let example of examples"
           value="{{ example.name }}"
           label="{{ example.label }}"
         >
@@ -57,65 +66,50 @@ import { map } from 'rxjs/operators';
       <button (click)="changeLocale('de-DE')">Change locale to de-DE</button>
       <button (click)="changeLocale('en-US')">Change locale to en-US</button>
       Current locale: {{ currentLocale }}
-      <button (click)="setReadonly()">
+      <button (click)="toggleReadonly()">
         {{ readonly ? 'Unset' : 'Set' }} Readonly
       </button>
     </div>
-    <jsonforms-outlet></jsonforms-outlet>
+    <jsonforms
+      [data]="selectedExample.data"
+      [schema]="selectedExample.schema"
+      [uischema]="selectedExample.uischema"
+      [renderers]="renderers"
+      (dataChange)="onDataChange($event)"
+      [locale]="currentLocale"
+      [uischemas]="uischemas"
+      [readonly]="readonly"
+    ></jsonforms>
   `
 })
 export class AppComponent {
-  readonly exampleData$ = of(getExamples());
+  readonly renderers = angularMaterialRenderers;
+  readonly examples = getExamples();
+  selectedExample: ExampleDescription;
   currentLocale = 'en-US';
   private readonly = false;
   data: any;
+  uischemas: { tester: UISchemaTester; uischema: UISchemaElement; }[] = [
+    { tester: itemTester, uischema: uiSchema }
+  ];
 
-  constructor(private jsonformService: JsonFormsAngularService) {
-    const examples = getExamples();
-    const selectedExample = examples[0];
-    this.jsonformService.updateCore(
-      Actions.init(
-        selectedExample.data,
-        selectedExample.schema,
-        selectedExample.uischema
-      )
-    );
-
-    this.jsonformService.$state.pipe(
-      map(state => getData(state))
-    ).subscribe(data => this.data = data);
-  }
-  examples() {
-    return getExamples();
+  constructor() {
+    this.selectedExample = this.examples[19];
   }
 
-  onChange = (ev: any) => {
-    const examples = getExamples();
-    const selectedExample = examples.find(e => e.name === ev.target.value);
-    this.jsonformService.updateCore(
-      Actions.init(
-        selectedExample.data,
-        selectedExample.schema,
-        selectedExample.uischema
-      )
-      );
-      this.jsonformService.updateConfig(setConfig(selectedExample.config));
-      this.jsonformService.updateLocale(setLocale(this.currentLocale));
-  };
+  onDataChange(data: any) {
+    this.data = data;
+  }
+
+  onChange(ev: any) {
+    this.selectedExample = this.examples.find(e => e.name === ev.target.value);
+  }
 
   changeLocale(locale: string) {
     this.currentLocale = locale;
-    this.jsonformService.updateLocale(setLocale(locale));
   }
 
-  setReadonly() {
-    const uischema = getUiSchema(this.jsonformService.getState());
-    if (this.readonly) {
-      unsetReadonly(uischema);
-    } else {
-      setReadonly(uischema);
-    }
+  toggleReadonly() {
     this.readonly = !this.readonly;
-    this.jsonformService.updateCore(Actions.setUISchema(uischema));
   }
 }
--- a/packages/angular-test/src/boolean.ts
+++ b/packages/angular-test/src/boolean.ts
@@ -90,9 +90,8 @@ export const booleanBaseTest = <C extends JsonFormsControl, I>(
   it('should render', () => {
     component.uischema = uischema;
 
-    getJsonFormsService(component).init();
-    getJsonFormsService(component).updateCore(
-      Actions.init(data, defaultBooleanTestSchema)
+    getJsonFormsService(component).init(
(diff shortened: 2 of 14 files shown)


AlexandraBuzila (changes requested):
Overall the changes look good, but there seem to be some problems now with the example app:
- there's no validation. Open any example with a required property (e.g. Day 2) and remove the data from the required field -> no errors are shown
- some examples don't work anymore, e.g. Day 1 now shows `No applicable renderer found!`

AlexandraBuzila (approved):
Thanks, looks good now!

packages/angular-material/example/app/app.component.ts
```diff
@@ -61,61 +76,52 @@ import { map } from 'rxjs/operators';
...
     if (this.readonly) {
-      unsetReadonly(uischema);
+      unsetReadonly(newUiSchema);
     } else {
-      setReadonly(uischema);
+      setReadonly(newUiSchema);
     }
     this.readonly = !this.readonly;
-    this.jsonformService.updateCore(Actions.setUISchema(uischema));
+    this.selectedExample.uischema = newUiSchema;
```
sdirix, line 125: This could be replaced with the new readonly support we have in the core, right?

packages/angular-test/src/boolean.ts
```diff
@@ -279,22 +279,26 @@ export const booleanErrorTest = <C extends JsonFormsControl, I>(
...
+      {
+        dataPath: 'foo',
+        message: 'Hi, this is me, test error!',
+        keyword: '',
+        schemaPath: '',
+        params: ''
+      }
+    ];
+    // tslint:disable-next-line: no-string-literal
+    formsService['updateSubject']();
```
sdirix, line 301: Why do we have to do it like this, couldn't we also use the `setErrors` action?

packages/angular/src/jsonforms.service.ts
```diff
@@ -1,58 +1,108 @@
...
         this.state = new BehaviorSubject({ jsonforms: this._state });
+        if (initialState?.core.data && initialState?.core.schema) {
+            this.updateCore(
+                Actions.init(
+                    initialState.core.data,
+                    initialState.core.schema,
+                    initialState.core.uischema ?? generateDefaultUISchema(initialState.core.schema)
+                )
+            );
+        }
```
sdirix, line 67: We only use `Actions.init` when the user provides data and schema? I would prefer dispatching `Actions.init` in any case no matter what the user hands over to guarantee a consistent state. In the core package and at least in React we always assume that `Actions.init` was executed.

packages/angular/src/jsonforms.service.ts
```diff
@@ -82,25 +132,62 @@ export class JsonFormsAngularService {
...
+
     updateConfig<T extends SetConfigAction>(setConfigAction: T): T {
         const configState = configReducer(this._state.config, setConfigAction);
         this._state.config = configState;
         this.updateSubject();
         return setConfigAction;
     }
 
+    setUiSchema(uischema: UISchemaElement): void {
+        const newUiSchema = uischema ?? generateDefaultUISchema(this._state.core.schema);
```
sdirix, line 158: shouldn't it be reflected in the type that it could be `undefined`?

packages/angular/src/jsonforms.service.ts
```diff
@@ -82,25 +132,62 @@ export class JsonFormsAngularService {
...
+        const coreState = coreReducer(this._state.core, Actions.updateCore(this._state.core.data, this._state.core.schema, newUiSchema));
+        this._state.core = coreState;
+        this.updateSubject();
+    }
+
+    setSchema(schema: JsonSchema): void {
+        const coreState = coreReducer(this._state.core, Actions.updateCore(this._state.core.data, schema, this._state.core.uischema));
```
sdirix, line 165: In `setUiSchema` we generate the ui schema when the user hands over `undefined`. We should do the same here.

sdirix (changes requested):
Looks good in general! Can you add a small description in the migration guide?

sdirix:
I assume an update to the Angular seed app is then required? Could you provide the changes in a PR so we can test them locally (for example via yalc)?

packages/angular-test/src/boolean.ts
eneufeld: where is the setErrors action defined?

sdirix: It's just an action for the core reducer.

sdirix: Another thing: Don't we have `refresh` defined on the forms service. Couldn't we use that instead of `['updateSubject']`.

packages/angular-material/example/app/app.component.ts
sdirix: As it's only an example app this is not strictly necessary. However maybe it makes sense to also expose the readOnly setter in the forms service for users. In this case we should also adjust the example.

eneufeld: I add it to the new root component and thus also to the service

packages/angular-test/src/boolean.ts
eneufeld: I'll check

packages/angular/src/jsonforms-root.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        const newSchema = changes.schema;
+        if (newSchema && !newSchema.isFirstChange()) {
+            this.jsonformsService.setSchema(newSchema.currentValue);
+        }
+
+        const newUiSchema = changes.uischema;
+        if (newUiSchema && !newUiSchema.isFirstChange()) {
+            this.jsonformsService.setUiSchema(newUiSchema.currentValue);
+        }
```
sdirix, line 83: [In React](https://github.com/eclipsesource/jsonforms/blob/a51e3736ddc33a595887f29263dcb9242608a5d8/packages/react/src/JsonFormsContext.tsx#L124-L128) we use the `updateCore` action here as we didn't want to update (and trigger validation, emit events etc.) three separate times when `data`, `schema` and `uischema` change at the same time. Can we implement this same behavior here too?

sdirix (commented):
I had a look! Looks good but I also saw some things which I missed on the first review.

packages/angular-material/example/app/app.component.ts
```diff
@@ -57,65 +69,50 @@ import { map } from 'rxjs/operators';
...
   }
 
-  setReadonly() {
-    const uischema = getUiSchema(this.jsonformService.getState());
-    if (this.readonly) {
-      unsetReadonly(uischema);
-    } else {
-      setReadonly(uischema);
-    }
+  triggerReadonly() {
```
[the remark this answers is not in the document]
eneufeld: `toggle` is good, I don't want to make it `setReadonly` because I don't want to have logic in the template

packages/angular/src/jsonforms.service.ts
```diff
@@ -1,58 +1,106 @@
...
-    init(initialState: JsonFormsSubStates = {}) {
+    init(initialState: JsonFormsSubStates = { core: { data: undefined, schema: undefined, uischema: undefined } }) {
         this._state = initialState;
         this._state.config = configReducer(undefined, setConfig(this._state.config));
         this.state = new BehaviorSubject({ jsonforms: this._state });
+        this.updateCore(
+            Actions.init(
+                initialState.core.data,
+                initialState.core.schema,
+                initialState.core.uischema ?? generateDefaultUISchema(initialState.core.schema)
```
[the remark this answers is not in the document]
eneufeld: I'll do it in the service as the component is not used by the tests

packages/angular/src/jsonforms-root.component.ts
eneufeld: probably yes, I check

packages/angular/src/jsonforms.service.ts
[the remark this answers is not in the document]
sdirix: :+1: We only need to make sure that our update logic doesn't accidentally bring in an `undefined` schema or ui schema (for which the model was then generated) from the outside into the service, especially when using the optimized `updateCore` action.

packages/angular/src/jsonforms-root.component.ts
```diff
@@ -0,0 +1,112 @@
...
+            this.jsonformsService.updateCore(
+                Actions.updateCore(newData.currentValue, newSchema.currentValue, newUiSchema.currentValue)
+            );
+        } else if (newData && !newData.isFirstChange()) {
+            this.jsonformsService.setData(newData.currentValue);
+        } else if (newSchema && !newSchema.isFirstChange()) {
+            this.jsonformsService.setSchema(newSchema.currentValue);
+        } else if (newUiSchema && !newUiSchema.isFirstChange()) {
+            this.jsonformsService.setUiSchema(newUiSchema.currentValue);
+        }
```
sdirix, line 94: This will not update the schema when both data and schema change but not the ui schema.

I would do it similar to this:
```typescript
        if (newData || newSchema || newUiSchema) {
            const updateData = newData ? newData.currentValue : this.jsonformsService.$state.core.data;
            let updateSchema = newSchema ? newSchema.currentValue : this.jsonformsService.$state.core.schema;
            let updateUiSchema = newUiSchema ? newUiSchema.currentValue : this.jsonformsService.$state.core.uischema;
            if(updateSchema === undefined) { updateSchema = /* Generate schema */}
            if(updateUiSchema === undefined) { updateUiSchema = /* Generate ui schema */}
            this.jsonformsService.updateCore(
                Actions.updateCore(updateData, updateSchema, updateUiSchema)
            );
        }
```

packages/angular/src/jsonforms.service.ts
```diff
@@ -1,58 +1,107 @@
...
-    init(initialState: JsonFormsSubStates = {}) {
+    init(initialState: JsonFormsSubStates = { core: { data: undefined, schema: undefined, uischema: undefined } }) {
         this._state = initialState;
         this._state.config = configReducer(undefined, setConfig(this._state.config));
         this.state = new BehaviorSubject({ jsonforms: this._state });
+        this.updateCore(
+            Actions.init(
+                initialState.core.data,
+                initialState.core.schema ?? generateJsonSchema(initialState.core.data),
+                initialState.core.uischema ?? generateDefaultUISchema(initialState.core.schema)
```
sdirix, line 64: The generation of the ui schema needs to be done on the possibly generated schema. Otherwise the ui schema will be generated on an undefined schema when both schema and ui schema are undefined.

sdirix (changes requested):
Found two problems in the new code.

packages/angular/src/jsonforms-root.component.ts
```diff
@@ -0,0 +1,116 @@
...
+            if (newData && !newData.isFirstChange()) {
+                this.jsonformsService.setData(newData.currentValue);
+            }
+            if (newSchema && !newSchema.isFirstChange()) {
+                this.jsonformsService.setSchema(newSchema.currentValue);
+            }
+            if (newUiSchema && !newUiSchema.isFirstChange()) {
+                this.jsonformsService.setUiSchema(newUiSchema.currentValue);
+            }
+        }
```
sdirix, line 98: We don't optimize now for cases when only two of those attributes change. It could see this happen in cases where the user never specifies a ui schema but exchanges data and schema.

This is more of an edge case and only leads to a small performance hit instead of a misbehavior so we could leave it like this for now.

sdirix (approved):

packages/angular/src/jsonforms-root.component.ts
sdirix: Ah actually I missed a problem here: In the first case when all three props change and the new `schema` or `uiSchema` is undefined they will not be newly generated and the app will break.

If you don't want to go with the approach I originally suggested because we're then generating schema and or ui schema here then we actually need a new method in the forms service which we call which does it for us. Sadly the most obvious name `updateCore` is already used for the dispatching itself ;)

sdirix (approved):

sdirix (approved):
LGTM

PR #1658, opened 2020-12-01
fix: make NumberCell for Vanilla Renderer show 0

JovanShandro:
When using a default value of 0 for a number cell with vanilla renderers, the input is shown with an empty string. This is a quick fix for it.

### DIFF
--- a/packages/vanilla/src/cells/NumberCell.tsx
+++ b/packages/vanilla/src/cells/NumberCell.tsx
@@ -40,7 +40,7 @@ export const NumberCell = (props: CellProps & VanillaRendererProps) => {
     <input
       type='number'
       step='0.1'
-      value={data || ''}
+      value={data ?? ''}
       onChange={ev => handleChange(path, Number(ev.target.value))}
       className={className}
       id={id}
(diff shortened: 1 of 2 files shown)


packages/vanilla/src/cells/NumberCell.tsx
```diff
@@ -36,11 +36,13 @@ import { withVanillaCellProps } from '../util/index';
...
   const { data, className, id, enabled, uischema, path, handleChange } = props;
 
+  const value = data !== undefined ? data : '';
+
   return (
     <input
       type='number'
       step='0.1'
-      value={data || ''}
+      value={value}
```
sdirix, line 45:
```suggestion
      value={data ?? ''}
```

sdirix:
Thank you very much for the contribution! Could you also add a testcase?

sdirix (changes requested):
A smaller fix could be to replace `data || ''` with `data ?? ''`, right? If you agree please adapt this PR.

packages/vanilla/src/cells/NumberCell.tsx
JovanShandro: True, thanks! I am making the change now.

JovanShandro: Also I would love to add a testcase, but for the moment I am not familiar with how enzyme works, or testing for react components in general, so it might take me some time :sweat_smile:

sdirix (approved):

PR #1661, opened 2020-12-03
Add support to provide config to the root

eneufeld:
- add config input
- use two way databinding in example for data


packages/angular-material/example/app/app.component.ts
```diff
@@ -71,14 +71,14 @@ const itemTester: UISchemaTester = (_schema, schemaPath, _path) => {
       </button>
     </div>
     <jsonforms
-      [data]="selectedExample.data"
+      [(data)]="selectedExample.data"
```
sdirix, line 74: Do we want two-way binding here?

packages/angular/src/jsonforms-root.component.ts
```diff
@@ -63,7 +64,9 @@ export class JsonForms implements OnChanges, OnInit {
             },
             uischemas: this.uischemas,
             i18n: { locale: this.locale, localizedSchemas: undefined, localizedUISchemas: undefined },
-            renderers: this.renderers
+            renderers: this.renderers,
+            config: this.config,
+            readonly: this.readonly
```
sdirix, line 69: good catch

sdirix (approved):
LGTM

packages/angular-material/example/app/app.component.ts
eneufeld: yes, to showcase that the data is updated, we show this value in a div above the form

PR #1662, opened 2020-12-07
Improve dataChange emit

eneufeld:
The change emit is currently firing twice.
By checking for changed data the emit is only done once.


sdirix (approved):
Makes sense to me. Should we also output the errors like we do in React?

eneufeld:
> Makes sense to me. Should we also output the errors like we do in React?

yes that would make sense, I will create a new PR for this

Issue #1668, opened 2020-12-15
NumberControlRenderer does not provide a way to explicitly disable grouping


ttntrifork:
**Is your feature request related to a problem? Please describe.**
Sometimes you want a number input to not use grouping even if your locale has a grouping character (e.g. da-DK).  
This is useful when presenting inputs for numerical "codes" in some code system, e.g. the Danish "Central Business Register" which assigns a 8 digit numerical code to each business registered (which does NOT contain grouping characters). 

jsonforms/angular-material `NumberControlRenderer` does not set a `useGrouping` property in the `NumberFormatOptions` object passed to the `Intl.NumberFormat` constructor, which causes Intl.NumberFormat to use the locale specific grouping.

**Describe the solution you'd like**
A uischema options element that sets useGrouping to false. 
See PR https://github.com/eclipsesource/jsonforms/pull/1669

**Describe for which setup you like to have the improvement**
Framework: angular
RendererSet: material

sdirix:
Thank you for reporting this issue!

PR #1669, opened 2020-12-15
Angular-material: NumberControlRenderer now sets the 'useGrouping' property

ttntrifork:
NumberControlRenderer now sets the 'useGrouping' property in the NumberFormatOptions object passed to the Intl.NumberFormat constructor to the value of the uischema options element `useGrouping`, if defined.

See https://github.com/eclipsesource/jsonforms/issues/1668

**Note**: I cannot run the `npm run dev` in packages\material on my Windows OS. There are syntax errors.  
The change is tested by copy pasting the NumberControlRenderer code and creating it as a custom renderer in my project that depends on jsonforms.

It works when setting the options in `uischema`, e.g.
```json
                {
                  "type": "Control",
                  "scope": "#/properties/cvr",
                  "label": "CVR number",
                  "options": {
                    "useGrouping": false
                  }
                }
```

### DIFF
--- a/packages/angular-material/src/controls/number.renderer.ts
+++ b/packages/angular-material/src/controls/number.renderer.ts
@@ -135,6 +135,7 @@ export class NumberControlRenderer extends JsonFormsControl {
       if (this.locale === undefined || this.locale !== currentLocale) {
         this.locale = currentLocale;
         this.numberFormat = new Intl.NumberFormat(this.locale, {
+          useGrouping: this.uischema.options ? this.uischema.options.useGrouping : undefined,
           maximumFractionDigits: this.MAXIMUM_FRACTIONAL_DIGITS
         });
         this.determineDecimalSeparator();


sdirix:
Thank you for the contribution!

packages/angular-material/src/controls/number.renderer.ts
```diff
@@ -135,6 +135,7 @@ export class NumberControlRenderer extends JsonFormsControl {
       if (this.locale === undefined || this.locale !== currentLocale) {
         this.locale = currentLocale;
         this.numberFormat = new Intl.NumberFormat(this.locale, {
+          useGrouping: this.uischema.options ? this.uischema.options.useGrouping : undefined,
```
sdirix, line 138: This is fine in general but should be checked with the `config` object as it serves as a default.

sdirix (changes requested):
Looks mostly fine in code. Would be great to also have a testcase for this.

packages/angular-material/src/controls/number.renderer.ts
sdirix: See for example the [MaterialListWithDetailRenderer](https://github.com/eclipsesource/jsonforms/blob/cd457e1af2a7515263661c833fd2e833111a194a/packages/material/src/additional/MaterialListWithDetailRenderer.tsx#L98)

ttntrifork: Thanks. I see your requested changes have already been implemented on #1672

ttntrifork:
I have a hard time helping out with tests as I don't think your build system supports Windows (or am I doing something wrong?). 

I see you have created a new PR #1672 based on my suggestion. Should I just close this PR then?

sdirix:
Hi @ttntrifork, sadly our build scripts are incompatible with Windows at the moment, however I think with WSL could help. Yes we created a PR based on this one, so feel free to close this PR if you'd like to. Thanks again for your contribution! Happy new year!

PR #1675, opened 2021-01-14
Vanilla styles - Automatically use default styles, remove redux style support, support style def merging

lucas-koehler:
## Apply default vanilla styles if none are explicitly specified
* Set vanillaStyles as default value in JsonFormsStyleContext
* Move vanillaStyles, StyleDef interface and the styles context to new module styles
* Adapt documentation

## Remove vanilla style support via redux store
Vanilla renderer style definitions can no longer be provided via the redux store.
Instead, the plain react JsonFormsStyleContext needs to be used.

* Adapt style context
* Adapt tests
* Extend migration documentation

## Merge vanilla styles with same id
* adapt findStyle method
* extend Styles documentation

### DIFF
--- a/packages/vanilla/Styles.md
+++ b/packages/vanilla/Styles.md
@@ -79,3 +79,32 @@ const styleContextValue = { styles: [
   />
 </JsonFormsStyleContext.Provider>
 ```
+
+You can also extend the existing default styles.
+Thereby, the existing style classes as well as your custom ones will be applied.
+This is the case because all style definitions for an ID are merged.
+
+```typescript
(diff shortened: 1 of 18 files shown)


packages/vanilla/Styles.md
```diff
@@ -77,3 +79,33 @@ const styleContextValue = { styles: [
   />
 </JsonFormsStyleContext.Provider>
 ```
+
+You can also extend the existing default styles.
+Thereby, the existing style classes as well as your custom ones will be applied.
+This is the case, because all style definitions for an ID are merged.
+Add the default styles to your custom style definition by using the spread operator.
```
sdirix, line 86:
```suggestion
```

sdirix (approved):
LGTM

PR #1693, opened 2021-02-15
Add material boolean toggle control and cell

lucas-koehler:
In addition to the default checkbox renderers for boolean controls and cells,
provide a control and cell rendering a boolean as a toggle.

* Add MuiToggle wrapping a Material UI Switch
* Add material boolean toggle control and cell that can be activated by UISchema option `toggle`
* Add unit tests for control and cell
* Add boolean toggle example to material examples showcasing a checkbox and toggle side by side.
---
New example:
image

### DIFF
--- a/packages/examples/src/control-options.ts
+++ b/packages/examples/src/control-options.ts
@@ -1,7 +1,7 @@
 /*
   The MIT License
   
-  Copyright (c) 2017-2019 EclipseSource Munich
+  Copyright (c) 2017-2021 EclipseSource Munich
   https://github.com/eclipsesource/jsonforms
   
   Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -137,6 +137,10 @@ export const extendedSchema = {
     hideRequiredAsterisk: {
       type: 'string',
       description: 'Hides the "*" symbol, when the field is required',
+    },
+    toggle: {
+      type: 'boolean',
+      description: 'The "toggle" option renders boolean values as a toggle.'
     }
   },
   required: ['hideRequiredAsterisk', 'restrictText']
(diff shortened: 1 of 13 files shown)


sdirix (changes requested):
Works great! Thanks!

I only have minor nitpicks. Also please add the toggle boolean to the `control` example.

lucas-koehler:
@sdirix I made the boolean toggle a separate example and added a boolean checkbox and toggle to the control options example.

packages/examples/src/control-options.ts
```diff
@@ -137,6 +137,14 @@ export const extendedSchema = {
     hideRequiredAsterisk: {
       type: 'string',
       description: 'Hides the "*" symbol, when the field is required',
+    },
+    checkbox: {
+      type: 'boolean',
+      description: 'Checkbox is the default control for boolean values.'
+    },
```
sdirix, line 144: We already have a `boolean`/`checkbox` in the first `schema`. No need to re-add it here in `extendedSchema`.

lucas-koehler: That's true but I'm concerned this might break the jsonforms website. There we use the base example and the one with options separately. Thus, I imagine we only use the `extentedData` and `extendedSchema` there. But I can also remove it.

sdirix: That's true :) I meant to remove both, the `checkbox` property and its `Control`. We don't need to show the normal checkbox in the `extendedSchema` and `extendedUischema` as its already a part of the normal `schema` and `uischema`.

lucas-koehler: I see, that makes sense. I'll remove it :)

lucas-koehler:
@sdirix  I removed the duplicate normal checkbox from the control options example and integrated it in the original commit.

sdirix (approved):
LGTM!

PR #1700, opened 2021-02-22
Offer MuiInputText inputComponent customization

sdirix:
Also export MaterialInputControl for more convenient reuse.


packages/material/src/controls/MaterialInputControl.tsx
```diff
@@ -36,7 +36,7 @@ import { Hidden, InputLabel } from '@material-ui/core';
 import { FormControl, FormHelperText } from '@material-ui/core';
 import merge from 'lodash/merge';
 
-interface WithInput {
+export interface WithInput {
   input: any;
```
AlexandraBuzila, line 40: Should we type this?

sdirix: In general yes, but I would not change this now to avoid typing problems in user code.

AlexandraBuzila (approved):

PR #1701, opened 2021-02-23
Fixes group's label being reformatted with startCase

sifthedog:


sdirix (commented):
This certainly improves the current situation but we probably shouldn't change the label unconditionally, i.e. this still overrides the label which is manually set on a group detail.

sdirix (approved):

sdirix:
I opened a follow up: #1712

sifthedog:
@sdirix need a hand with the follow up? Let me know if I can help in any way.

sdirix:
> @sdirix need a hand with the follow up? Let me know if I can help in any way.

Sure that would be great ;) The main problem here is that `findUISchema` doesn't indicate whether it generated or found the ui schema. In the latter case we should not modify it. Also any change should be backwards compatible.

PR #1703, opened 2021-02-25
Add Vue 3 Vanilla renderers

sdirix:
Adds primitive control renderers, basic layout renderers as well as an
array list renderer based on Vanilla HTML5.

CSS classes are completely customizable programmatically as well as via
the ui schema to allow straightforward integration with CSS libraries
like Tailwind.

Complimentary changes:

The Vue 'config' object was only top-level reactive. Now JSON Forms
will also react to nested changes.

Export the whole content of Vue `jsonFormsCompositions` to make them
reusable by the vanilla-renderers as well as any custom library.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,14 +1,19 @@
 {
   "lerna": "2.5.1",
   "packages": [
-    "packages/*"
+    "packages/angular",
+    "packages/angular-material",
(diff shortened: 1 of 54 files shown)


jfaltermeier (approved):
LGTM

PR #1709, opened 2021-03-01
Add MultiEnum renderer for enums and oneOf

AlexandraBuzila:
Render an anyOf array of consts as a list of checkboxes. The checkbox is
checked if the array item it represents is present in the data.

### DIFF
--- a/packages/material/src/index.ts
+++ b/packages/material/src/index.ts
@@ -36,7 +36,9 @@ import {
   materialObjectControlTester,
   MaterialObjectRenderer,
   materialOneOfControlTester,
-  MaterialOneOfRenderer
+  MaterialOneOfRenderer,
+  MaterialEnumArrayRenderer,
+  materialEnumArrayRendererTester
 } from './complex';
 import {
   MaterialLabelRenderer,
@@ -171,6 +173,10 @@ export const materialRenderers: JsonFormsRendererRegistryEntry[] = [
   {
     tester: materialAnyOfStringOrEnumControlTester,
     renderer: MaterialAnyOfStringOrEnumControl
+  },
+  {
+    tester: materialEnumArrayRendererTester,
+    renderer: MaterialEnumArrayRenderer
   }
 ];
 
--- a/packages/core/src/util/renderer.ts
+++ b/packages/core/src/util/renderer.ts
@@ -187,7 +187,7 @@ export const enumToEnumOptionMapper = (e: any): EnumOption => {
 export const oneOfToEnumOptionMapper = (e: any): EnumOption => ({
   value: e.const,
   label:
-    e.title || (typeof e.const === 'string' ? e.const : JSON.stringify(e.const))
+    e.title ?? (typeof e.const === 'string' ? e.const : JSON.stringify(e.const))
 });
 
 export interface OwnPropsOfRenderer {
--- a/packages/core/src/util/renderer.ts
+++ b/packages/core/src/util/renderer.ts
@@ -184,6 +184,12 @@ export const enumToEnumOptionMapper = (e: any): EnumOption => {
   return { label: stringifiedEnum, value: e };
 };
 
+export const oneOfToEnumOptionMapper = (e: any): EnumOption => ({
+  value: e.const,
+  label:
+    e.title || (typeof e.const === 'string' ? e.const : JSON.stringify(e.const))
+});
+
 export interface OwnPropsOfRenderer {
   /**
    * The UI schema to be rendered.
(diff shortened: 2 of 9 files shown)


packages/material/src/complex/MaterialEnumArrayControl.tsx
```diff
@@ -0,0 +1,115 @@
...
+        schema => hasType(schema, 'array') && !Array.isArray(schema.items)
+      ),
+      schemaSubPathMatches('items', schema => {
+        return schema.anyOf !== undefined && schema.anyOf.length > 0;
+      })
+    )
+  )
+);
+
+export default MaterialEnumArrayControl;
```
sdirix, line 115: To be consistent with other controls we should default export the wrapped version

packages/material/src/index.ts
```diff
@@ -171,6 +174,10 @@ export const materialRenderers: JsonFormsRendererRegistryEntry[] = [
   {
     tester: materialAnyOfStringOrEnumControlTester,
     renderer: MaterialAnyOfStringOrEnumControl
+  },
+  {
+    tester: materialEnumArrayControlTester,
+    renderer: withJsonFormsControlProps(MaterialEnumArrayControl)
```
sdirix, line 180: Don't wrap here, but use the already wrapped imported version

packages/material/src/complex/MaterialEnumArrayControl.tsx
```diff
@@ -0,0 +1,115 @@
...
+  return (
+    <Hidden xlUp={!visible}>
+      <FormControl component='fieldset'>
+        <FormGroup row>
+          {uiData.map((value: any, index: number) => {
+            const option = options[index];
+            const optionPath = Paths.compose(path, `${index}`);
+            return (
+              <FormControlLabel
+                id={option.const}
```
sdirix, line 70: Careful: The value of `const` can be anything! It doesn't need to be a string, but could for example be an `object`, `array`, etc. Please see `mapStateToOneOfEnumProps` where we already had to handle the case.

packages/material/src/complex/MaterialEnumArrayControl.tsx
```diff
@@ -0,0 +1,115 @@
...
+      // option was added
+      newData.push(option);
+    } else {
+      // option was removed
+      const indexInData = newData.indexOf(option);
+      newData.splice(indexInData, 1);
+    }
+
+    handleChange(path, newData);
+  };
```
sdirix, line 56: If you would use `withJsonFormsArrayControlProps` you would get functions to add and remove items to the array for free, i.e. `addItem`, `removeItems`, `moveUp` and `moveDown` would get injected. `schema` would also already point to the `items` object.

sdirix (changes requested):
In general I like it very much. However there are some things I would like to see differently.

Currently the schema tested for looks like this
```js
"property": {
  "type": "array",
  "items": {
    "anyOf": [
      { },
    ]
  },
},
 ```
 which doesn't really fit to the use case:

- The schema allows the same entry multiple times, but our control only handles them uniquely (the control therefore misbehaves if the same entry is there multiple times)
- `anyOf` stands for that **at least** one item matches. However we're handling it like "enums", which are unique. So it should be `oneOf`.
- To be consistent with our other `oneOfEnum` controls we should check whether the `items` are `oneOf` and contain a `const` in each entry.
- Ideally we should also be able to handle `enum` content

I would prefer to support the following schemas:
```js
"property": {
  "type": "array",
  "items": {
    "oneOf": [
      {
        "const": <any>,
        "title": "mylabel"
      },
    ]
  },
  "uniqueItems": true
},
```
and
```js
"property": {
  "type": "array",
  "items": {
    "type": "string",
    "enum": ["a", "b", "c"]
  },
  "uniqueItems": true
},
```

So we check for

 - type array with unique items
 - the items are either `enum` or `oneOf` containing `const`
 
 ---
 
 Please also add an example to the example app.

sdirix (changes requested):
Additionally it would be great when we don't add all the logic into the renderer. Instead I would prefer a new `mapStateToMultiEnumProps` and `mapDispatchToMultiEnumProps`.

- `mapStateToMultiEnumProps`: extracting the `items` options from `oneOf` or `enum`, similar to `mapStateToEnumPros` and `mapStateToOneOfEnumProps`
- `mapDispatchToMultiEnumProps`: providing a `addItem` and `removeItem` function which can be used for toggling. Using this we can also remove the `useRef` workaround

By extracting the logic from the renderer we can easily write a multi-select enum with a different renderer (e.g. Material UI's multi select dropdown) as well as more easily add the same renderer in the other renderer sets too.

AlexandraBuzila:
I updated the commit to include you comments, Stefan. The renderer now supports both `oneOf` and `enum` items.

sdirix (changes requested):
Looks good!

sdirix (approved):

PR #1714, opened 2021-03-12
Improve error filtering for anyOf and oneOf

lucas-koehler:
- Do not filter object level validations (e.g. required) in anyOf/oneOf.
- Filter anyOf/oneOf/additionalProperties errors
- Extend core tests

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -353,15 +353,16 @@ export const errorsAt = (
     }
 
     let result = matchPath(error.dataPath);
-    // In anyOf and oneOf blocks with "primitive subschemas" (not defining an object),
+    // In anyOf and oneOf blocks with "primitive" (i.e. string, number etc.) or array subschemas,
     // we want to make sure that errors are only shown for the correct subschema.
     // Therefore, we compare the error's parent schema with the property's schema.
-    // This is necessary because in the primitive case the error's data path is the same for all subschemas:
+    // In the primitive case the error's data path is the same for all subschemas:
     // It directly points to the property defining the anyOf/oneOf.
+    // The same holds true for errors on the array level (e.g. min item amount).
     // In contrast, this comparison must not be done for errors whose parent schema defines an object
     // because the parent schema can never match the property schema (e.g. for 'required' checks).
     const parentSchema: JsonSchema | undefined = error.parentSchema;
-    if (result && parentSchema?.type !== 'object'
+    if (result && !isObjectSchema(parentSchema)
       && combinatorPaths.findIndex(p => instancePath.startsWith(p)) !== -1) {
       result = result && isEqual(parentSchema, schema);
     }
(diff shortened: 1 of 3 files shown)


packages/core/src/reducers/core.ts
```diff
@@ -340,20 +340,47 @@ export const errorsAt = (
...
 
+/**
+ * The error-type of an AJV error is defined by its `keyword` property.
+ * Certain errors are filtered because they don't fit to any rendered control.
+ * All of them have in common that we don't want to show them in the UI
+ * because controls will show the actual reason why they don't match their correponding sub schema.
+ * - additionalProperties: Indicates that a property is present that is not defined in the schema.
+ *      Jsonforms only allows to edit defined properties. These errors occur if an oneOf doesn't match.
+ * - anyOf: Indicates that an anyOf definition itself is not valid because none of its subschemas matches.
+ * - oneOf: Indicates that an oneOf definition itself is not valid because not exactly one of its subschemas matches.
```
sdirix, line 380: Are there no `allOf` errors we want to filter?

packages/core/src/reducers/core.ts
```diff
@@ -340,20 +340,47 @@ export const errorsAt = (
...
-      result = result && isEqual(error.parentSchema, schema);
+    // In anyOf and oneOf blocks with "primitive subschemas" (not defining an object),
+    // we want to make sure that errors are only shown for the correct subschema.
+    // Therefore, we compare the error's parent schema with the property's schema.
+    // This is necessary because in the primitive case the error's data path is the same for all subschemas:
+    // It directly points to the property defining the anyOf/oneOf.
+    // In contrast, this comparison must not be done for errors whose parent schema defines an object
+    // because the parent schema can never match the property schema (e.g. for 'required' checks).
+    const parentSchema: JsonSchema | undefined = error.parentSchema;
+    if (result && parentSchema?.type !== 'object'
```
sdirix, line 364: There is also `type='array'` which is not a primitive. Also a schema could only define `properties` and therefore implicitly be an `object`.

packages/core/src/reducers/core.ts
```diff
@@ -340,20 +340,47 @@ export const errorsAt = (
...
   return filter(errors, error => {
+    // Filter errors that match any keyword that we don't want to show in the UI
+    if (filteredErrorKeywords.indexOf(error.keyword) !== -1) {
+      return false;
+    }
+
     let result = matchPath(error.dataPath);
-    if (combinatorPaths.findIndex(p => instancePath.startsWith(p)) !== -1) {
-      result = result && isEqual(error.parentSchema, schema);
+    // In anyOf and oneOf blocks with "primitive subschemas" (not defining an object),
```
sdirix, line 356:
```suggestion
    // In anyOf and oneOf with "primitive subschemas" (i.e. string, number etc.),
```

packages/core/src/reducers/core.ts
lucas-koehler: It did not came up in this context but yes I think this makes sense. I added it.

packages/core/src/reducers/core.ts
lucas-koehler: Good catch. I changed this to check for this in a separate method to match object schemas. We still want this check for array schemas in case one/anyOf subschemas use arrays. We only want to match array-level validations (e.g. minItems) for the corresponding sub schema.

sdirix (approved):
LGTM Thanks!

PR #1727, opened 2021-03-29
update: vue2-vanilla json:

Felix-Puetz:
provide the custom styles example as actual json instead of js-code.

I'n my opinion I think its better to provide the example as a json and not as pseudo json in js-code since thats what are are actually working with


sdirix (changes requested):
Fine for me. Could you also update the Vue 3 Readme so they are kept in sync?

sdirix (approved):

PR #1730, opened 2021-03-31
Add Vue Vanilla default input style

sdirix:


sdirix:
@TheZoker Could you please double check whether other styles are referenced in one of the renderers but are missing in the defaultStyles?

PR #1731, opened 2021-03-31
Set Vue Vanilla button types

sdirix:
Buttons contained in forms are by default of type 'submit'. We now
explicitely set the ArrayListRenderer's buttons to type 'button' to
avoid submitting forms in case that they are rendered as part of a form.


sdirix:
@TheZoker Please add a test case to check whether all buttons rendered by this renderer are of type `button`.

TheZoker (approved):
LGTM :+1: 

I can confirm, that with this change the buttons no longer trigger the form to submit on click

PR #1732, opened 2021-03-31
Set placeholders in Vue Vanilla

sdirix:
Set placeholder attribute in Vue Vanilla controls when specified via
the ui schema.


sdirix:
@TheZoker Please add test cases

PR #1733, opened 2021-03-31
Improve Vue Layout renderer reactivity and Vue HTML output

sdirix:
Refactor the composition architecture to be based on 'computed' instead
of 'ref'. This avoids manually syncing ref content and fixes
rerendering issues for the layout renderers.

Make sure that only expected props are handed over to renderers
which unconditionally bind all properties. This avoids rendering
stringified attributes like rootSchema in the HTML.


TheZoker (approved):
Can confirm that now all uischema changes happen instantly without the need to call the update function multiple times :+1:

PR #1736, opened 2021-04-14
Empty FormHelperText should be visible

AlexandraBuzila:
The FormHelperText should be visible even when empty, to allow it to be styled.


AlexandraBuzila:
@sdirix only the `MaterialEnumArrayRenderer` required a fix, the helper texts from all the other renderers behave correctly. This doesn't change the UI in any way, unless there's an explicit style override for `MuiFormHelperText`

sdirix (approved):

PR #1746, opened 2021-05-03
package-lock.json changes done by npm run init

nene:
I followed the steps in README:

- cloned the repository
- `npm ci`
- `npm run init`

Running `init` produced these changes in `package-lock.json`. My guess is that one has simply forgotten to update the lockfile after changes in some dependencies. Though I might be wrong and the problem could be elsewhere.


sdirix:
Hi @nene! Thanks for the PR. The problem here is that the combination of `npm (node 12) + lerna + hoisting` produces an inconsistent `package-lock.json`, i.e. no matter how often it is generated it switches between two-three different states which are largely equivalent (sometimes marking a package as a devDependency and sometimes not).

There is currently no fix for that that I'm aware of.

nene:
Alright. After re-running the install and init with node 12 I no more get the diff in package-lock.json. So at least the problem went away for me for the time being.

PR #1749, opened 2021-05-10
Indicate compatibility with Angular 10 and 11

sdirix:


sdirix:
> ddi you check, that the flex-layout beta release is resolved correctly?

You're right, the `beta` releases were not covered correctly. This is now fixed.

eneufeld (approved):
thank you for fixing

PR #1762, opened 2021-06-10
Remove json-schema-ref-parser

sdirix:
Removes json-schema-ref-parser and corresponding resolving functionality
and dependencies from JSON Forms.

Resolving of external dependencies is not a core feature of JSON Forms
which is why we delegated it to a 3rd party dependency. However the
json-schema-ref-parser library is not only slow and mutating schemas in
place but also brings in a number of questionable Node dependent
dependencies.

The only binding actually using this functionality is the React
binding. The way resolving was implemented resulted in the same flow
as resolving it outside of JSON Forms and handing the
resolved schema over for all but the very niche use case of unresolved
schemas in the ui schema registry. Even that case can still be handled
using a custom renderer.

Based on these downsides and the very little upsides we decided to
remove json-schema-ref-parser without a replacement from JSON Forms.

Angular and Vue consumers don't need to spend any migration efforts. In
@jsonforms/react we still export the backward API compatible
'ResolvedJsonFormsDispatch'. Therefore only React consumers actually
using the now removed `refParserOptions` are forced to migrate while
for all other React consumers it depends on their use cases.

The built-in resolving mechanism in JSON Forms is quite basic. When
the schema was fully resolved anyway the basic resolving did quite
fine. However as this is no longer the case by default, there are
now cases which are no longer supported and break. This is especially
the case for complex combinator use cases involving references.

Of course these can still be used when the consumer resolves their
schema themselves before handing it over to JSON Forms. We also plan
to make our built-in resolving more robust before the full 3.0 release
to cover more of these cases. However we will not reach feature parity
without major additional efforts.


packages/material/test/renderers/MaterialOneOfRenderer.test.tsx
```diff
@@ -303,7 +303,7 @@ describe('Material oneOf renderer', () => {
     });
   });
 
-  it('should add an item within an array', async () => {
+  it.skip('should add an item within an array', async () => {
```
eneufeld, line 306: why skip all these tests?

eneufeld:
I tested the examples, the AnyOf, OneOf, AllOf Resolve example (/packages/examples/src/anyOf-oneOf-allOf-resolve.ts) is crashing.
I also think that the packages/examples/src/resolve.ts is not useful.
The `name: 'issue-1253-add-button-empty-row', ` example crashes.

I didn't test which of these crashed before.

eneufeld (approved):
As discussed, we should update the resolver code and the examples in a follow up commit.

Issue #1735, opened 2021-04-07
Support readonly defined in JSON Schema and UI Schema


sdirix:
**Feature**
- JSON Schema specifies a [`readOnly` attribute](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.9.4)
- The Angular renderers support `options: { readonly: true } `

We should support both of these in the JSON Forms Core, so that in all renderer sets `enabled` respects not only the state-global-`readonly` and rules, but also the JSON Schema and UI Schema settings.

**Describe for which setup you like to have the improvement**
Framework: [core]

sdirix:
## Suggestion

Determine the `enabled` flag by following these steps.
Whenever `enabled` is determined, skip the remaining steps.

1. If the form-wide `readonly` flag is `true` => `enabled: false`
2. If there is an `ENABLE` or `DISABLE` rule for the current ui schema element => set `enabled` accordingly
3. Check if `uischema.options.readonly` is set. If yes, set `enabled` accordingly
4. Check if `uischema.options.readOnly` is set. If yes, set `enabled` accordingly
5. Check if `config.readonly` is set. If yes, set `enabled` accordingly
6. Check if `config.readOnly` is set. If yes, set `enabled` accordingly
7. Check if `schema.readOnly` is set to `true`. If yes set `enabled: false`
8. Set `enabled` according to the parent's `enabled` (i.e. check `ownProps.enabled`).

These modifications need to be part of JSON Forms core. For example in `mapStateToControlProps`, but also all other `mapStateToXZY`. At the moment the core only checks for case 1, 2 and 8. See [here](https://github.com/eclipsesource/jsonforms/blob/d975bc7d5d6807c1e6352c6641458aadd709d436/packages/core/src/util/renderer.ts#L410-L414).

There should be unit tests.
There should be a new example in [examples](https://github.com/eclipsesource/jsonforms/tree/master/packages/examples/src), covering all cases (except case 1).

The Angular Material renderers also need to be modified. They currently check themselves for case 3. This is then no longer necessary. They should just take the determined value from the core.

sdirix:
Implemented with #1763

PR #1763, opened 2021-06-16
individual property readonly support

onerzafer:
Hello I'm using your awsome library and I needed read-only support for individual properties then I found a related issue https://github.com/eclipsesource/jsonforms/issues/1735 

I've tried to add the feature. If you can give some feedback and point me in the correct direction I would be happy.

### DIFF
--- a/packages/core/src/util/cell.ts
+++ b/packages/core/src/util/cell.ts
@@ -40,7 +40,7 @@ import {
   DispatchPropsOfControl,
   EnumOption,
   enumToEnumOptionMapper,
-  mapDispatchToControlProps,
+  mapDispatchToControlProps
 } from './renderer';
 import { JsonFormsState } from '../store';
 import { JsonFormsCellRendererRegistryEntry } from '../reducers/cells';
@@ -111,7 +111,8 @@ export const mapStateToCellProps = (
       : isVisible(uischema, rootData, undefined, getAjv(state));
   const readonly = state.jsonforms.readonly;
   const enabled =
-    !readonly && (ownProps.enabled !== undefined
+    !readonly &&
+    (ownProps.enabled !== undefined
       ? ownProps.enabled
       : isEnabled(uischema, rootData, undefined, getAjv(state)));
   const errors = formatErrorMessage(
@@ -166,7 +167,7 @@ export const defaultMapStateToEnumCellProps = (
   const options: EnumOption[] =
     ownProps.options ||
     props.schema.enum?.map(enumToEnumOptionMapper) ||
-    props.schema.const && [enumToEnumOptionMapper(props.schema.const)];
+    (props.schema.const && [enumToEnumOptionMapper(props.schema.const)]);
   return {
     ...props,
     options
--- a/packages/core/src/util/cell.ts
+++ b/packages/core/src/util/cell.ts
@@ -26,25 +26,19 @@ import isEmpty from 'lodash/isEmpty';
 import union from 'lodash/union';
 import { getConfig, getData, getErrorAt, getSchema, getAjv } from '../reducers';
 import {
-  AnyAction,
-  Dispatch,
-  formatErrorMessage,
-  isEnabled,
-  isVisible,
-  OwnPropsOfControl,
-  OwnPropsOfEnum,
-  Resolve,
-  StatePropsOfScopedRenderer
-} from '.';
-import {
   DispatchPropsOfControl,
   EnumOption,
   enumToEnumOptionMapper,
-  mapDispatchToControlProps
+  mapDispatchToControlProps,
+  OwnPropsOfControl,
+  OwnPropsOfEnum,
+  StatePropsOfScopedRenderer
 } from './renderer';
 import { JsonFormsState } from '../store';
 import { JsonFormsCellRendererRegistryEntry } from '../reducers/cells';
-import { JsonSchema } from '..';
+import { JsonSchema } from '../models/jsonSchema';
+import { isInherentlyEnabled, isVisible } from './runtime';
+import { AnyAction, Dispatch, formatErrorMessage, Resolve } from '.';
 
 export { JsonFormsCellRendererRegistryEntry };
 
@@ -109,17 +103,38 @@ export const mapStateToCellProps = (
     ownProps.visible !== undefined
       ? ownProps.visible
       : isVisible(uischema, rootData, undefined, getAjv(state));
-  const readonly = state.jsonforms.readonly;
-  const enabled =
-    !readonly &&
-    (ownProps.enabled !== undefined
-      ? ownProps.enabled
-      : isEnabled(uischema, rootData, undefined, getAjv(state)));
+
+  const rootSchema = getSchema(state);
+  const config = getConfig(state);
+
+  /* When determining the enabled state of cells we take a shortcut: At the
+   * moment it's only possible to configure enablement and disablement at the
+   * control level. Therefore the renderer using the cell, for example a 
+   * table renderer, determines whether a cell is enabled and should hand
+   * over the prop themselves. If that prop was given, we prefer it over
+   * anything else to save evaluation effort (except for the global readonly
+   * flag). For example it would be quite expensive to evaluate the same ui schema
+   * rule again and again for each cell of a table. */
+  let enabled;
+  if (state.jsonforms.readonly === true) {
+    enabled = false;
+  } else if (typeof ownProps.enabled === 'boolean') {
+    enabled = ownProps.enabled;
+  } else {
+    enabled = isInherentlyEnabled(
+      state,
+      ownProps,
+      uischema,
+      schema || rootSchema,
+      rootData,
+      config
+    );
+  }
+
   const errors = formatErrorMessage(
     union(getErrorAt(path, schema)(state).map(error => error.message))
   );
   const isValid = isEmpty(errors);
-  const rootSchema = getSchema(state);
 
   return {
     data: Resolve.data(rootData, path),
--- a/packages/core/src/util/runtime.ts
+++ b/packages/core/src/util/runtime.ts
@@ -181,11 +181,16 @@ export const isEnabled = (
   return true;
 };
 
+/**
+ * Indicates whether the given `uischema` element shall be enabled or disabled.
+ * Checks the global readonly flag, uischema rule, uischema options (including the config),
+ * the schema and the enablement indicator of the parent.
+ */
 export const isInherentlyEnabled = (
   state: JsonFormsState,
   ownProps: any,
   uischema: UISchemaElement,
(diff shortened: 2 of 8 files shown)


sdirix:
Hi @onerzafer, thank you very much for the contribution! I'll take a look.

packages/core/src/util/runtime.ts
```diff
@@ -177,3 +180,22 @@ export const isEnabled = (
...
+  const readonly = [state.jsonforms, uischema.options, schema, config].some(
+    (prop: any) => prop?.readonly || prop?.readOnly
+  );
+  const enabled: boolean =
+    !readonly &&
+    (ownProps.enabled === undefined || hasEnableRule(uischema)
+      ? isEnabled(uischema, rootData, ownProps.path, getAjv(state))
+      : ownProps.enabled);
+  return enabled;
+};
```
sdirix, line 201: I would prefer when this determination would exactly follow the algorithm presented in #1735, i.e.

```ts
if (state.readonly) {
  return false
}
if (hasEnableRule(uischema)) {
  return isEnabled(uischema, rootData, ownProps.path, getAjv(state));
}
if (typeof uischema.options.readonly === 'boolean') {
  return uischema.options.readonly
}
// ...
if (typeof ownProps.enabled === 'boolean) {
  return ownProps.enabled;
}
return true;
```

While the code might not look nice, the determination can be very easily followed ;)
The idea of this algorithm is not to `||` all (most) potential readonlys but to stop once a value for it was given on purpose. 

Is this fine for you or do you need to `OR` everything for your use cases?

sdirix (changes requested):
In general the code is fine :) Thank you again!

I would prefer to adhere to the algorithm presented in #1735, see the respective comment. I'm open for other suggestions, so if you prefer a different algorithm please elaborate why.

Also we definitely need test cases for the `isInherentlyEnabled` and ideally an `example` in the `examples` package which showcases all of them.

packages/core/src/util/runtime.ts
onerzafer: got it will update it accordingly. I assumed OR would do :)

onerzafer:
Hello, @sdirix I've noticed that there are a couple of tests that are failing some packages. I tried to run them locally but somehow I couldn't manage it. Probably I couldn't install all the packages (test runner is complaining about missing packages etc.) I've read the developer docs also and followed the steps yet we are here me asking about how-tos. How can I be sure my local environment is up and running without any problem and I can test locally? Have you ever encountered such problems before? is it just me. Thanks

sdirix:
Hi @onerzafer, you should be able to run the tests via `npm ci && npm run init && npm run build && npm run test`. On Linux you should be able to use Node > 12 while on Mac or Windows it needs to be Node 12 exactly. This is also how our CI is configured.

packages/core/src/util/runtime.ts
```diff
@@ -177,3 +180,51 @@ export const isEnabled = (
...
+    return false;
+  }
+  if (typeof schema?.readOnly === 'boolean' && schema.readOnly) {
+    return false;
+  }
+  if (typeof ownProps?.enabled === 'boolean') {
+    return ownProps.enabled;
+  }
+  return true;
+};
```
sdirix, line 230: This is still performing an `OR` operation ;) I wanted it to look like this:
```suggestion
  if (state?.jsonforms?.readonly) {
    return false;
  }
  if (
    uischema &&
    hasEnableRule(uischema)
  ) {
    return isEnabled(uischema, rootData, ownProps?.path, getAjv(state));
  }
  if (
    typeof uischema?.options?.readonly === 'boolean'
  ) {
    return !uischema.options.readonly;
  }
  if (
    typeof uischema?.options?.readOnly === 'boolean'
  ) {
    return !uischema.options.readOnly;
  }
  if (typeof config?.readonly === 'boolean') {
    return !config.readonly;
  }
  if (typeof config?.readOnly === 'boolean') {
    return !config.readOnly;
  }
  if (typeof schema?.readOnly === 'boolean') {
    return !schema.readOnly;
  }
  if (typeof ownProps?.enabled === 'boolean') {
    return ownProps.enabled;
  }
  return true;
};
```

sdirix (changes requested):

packages/core/src/util/runtime.ts
onerzafer: Hey, I've done this at first but I noticed it's killing the whole override mentality you built. But if you need that sure thing, it will my pleasure to do it. On it.

sdirix: The question for me is where to place the rule evaluation. Whether it makes more sense at the top, like it is right now, or whether local uischema.options can override them, i.e. they should be placed to before the `schema.readOnly` check.

onerzafer: That's the thing I don't know. It feels more natural to me to keep the original behaviour as is except if there is an explicit "readonly" on the schema of the field or respective uischema. What do you think?

onerzafer: By the way no luck on running the test (vue part is failing to run) due to limitations of M1 I have to use node 15 :(

onerzafer: I updated the algorithm per your suggestion

onerzafer:
Hello, @sdirix do you think it is worth merging? Do you have more feedback? I'd love to here :D

packages/examples/src/readonly.ts
```diff
@@ -0,0 +1,81 @@
...
+  THE SOFTWARE.
+*/
+import { registerExamples } from './register';
+
+export const schema = {
+  type: 'object',
+  properties: {
+    readonly: {
+      type: 'string',
+      readonly: true
```
sdirix, line 32:
```suggestion
      readOnly: true
```

packages/core/src/util/runtime.ts
```diff
@@ -177,3 +180,41 @@ export const isEnabled = (
...
+  }
+  if (typeof config?.readOnly === 'boolean') {
+    return !config.readOnly;
+  }
+  if (typeof schema?.readOnly === 'boolean') {
+    return !schema.readOnly;
+  }
+  if (typeof schema?.readonly === 'boolean') {
+    return !schema.readonly;
+  }
```
sdirix, line 215: The schema specification only defines `readOnly` so we should not react on `readonly`
```suggestion
```

packages/core/src/util/cell.ts
```diff
@@ -109,17 +110,22 @@ export const mapStateToCellProps = (
...
 
+  const config = getConfig(state);
+  const enabled: boolean = isInherentlyEnabled(
+    state,
+    ownProps,
+    uischema,
+    schema || rootSchema,
+    rootData,
+    config
+  );
```
sdirix, line 127: For cells we should keep the old behavior. The old algorithm preferred `ownProps` over `rules`. This makes sense as we don't distinguish between cells. Either all cells are enabled or all cells are disabled. The Table renderer by default just passes its own `enabled` prop to all cells.

Therefore the enabled calculation via rules for cells is just a fallback for non-default table renderers which for some reason don't pass on the `enabled` prop. We should prefer the passed on prop (i.e. `ownProp`) as otherwise the same ui schema rule might be evaluated again and again for each of the cells. This will get really expensive for large tables as validating is not cheap.
```suggestion
  const config = getConfig(state);
  const enabled: boolean = typeof ownProps.enabled === 'boolean' ? ownProps.enabled : isInherentlyEnabled(
    state,
    ownProps,
    uischema,
    schema || rootSchema,
    rootData,
    config
  );
```

The old cell tests actually tested for this and need to be restored.

packages/core/test/util/cell.test.ts
```diff
@@ -152,10 +152,10 @@ test('mapStateToCellProps - visible via state ', t => {
   t.true(props.visible);
 });
 
-test('mapStateToCellProps - enabled via ownProps ', t => {
+test('mapStateToCellProps - enabled via rule ', t => {
```
sdirix, line 155: CI is failing as the `disableRule` no longer occurs in any test

packages/core/test/util/runtime.test.ts
```diff
@@ -514,6 +531,96 @@ test('evalEnablement disable invalid case based on schema condition', t => {
...
+    expectedValue: 'bar'
+  };
+  const uischema: ControlElement = {
+    type: 'Control',
+    scope: '#/properties/value',
+    rule: {
+      effect: RuleEffect.DISABLE,
+      condition: leafCondition
+    }
+  };
```
sdirix, line 609: We no longer recommend this old type of rules for a long time. Would be preferable to use a [schema-based rule](https://jsonforms.io/docs/uischema/rules).

packages/core/test/util/runtime.test.ts
```diff
@@ -514,6 +531,96 @@ test('evalEnablement disable invalid case based on schema condition', t => {
...
+
+test('isInherentlyEnabled disabled by config', t => {
+  t.false(
+    isInherentlyEnabled(null, null, null, null, null, { readonly: true })
+  );
+});
+
+test('isInherentlyEnabled enabled', t => {
+  t.true(isInherentlyEnabled(null, null, null, null, null, null));
 });
```
sdirix, line 626: Would be great to add some tests which also test the combination of multiple readOnlys being present, especially when they overwrite each other. For example when an `ownProp.enabled` is `false` but a `uischema.option.readonly: false` is set.

Obviously testing all combinations is a bit much but a reasonable subset would be great.

sdirix (changes requested):
We're coming closer to merging but there are still some issues.

onerzafer:
Hello, @sdirix I hope I'm not bothering you. I feel like this is ready but I'm not totally sure since I have no deeper understanding of the codebase yet. This is the core package and obviously needs to be more critical while merging. May I ask if you think it's done or needs more work? And from an optimistic point of view in case we managed to merge it, is it realistic to expect a release afterwards anytime soon? Thanks again for bearing with me. Cheers

sdirix:
Hi @onerzafer! I'll take a detailed look in the near future, trying to break the new functionality. It's probably merged within the next days in case we didn't overlook something major (don't think so). We'll also release a 3.0 alpha version in the same time frame which will then probably contain this change.

sdirix (approved):
LGTM ;)

sdirix:
Hi @onerzafer, JSON Forms `3.0.0-alpha.0` is now released containing the readonly changes. You can consume it by referring to the concrete version or simply use the `next` tag, for example `npm install @jsonforms/core@next @jsonforms/react@next @jsonforms/material-renderers@next`.

Thank you again for the contribution :)

onerzafer:
Hi @sdirix this is wunderbar! I appreciate your guidance and effort. Have a good one.

PR #1766, opened 2021-06-23
Move logic out of index files

TheZoker:
- Move logic out of index files
- Index only exports from other files
- Direct imports within the same directory
- Index imports from other directories


sdirix (changes requested):
I briefly looked over the changes. Already looks pretty good! Can you please rebase on the latest master?

sdirix (approved):
LGTM

PR #1775, opened 2021-07-06
Limit github actions run to 30 min to avoid 6h runs

TheZoker:
Sometimes Github actions don't finish and then they run for 6h:
https://github.com/eclipsesource/jsonforms/actions/runs/970989014

To avoid this, this PR sets a timeout of 30m, which should work just fine (last successful builds took ~ 15m - 19m).


sdirix (approved):
LGTM

Issue #1748, opened 2021-05-10
Material-UI: the ExpansionPanel component was renamed to Accordion


boerngen-schmidt:
Using the bundled material-ui renderers for rendering an Array gives the following error in the browsers console.

```
Material-UI: the ExpansionPanel component was renamed to Accordion to use a more common naming convention.

You should use `import { Accordion } from '@material-ui/core'`
or `import Accordion from '@material-ui/core/Accordion'`
```

The rendering still works fine! But this might change.

sdirix:
Thanks for the notice!

stevenmckinnon:
@boerngen-schmidt what version of material ui are you using? I'm testing the 5.0.0-alpha and it won't compile because of this.

boerngen-schmidt:
@stevenmckinnon I do use 

    "@material-ui/core": "^4.11.3",
    "@material-ui/icons": "^4.11.2",
    "@material-ui/styles": "^4.11.3",

stevenmckinnon:
@boerngen-schmidt Thanks, looks like it's going to be a big shift to 5.0.0!

TomerAmitzur:
whats going on with that issue? when it supposed to be solved?

sdirix:
@TomerAmitzur We currently plan it for the next stable release of JSON Forms which will be `3.0`. Currently @max-elia is assigned to work on this issue but if you'd like to contribute it yourself then this is also fine. Just let me know in advance so that this work is not done twice.

Once this fix is merged to master it'll of course also be available in the next prerelease of JSON Forms too.

sdirix:
Version `3.0.0-alpha.1` containing the fix for this issue is now released.

PR #1776, opened 2021-07-06
Rename ExpansionPanel to Accordion as suggested by Material UI

TheZoker:
The `ExpansionPanel` component was renamed to `Accordion` within this PR:
https://github.com/mui-org/material-ui/pull/21494

With the upcoming version v5 of material UI, the expansion panel will be completely removed.

The docs suggest, just to switch to Accordion, no further migration needs to be done:
https://material-ui.com/api/expansion-panel/

Fixes: #1748


sdirix (approved):
LGTM

PR #1786, opened 2021-07-13
Improve label handling in MasterListComponent

### LINKED ISSUE #1779: Handle primitive labeling in angular-material MasterListComponent

When using a `ListWithDetail` on an array of strings, the label stays undefined even if the `options.labelRef` is set to `#`.

```json
{
  "type": "ListWithDetail",
  "scope": "#/properties/an-array-of-strings",
  "options": {
    "labelRef": "#"
  }
}
```

I'd like the label to be the value of the item. This could be achieved by modifiying the following in [master.ts](https://github.com/eclipsesource/jsonforms/blob/f40e95b5d6815b94f30460fa68e03b8e815a36be/packages/angular-material/src/other/master-detail/master.ts#L186)

```typescript
const dataIsPrimitive = d && typeof d !== 'object';
const masterItem = {
  label: get(d, labelRefInstancePath, dataIsPrimitive ? d : ''),
  data: d,
  path: `${path}.${index}`,
  schema,
  uischema: detailUISchema
};
```

Using "my own" version of `MasterListComponent`.

### Framework

Angular

### RendererSet

Material

### Additional context

_No response_

eneufeld:
The MasterListComponent did only show labels for object lists.
Furthermore the options property in the uischema was mandatory.
Also the property to use as label for objects had to be provided.

The fix now handles primitives, a missing options property and
in the case of a missing `labelRef` the first primitive property
is used. If no primitive property is available the first property in
the schema is used.

Fix #1779

### DIFF
--- a/packages/angular-material/src/other/master-detail/master.ts
+++ b/packages/angular-material/src/other/master-detail/master.ts
@@ -34,6 +34,7 @@ import {
   ControlElement,
   createDefaultValue,
   findUISchema,
+  getFirstPrimitiveProp,
   JsonFormsState,
   mapDispatchToArrayControlProps,
   mapStateToArrayControlProps,
@@ -169,22 +170,22 @@ export class MasterListComponent extends JsonFormsArrayControl {
     const { data, path, schema, uischema } = props;
     const controlElement = uischema as ControlElement;
     this.propsPath = props.path;
-    const detailUISchema =
-      controlElement.options.detail ||
-      findUISchema(
+    const detailUISchema = findUISchema(
         props.uischemas,
         schema,
         `${controlElement.scope}/items`,
         props.path,
-        'VerticalLayout'
+        'VerticalLayout',
+        controlElement
       );
 
     const masterItems = (data || []).map((d: any, index: number) => {
-      const labelRefInstancePath = removeSchemaKeywords(
+      const labelRefInstancePath = controlElement.options?.labelRef && removeSchemaKeywords(
         controlElement.options.labelRef
       );
+      const isPrimitive = d && typeof d !== 'object';
       const masterItem = {
-        label: get(d, labelRefInstancePath),
+        label: isPrimitive ? d.toString() : get(d, labelRefInstancePath ?? getFirstPrimitiveProp(schema)),
         data: d,
         path: `${path}.${index}`,
         schema,
--- a/packages/angular-material/src/other/master-detail/master.ts
+++ b/packages/angular-material/src/other/master-detail/master.ts
@@ -183,7 +183,7 @@ export class MasterListComponent extends JsonFormsArrayControl {
       const labelRefInstancePath = controlElement.options?.labelRef && removeSchemaKeywords(
         controlElement.options.labelRef
       );
-      const isPrimitive = d && typeof d !== 'object';
+      const isPrimitive = d !== undefined && typeof d !== 'object';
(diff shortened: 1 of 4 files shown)


packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -247,6 +249,15 @@ export class MasterListComponent extends JsonFormsArrayControl {
...
 }
 
+const getFirstUsableProperty = (schema: JsonSchema): string => {
+  const prop = Object.entries(schema?.properties).find(e => {
+    if(e[1].type === 'object' || e[1].properties) return false;
+    if(e[1].type === 'array' || e[1].items) return false;
+    return true;
+  });
+  return prop ? prop[0] : Object.keys(schema?.properties)[0];
+}
```
sdirix, line 259: There is a utility for this in `@jsonforms/core` which you can import: `getFirstPrimitiveProp`.

packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -170,7 +171,7 @@ export class MasterListComponent extends JsonFormsArrayControl {
     const controlElement = uischema as ControlElement;
     this.propsPath = props.path;
     const detailUISchema =
-      controlElement.options.detail ||
+      controlElement.options?.detail ||
       findUISchema(
```
sdirix, line 175: Not directly related to this PR but I just noticed this issue here: `findUISchema` already checks the `options.detail`, so we don't need to do this manually. Especially as our API also allows the strings `DEFAULT`, `GENERATED` and `REGISTERED` as valid `detail` values ;)

We can just hand over the `controlElement` as the next parameter after `'VerticalLayout'`, i.e.

```ts
const detailUISchema = findUISchema(
  props.uischemas,
  schema,
  `${controlElement.scope}/items`,
  props.path,
  'VerticalLayout',
  controlElement,
);
```

sdirix (changes requested):
Thanks for the fix and the additional improvements! Noticed some minor issues.

packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -169,22 +170,22 @@ export class MasterListComponent extends JsonFormsArrayControl {
...
+        'VerticalLayout',
+        controlElement
       );
 
     const masterItems = (data || []).map((d: any, index: number) => {
-      const labelRefInstancePath = removeSchemaKeywords(
+      const labelRefInstancePath = controlElement.options?.labelRef && removeSchemaKeywords(
         controlElement.options.labelRef
       );
+      const isPrimitive = d && typeof d !== 'object';
```
sdirix, line 186: this is false for `0` and empty strings `''`

eneufeld: Will fix, JS and auto conversion of types ...

packages/examples/src/1779.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+const uischema_number = {
+    "type": "ListWithDetail",
+    "scope": "#/properties/an-array-of-numbers"
+};
+
+registerExamples([
+    {
+        name: '1779',
+        label: 'List With Detail primitive (string)',
```
sdirix, line 78:
```suggestion
        label: 'List With Detail primitive (number)',
```

sdirix (approved):

PR #1789, opened 2021-07-26
Use filtered categories in childProps

### LINKED ISSUE #1777: Categorization Stepper of React Material broken for conditional categories
### Describe the bug

When an "inbetween" category is removed via a hide rule the `Next` button will still navigate to it. 

### Expected behavior

`Next` should navigate to the correct next categorization. The stepper indicators are already correctly indicated so it should be rather straightforward to fix.

Can be reproduced in the last categorization example https://jsonforms.io/examples/categorization/

1. Uncheck `Provide Address`
2. Check `Vegetarian`
3. Click `Next`

### Screenshots

image

`Address` category is shown although it should have transitioned to `Additional`

Chrome (latest)

### Framework

React

### RendererSet

Material

### Additional context

_No response_

max-elia:
While categories have been filtered in Stepper, they are now also being used in childProps.

Fixes #1777


max-elia:
@sdirix

PR #1792, opened 2021-07-30
Improve Material Time Pickers

sdirix:
Reworks date, time and date-time support in the React Material
renderer set. The controls now support many customization options
and can be invoked separately via the ui schema.

We now use the much smaller 'dayjs' library (compared to 'moment')
for Javascript date support. It can be configured globally and is
therefore easily customizable on client side for their locales.

Also streamlines the label prop which was only supporting string for
a long time now.


eneufeld (commented):
This looks good to me, thank you!

eneufeld (approved):

PR #1793, opened 2021-07-31
Fix small bug in README example code

larsw:
const data -> const initialData


sdirix (approved):
LGTM! Thanks for the contribution!

Issue #1352, opened 2019-04-27
Decimals lower than 0.1 are unsupported in 'number' fields


johnmorrell:
**Describe the bug**

Entering a decimal lower than 0.1 (eg: 0.09) in to a number field, clears the file on the entry of the zero after the decimal place.

**To Reproduce**

Enter 0.09 in to a number field

**Expected behavior**

Decimals of lower than 0.1 to be retained in the UI and data

**Browser**
 - Browser: Chrome

**Used Setup**
 - Framework: react
 - RendererSet: material

**Additional context**

eneufeld:
fixed via #1353

Issue #1351, opened 2019-04-27
Zero value not displaying in 'number' fields


johnmorrell:
**Describe the bug**

When entering zero in to a number field (where 0 is the intended value) the number does not show in the input field but is present in the data. This is also the case for decimals such as "0.1", where ".1" is shown.

**To Reproduce**

Enter "0" and "0.1" in to a number field

**Expected behavior**

"0" and "0.1" to show in the number field

**Browser**
 - Chrome

**Used Setup**
 - Framework: react
 - RendererSet: material

PR #1796, opened 2021-08-12
Explicitly check for undefined and null

awolokita:
This addresses the same issue as seen in  #1351 #1352 and fixed for the number controller in #1353.

### DIFF
--- a/packages/material/src/mui-controls/MuiSelect.tsx
+++ b/packages/material/src/mui-controls/MuiSelect.tsx
@@ -50,7 +50,7 @@ export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
       id={id}
       disabled={!enabled}
       autoFocus={appliedUiSchemaOptions.focus}
-      value={data || ''}
+      value={data !== undefined && data !== null ? data : ''}
       onChange={ev => handleChange(path, ev.target.value)}
       fullWidth={true}
     >
--- a/packages/material/src/mui-controls/MuiSelect.tsx
+++ b/packages/material/src/mui-controls/MuiSelect.tsx
@@ -50,7 +50,7 @@ export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
       id={id}
       disabled={!enabled}
       autoFocus={appliedUiSchemaOptions.focus}
-      value={data !== undefined && data !== null ? data : ''}
+      value={data !== undefined ? data : ''}
       onChange={ev => handleChange(path, ev.target.value)}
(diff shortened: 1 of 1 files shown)


packages/material/src/mui-controls/MuiSelect.tsx
```diff
@@ -50,7 +50,7 @@ export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
       id={id}
       disabled={!enabled}
       autoFocus={appliedUiSchemaOptions.focus}
-      value={data || ''}
+      value={data !== undefined && data !== null ? data : ''}
```
CarlosGomez-dev, line 53:
```suggestion
      value={data ?? ''}
```

awolokita: I took the same solution as implemented in **MuiInputInteger.tsx**. I'm assuming then that we're targetting TS >= 3.7 with nullish coalescing?

sdirix:
Thank you very much for the contribution :+1:

packages/material/src/mui-controls/MuiSelect.tsx
sdirix: Yes we can use the `??` operator here. The other fix is over 2 years old, so that's why it wasn't used there.

However conceptually `null` is a valid value for enums, so what we really should do is just check for `undefined`, i.e.
```suggestion
      value={data !== undefined ? data : ''}
```

sdirix (changes requested):
Thanks for the contribution. As indicated in https://github.com/eclipsesource/jsonforms/pull/1796/files#r688268551 we should still allow for `null` values.

packages/material/src/mui-controls/MuiSelect.tsx
awolokita: Thanks. I'll make the change. Out of curiosity, how is `null` rendered?

sdirix: If no label is given via the [`oneOf` `const` construct](https://github.com/eclipsesource/jsonforms/blob/ba6bdc5a93044a872285cd71ed40d9a54f437561/packages/examples/src/enum.ts#L38-L45) we just [stringify the value](https://github.com/eclipsesource/jsonforms/blob/ba6bdc5a93044a872285cd71ed40d9a54f437561/packages/core/src/util/renderer.ts#L162-L165), so it will be shown as `"null"`.

sdirix (approved):

PR #1798, opened 2021-08-16
Add required prop to Inputs

### LINKED ISSUE #1695: Use the Material UI "required" prop
**Is your feature request related to a problem? Please describe.**
When marking a field as required, it only adds the asterisk to the label text rather than using the required prop in InputLabel, which doesn't add the separate `<span />` Material UI renders. Therefore the styling can't be overridden (i.e. style it as red by default). 

**Describe the solution you'd like**
Use InputLabels's required prop

**Describe for which setup you like to have the improvement**
Framework: React
RendererSet: material

**Additional context**
[https://material-ui.com/api/input-label/#props:~:text=FormControl.-,required](https://material-ui.com/api/input-label/#props:~:text=FormControl.-,required)

LukasBoll:
Material-Ui components are now using the required prop where possible.

closes #1695
Signed-off-by: Lukas Boll <lukas-bool@web.de>


TheZoker (changes requested):
LGTM, but I think this can also be applied to `MaterialNativeControl`
https://github.com/eclipsesource/jsonforms/blob/master/packages/material/src/controls/MaterialNativeControl.tsx#L77-L81

LukasBoll:
Thank You! I updated the pull request and also included the MaterialNativeControl.

TheZoker (approved):
LGTM, thanks 🎉

packages/core/src/util/renderer.ts
```diff
@@ -98,6 +98,19 @@ export const computeLabel = (
   return required && !hideRequiredAsterisk ? label + '*' : label;
 };
 
+/**
+ * Decides whether to mark a field as required.
+ *
+ * @param {boolean} required whether the label belongs to a control which is required
+ * @returns {boolean} should the field be marked as required
+ */
```
sdirix, line 106: The doc is missing the second parameter.

packages/core/src/util/renderer.ts
```diff
@@ -98,6 +98,19 @@ export const computeLabel = (
   return required && !hideRequiredAsterisk ? label + '*' : label;
 };
 
+/**
+ * Decides whether to mark a field as required.
```
sdirix, line 102:
```suggestion
 * Indicates whether to mark a field as required.
```

packages/core/src/util/renderer.ts
```diff
@@ -98,6 +98,19 @@ export const computeLabel = (
...
+ *
+ * @param {boolean} required whether the label belongs to a control which is required
+ * @returns {boolean} should the field be marked as required
+ */
+ export const displayRequired = (
```
sdirix, line 107: I would prefer renaming this a bit for more natural reading, i.e. `displayAsRequired`. Maybe also use `show` instead of `display`?
```suggestion
 export const showAsRequired = (
```

packages/material/src/controls/MaterialSliderControl.tsx
```diff
@@ -94,14 +95,20 @@ export class MaterialSliderControl extends Control<ControlProps, ControlState> {
...
-              required,
-              appliedUiSchemaOptions.hideRequiredAsterisk
-            )}
-          </Typography>
+          <FormLabel
+            htmlFor={id}
+            error={!isValid}
+            component={'legend' as 'label'}
+            required={displayRequired(required,
+              appliedUiSchemaOptions.hideRequiredAsterisk)}
```
sdirix, line 106: Good catch!

sdirix (changes requested):
Works for me! Good work!

I only have some minor comments.

LukasBoll:
I applied your suggested changes

packages/core/src/util/renderer.ts
```diff
@@ -88,6 +88,7 @@ const isRequired = (
  *
  * @param {string} label the label string
  * @param {boolean} required whether the label belongs to a control which is required
+ * @param {boolean} applied hideRequiredAsterisk option
```
sdirix, line 91: Nice that you also changed it here!

packages/core/src/util/renderer.ts
```diff
@@ -98,6 +99,20 @@ export const computeLabel = (
   return required && !hideRequiredAsterisk ? label + '*' : label;
 };
 
+/**
+ * Indicates whether to mark a field as required.
+ *
+ * @param {boolean} required whether the label belongs to a control which is required
+ * @param {boolean} applied hideRequiredAsterisk option
```
sdirix, line 106: The first word should be the parameter name as it's specially handled by the JSDoc
```suggestion
 * @param {boolean} hideRequiredAsterisk applied UI Schema option
```

LukasBoll: Thank you for the hint, i updated the doc

sdirix (approved):
LGTM

PR #1809, opened 2021-09-28
Remote container setup for VS-Code

tortmayr:
Setup a remote container dev setup using the "ms-vscode-remote.remote-containers" extension

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.191.1/containers/javascript-node/.devcontainer/base.Dockerfile
+
+# [Choice] Node.js version: 16, 14, 12
+ARG VARIANT="16-buster"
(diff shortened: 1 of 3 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.191.1/containers/javascript-node/.devcontainer/base.Dockerfile
+
+# [Choice] Node.js version: 16, 14, 12
+ARG VARIANT="16-buster"
```
sdirix, line 4: Here it's `16` but in `decontainer.json` it's `12`. Is that intended?

sdirix (changes requested):
Please remove the `package-lock` changes.

.devcontainer/Dockerfile
tortmayr: That's intended. "16-buster" is the default value to ensure that the docker file is also working standalone, and we overwrite the variant arg to "12" in the `devcontainer.json`.

sdirix (approved):

PR #1812, opened 2021-10-04
Update MUI to version 5

### LINKED ISSUE #1805: Update React Material renderer set to Material UI v5

Material UI v5 has some breaking changes, most notably the import paths. 

Update the React Material renderer set to Material UI v5

Adjust React Material renderer set to be compatible with Material UI v4 and v5. However this requires additional effort and may have consequences on the renderer set architecture.

### Framework

React

### RendererSet

Material

### Additional context

https://next.material-ui.com/guides/migration-v4/

TheZoker:
Closes #1805

- Update material ui to version 5.1.0
- Replace `@material-ui/pickers` with `@mui/lab` pickers
- Update enzym adapter to version 17


LukasBoll (commented):
Code is looking good to me!
I checked all the examples, there is just one issue with the Categorization.
The label of the selected tap/category is not visible in the header.

TheZoker:
Should be fixed:

image

sdirix (changes requested):
Works great for me! Thanks for the work.

I have just minor feedback, see my comments.

packages/material/package.json
```diff
@@ -71,32 +71,29 @@
...
     "@jsonforms/react": "^3.0.0-alpha.0",
-    "@material-ui/core": "^4.7.0",
-    "@material-ui/icons": "^4.5.1",
-    "@material-ui/pickers": "^3.3.10"
-  },
-  "optionalPeerDependencies": {
-    "@material-ui/lab": "^4.0.0-alpha.56"
+    "@mui/icons-material": "^5.2.0",
+    "@mui/lab": "^5.0.0-alpha.58",
+    "@mui/material": "^5.2.2"
```
sdirix, line 76: Do we really depend on the latest versions of all these frameworks? If 5.0.0 is good enough then we should indicate as such to not force users to upgrade from 5.1 to 5.2 for example when we don't actually require them to.

sdirix (approved):
LGTM. Thanks for the great work!

PR #1813, opened 2021-10-06
Add debounce for input material controls

### LINKED ISSUE #1645: Debounce changes
Increase performance by debouncing changes. See also https://github.com/eclipsesource/jsonforms/issues/1641 and [lag happening on typing](https://spectrum.chat/jsonforms/general/performance-issue-lag-happening-on-typing~c2b961a2-988f-4ac3-ab89-02c70cb53f81).

eneufeld:
- add a custom hook which handles debounce
- add example to test form performance

Fix #1645

### DIFF
--- a/packages/material/src/mui-controls/MuiInputInteger.tsx
+++ b/packages/material/src/mui-controls/MuiInputInteger.tsx
@@ -27,6 +27,11 @@ import { CellProps, WithClassname } from '@jsonforms/core';
 import Input from '@material-ui/core/Input';
 import { areEqual } from '@jsonforms/react';
 import merge from 'lodash/merge';
+import { useDebouncedChange } from '../util';
+
+const toNumber = (value: string) =>
+      value === '' ? undefined : parseInt(value, 10);
+const eventToValue = (ev:any) => toNumber(ev.target.value);
 
 export const MuiInputInteger = React.memo(
   (props: CellProps & WithClassname) => {
@@ -41,15 +46,16 @@ export const MuiInputInteger = React.memo(
       config
     } = props;
     const inputProps = { step: '1' };
-    const toNumber = (value: string) =>
-      value === '' ? undefined : parseInt(value, 10);
+    
     const appliedUiSchemaOptions = merge({}, config, uischema.options);
+
+    const [inputValue, onChange] = useDebouncedChange(handleChange, '', data, path, eventToValue);
 
     return (
       <Input
         type='number'
-        value={data !== undefined && data !== null ? data : ''}
-        onChange={ev => handleChange(path, toNumber(ev.target.value))}
+        value={inputValue}
+        onChange={onChange}
         className={className}
         id={id}
         disabled={!enabled}
--- a/packages/material/src/mui-controls/MuiInputNumber.tsx
+++ b/packages/material/src/mui-controls/MuiInputNumber.tsx
@@ -27,7 +27,11 @@ import { CellProps, WithClassname } from '@jsonforms/core';
 import Input from '@material-ui/core/Input';
 import { areEqual } from '@jsonforms/react';
 import merge from 'lodash/merge';
+import {useDebouncedChange} from '../util';
 
+const toNumber = (value: string) =>
+    value === '' ? undefined : parseFloat(value);
+const eventToValue = (ev:any) => toNumber(ev.target.value);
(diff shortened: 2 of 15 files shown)


packages/material/src/mui-controls/MuiInputInteger.tsx
```diff
@@ -40,16 +45,17 @@ export const MuiInputInteger = React.memo(
       handleChange,
       config
     } = props;
-    const inputProps = { step: '1' };
-    const toNumber = (value: string) =>
-      value === '' ? undefined : parseInt(value, 10);
+    const inputProps = useRef({ step: '1' });
```
sdirix, line 48: Let's remove the `useRef` here. I'm wondering why this doesn't break anything as this is still given unchanged to `Input`.

packages/material/src/mui-controls/MuiInputNumber.tsx
```diff
@@ -39,16 +43,16 @@ export const MuiInputNumber = React.memo((props: CellProps & WithClassname) => {
     handleChange,
     config
   } = props;
-  const inputProps = { step: '0.1' };
-  const toNumber = (value: string) =>
-    value === '' ? undefined : parseFloat(value);
+  const inputProps = useRef({ step: '0.1' });
```
sdirix, line 46: I don't think we need the `useRef` here.

packages/material/src/util/debounce.ts
```diff
@@ -0,0 +1,16 @@
+import { debounce } from 'lodash';
+import { useState, useCallback } from 'react'
+
+
+const eventToValue = (ev: any) => ev.target.value;
+export const useDebouncedChange = (handleChange: (path: string, value: any) => void, defaultValue: any, data: any, path: string, eventToValueFunction: (ev: any) => any = eventToValue, timeout = 300): [any, React.ChangeEventHandler, () => void] => {
+  const [input, setInput] = useState(data || defaultValue);
+  const debouncedUpdate = useCallback(debounce((newValue: string) => handleChange(path, newValue), timeout), [handleChange]);
```
sdirix, line 8: This should be a `useMemo` as we are calculating a function and not creating the function ourselves in place. A proper ESLint setup would complain here ;) 
```suggestion
  const debouncedUpdate = useMemo(() => debounce((newValue: string) => handleChange(path, newValue), timeout), [handleChange]);
```

sdirix (changes requested):
I did not test the changes yet but conceptionally this looks as expected.

However we don't need any of the `useRef`, `useMemo` and `useCallbacks`. They complicate the code and don't prevent any rerenderings.

The renderers are already memoized, so if nothing changed they should not even render. If something changed they need to rerender anyway, so memoizing some of the props for their children will not prevent their rerender as something else changes too.

The only exception is the debounced_setter. This needs to be placed in a `useMemo`, or even better in a `useState` to guarantee it's always the same and debouncing works.

packages/material/src/util/debounce.ts
eneufeld: we are actually creating a function, a function that is used to call handleChange

sdirix: Yes that's true, and technically that's completely fine. 

I just noticed the pattern because ESLint usually complaints about it ("exhaustive-deps") as it can't properly analyze the function and therefore doesn't give you the usual guarantees. As a workaround we can use `useMemo`. We just don't have ESLint yet configured. I'm fine leaving it like this, so we can also change this later when we finally use ESLint.

packages/material/src/util/debounce.ts
```diff
@@ -0,0 +1,16 @@
...
+  const onChange = useCallback((ev: any) => {
+    const newValue = eventToValueFunction(ev);
+    setInput(newValue);
+    debouncedUpdate(newValue);
+  }, [debouncedUpdate]);
+  const onClear = useCallback(() => { setInput(defaultValue); handleChange(path, undefined) }, [handleChange]);
```
sdirix, line 14: Now that we're talking about "exhaustive-deps" in the other comment chain, I just realized that `defaultValue` and `path` is missing here in the dependencies.

packages/material/src/util/debounce.ts
sdirix: Now that we're talking about it: `timeout` is missing in the dependencies as well as `path`.

PR #1825, opened 2021-10-26
Rework i18n support in JSON Forms core

sdirix:
Labels, descriptions and error messages can now be translated via
special translation functions handed over to JSON Forms.

The translations are automatically handed within the default
mapping functions and are therefore available in all renderer sets.
This includes a key-determination algorithm, allowing to either rely
on using labels as keys or specifying 'i18n' keys in UI Schema options
or directly within the JSON Schema. Errors are handled separately to
allow for maximum flexibility.

Includes test cases for the most common mapping functions.

Also AJV is set to non-strict by default to not throw errors when
handing over JSON Schemas containing 'i18n' keys.

### DIFF
--- a/packages/angular-material/example/app/app.component.ts
+++ b/packages/angular-material/example/app/app.component.ts
@@ -75,7 +75,7 @@ const itemTester: UISchemaTester = (_schema, schemaPath, _path) => {
       [schema]="selectedExample.schema"
       [uischema]="selectedExample.uischema"
       [renderers]="renderers"
-      [locale]="currentLocale"
+      [i18n]="i18n"
       [uischemas]="uischemas"
       [readonly]="readonly"
       [config]="config"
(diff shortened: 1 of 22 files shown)


andresgutgon:
@sdirix could you put an example here with i18n keys in the schema or uischema? I don't find the right way of setting this in the case of `enum`s

yaffol:
There is an example in the Vue2-Vuetify renderer set for the 'basic' example https://jsonforms-vuetify-renderers.netlify.app/#/example/basic

That has i18n keys and also an enum, maybe that will be helpful.

sdirix:
The enums actually have a bug right now in the React Material renderers (Angular and Vue are fine). They will render correctly with the first locale selection but will not update their labels when the translator is changed without an additional schema changed. This will be fixed here: https://github.com/eclipsesource/jsonforms/pull/1879

Issue #1491, opened 2019-08-28
Unable to use recursive schema


jondoe1337:
**Describe the bug**
I have a json schema which uses a `$ref` as a items reference: https://pastebin.com/EJBLXZ3b (line 57)
After including it in the example and running it: I see in the dev console of my chrome:

```
index.js:42 Uncaught TypeError: Converting circular structure to JSON
    at stringify (index.js:42)
    at stringify (index.js:50)
    at stringify (index.js:50)
    at stringify (index.js:50)
    at stringify (index.js:50)
    at stringify (index.js:50)
    at stringify (index.js:50)
    at stringify (index.js:50)
    at stringify (index.js:50)
````

When I remove line 57 in the schema, it works. 

**To Reproduce**
1. Use React without redux example and use the schema shown above
2. Start example
3. Open dev console

**Expected behavior**
A schema with recursion should work, or at least not fail to be used.

**Browser:**
 - Chrome
 - 76.0.3809.100 (OSX 64bit)

**Used Setup:**
 - Framework: react + express
 - RendererSet: material

Lily418:
Hi @jondoe1337 

This is a related issue
https://github.com/eclipsesource/jsonforms/issues/1476

and some discussion about cyclic refs.
https://github.com/eclipsesource/jsonforms/pull/1474#issuecomment-522529303

Also which version of JSON Forms are you using? I think this error is caused by a change introduced in #1474 but that's not included in latest stable release v2.3.0

jondoe1337:
Hi @Lily418 sorry for overseeing those issues.

I use v2.3.1

Lily418:
I'd try v2.3.0 to avoid that error, but be aware of Stefan's comment https://github.com/eclipsesource/jsonforms/pull/1474#issuecomment-522529303 this may be an issue depending on your use case.

sdirix:
The non-redux version of the react-seed partially works for me with your schema when, ironically, the ref parser is configured in the redux store with ```{
    refParserOptions: {
      dereference: {
        circular: "ignore"
      }
    }
  }``` as the last argument of the `init` action.

There seems to be some unintended interaction between the examples in the seed. Also the standalone version seems to hiccup on some of the nested objects while the redux version has no problems there.

We'll need to investigate the behavior of our example seed and how JSON Forms behaves with your schema in a pure non-redux app. Note that you can also pass the refParser options to the `JsonForms` component when needed, although it seems to worsen the behavior in the seed app.

[...]

PR #1829, opened 2021-11-04
Fix ref resolving for oneOf schemas

AlexandraBuzila:
- resolve items subschema refs for renderers and testers
- make rootSchema a required parameter in resolveSchema, as it is needed
while resolving
- add rootSchema as an optional parameter to Testers, so we can
correclty resolve subschemas in testers
- use resolved schemas in MaterialOneOfRenderer when computing default
value
- enable and update tests
- remove 'Resolve Remote Schema' example as it's unsupported


packages/core/src/util/resolvers.ts
```diff
@@ -157,10 +157,17 @@ export const resolveSchema = (
     try {
       return retrieveResolvableSchema(schema, resultSchema.$ref);
     } catch (e) {
-      return retrieveResolvableSchema(rootSchema, resultSchema.$ref);
+      if (rootSchema){
+        return retrieveResolvableSchema(rootSchema, resultSchema.$ref);
```
sdirix, line 161: Can we get rid off the `retrieveResolvableSchema` function completely in this refactoring? It modifies the schema which is actually a huge no-no.

packages/material/src/complex/MaterialOneOfRenderer.tsx
```diff
@@ -79,7 +79,7 @@ export const MaterialOneOfRenderer =
     const openNewTab = (newIndex: number) => {
       handleChange(
         path,
-        createDefaultValue(schema.oneOf[newIndex])
+        createDefaultValue(_schema.oneOf[newIndex])
```
sdirix, line 82: Can we rename `_schema` in the renderer? This is not obvious at all xD

AlexandraBuzila:
Open points that should be integrated in the PR:
- we are currently resolving the refs for the `items` property directly in the `resolveSchema` method (and thus modifying the schema). This should be removed from the method and directly happen in the testers and renderers. 
- the `rootSchema` property in the `resolveSchema` method should be mandatory
- check that all examples are working

packages/core/src/util/resolvers.ts
```diff
@@ -155,35 +155,25 @@ export const resolveSchema = (
...
-      return retrieveResolvableSchema(schema, resultSchema.$ref);
+      const resolvableSchema = resolveSchema(schema, resultSchema.$ref);
+      if (resolvableSchema !== undefined){
+        return resolvableSchema;
+      } 
+      if (rootSchema){
+        return resolveSchema(rootSchema, resultSchema.$ref);
+      }
+      return schema;
     } catch (e) {
```
AlexandraBuzila, line 166: Try/catch should be removed as it's no longer necessary. Lines 157-171 can be replaced with `return resolveSchema(rootSchema, resultSchema.$ref)`

packages/core/src/util/resolvers.ts
```diff
@@ -59,6 +59,17 @@ export const resolveData = (instance: any, dataPath: string): any => {
     }, instance);
 };
 
+export const resolveArraySchema = (schema: JsonSchema,
```
sdirix, line 62: We have to be careful with this method. It not only resolves something, but it actually creates a new object instance. So when it's used in some `mapState...` or in a renderer it will break memoization and lead to a full rerender every cycle.

At the moment it's used in `resolveSubSchemas` which already returns a new instance and therefore has the same problem. I guess we missed that originally as the combinator renderers (where it is used) are not that common. However this `resolveArraySchema` is now also called in `mapStateToControlProps` and will therefore break memoization for all schemas which have arrays whose items need to be resolved.

It's such a special case of resolving that I would not offer a special method for it. So we should remove `resolveArraySchema` but just handle it where it's needed:

- The testers for array items need to resolve themselves (which they can do when we give them the rootSchema)
- The `resolveSubSchemas` is actually not needed at all and can be removed completely. Instead the resolving can be done in `createCombinatorRendererInfos` and in `mapStateToCombinatorRendererProps` at the point of time it's needed.
- The resolving in `mapStateToControlProps` needs to be removed again and placed where it's needed, e.g. in `mapStateToArrayControlProps` where we can perform the item resolving.

sdirix (changes requested):
Already works much better than before. Please see my detailed comments for further improvements. Please also test against the schema and uischema from:

- https://github.com/eclipsesource/jsonforms/issues/1491
- https://github.com/jamesbmorris92/jsonforms-react-seed

AlexandraBuzila:
> Please also test against the schema and uischema from:
> 
> * #1491
> * https://github.com/jamesbmorris92/jsonforms-react-seed

Both examples work. #1491 is covered by the `oneOf-recursive` example.

packages/core/src/util/renderer.ts
```diff
@@ -946,7 +957,6 @@ export const mapStateToCombinatorRendererProps = (
 
   const ajv = state.jsonforms.core.ajv;
   const schema = resolvedSchema || rootSchema;
```
sdirix, line 959: This `schema` is the one which we hand over to the renderer. In case of resolving, this will always be a new instance and therefore break memorization. We should hand over an unchanged schema to the renderer and then deal with the consequences.

sdirix (commented):
Looks good! Added some comments and questions.

packages/core/src/testers/testers.ts
```diff
@@ -105,7 +110,10 @@ export const schemaSubPathMatches = (
   const schemaPath = uischema.scope;
   let currentDataSchema: JsonSchema = schema;
   if (hasType(schema, 'object')) {
-    currentDataSchema = resolveSchema(schema, schemaPath);
+    currentDataSchema = resolveSchema(schema, schemaPath, schema);
+  }
+  if (hasType(currentDataSchema, 'array')) {
+    currentDataSchema = { ...currentDataSchema, items: (resolveSchema(currentDataSchema, 'items', schema)) } as JsonSchema
   }
```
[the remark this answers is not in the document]
AlexandraBuzila: No, I was able to resolve the schema here as well in the predicate, where necessary.

packages/core/src/util/renderer.ts
AlexandraBuzila: I reverted the change.

packages/core/src/util/resolvers.ts
```diff
@@ -155,36 +156,10 @@ export const resolveSchema = (
   // with absolute paths here, so that we don't need to keep two different
   // schemas around
   if (resultSchema !== undefined && resultSchema.$ref !== undefined) {
-    try {
-      return retrieveResolvableSchema(schema, resultSchema.$ref);
-    } catch (e) {
-      return retrieveResolvableSchema(rootSchema, resultSchema.$ref);
-    }
+    return resolveSchema(schema, resultSchema.$ref, rootSchema)
```
sdirix, line 159: Is there an use case of trying to resolve the `$ref` against the `schema` first instead of directly going to the `rootSchema`?

sdirix (commented):
Thanks for the update! I have some more comments and questions

packages/core/src/testers/testers.ts
```diff
@@ -47,15 +47,17 @@ import { deriveTypes, hasType, resolveSchema } from '../util';
 export const NOT_APPLICABLE = -1;
 /**
  * A tester is a function that receives an UI schema and a JSON schema and returns a boolean.
+ * Optionally, a root JSON schema can be provided, which will be used for resolving refs in the schema.
  */
-export type Tester = (uischema: UISchemaElement, schema: JsonSchema) => boolean;
+export type Tester = (uischema: UISchemaElement, schema: JsonSchema, rootSchema?: JsonSchema) => boolean;
```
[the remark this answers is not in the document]
AlexandraBuzila: It doesn't need to be optional, but I made it so such that existing client code does not break. Would you prefer to have it as a required prop?

packages/core/src/testers/testers.ts
```diff
@@ -72,8 +74,8 @@ export const isControl = (uischema: any): uischema is ControlElement =>
  *        applied to the resolved sub-schema
  */
 export const schemaMatches = (
-  predicate: (schema: JsonSchema) => boolean
-): Tester => (uischema: UISchemaElement, schema: JsonSchema): boolean => {
+  predicate: (schema: JsonSchema, rootSchema?: JsonSchema) => boolean
```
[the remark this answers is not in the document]
AlexandraBuzila: Same as above, it doesn't need to be optional, but I made it so such that existing client code does not break. Would you prefer to have it as a required prop?

packages/core/src/testers/testers.ts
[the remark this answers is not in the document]
sdirix: Does it really break when it's not optional? Their existing testers will just ignore the additional parameter. Can you give an example which then breaks?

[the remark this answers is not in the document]
AlexandraBuzila: When calling the testers directly, like will do in the tests, e.g. `schemaTypeIs('string')(uischema, schema, schema)`. But I guess this does not happen that often outside our code.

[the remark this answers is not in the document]
sdirix: I see. If it's fine for you I would like to see it made mandatory. When it's not provided it might be masking a bug. Within the code where we finally access the `rootSchema` we can stay defensive and still add a `rootSchema ?? schema` check to make sure we don't crash at runtime.

packages/core/src/util/resolvers.ts
AlexandraBuzila: No, we can directly resolve against the rootSchema here, since we tried resolving against schema beforehand.

AlexandraBuzila:
Thank you for for the review, Stefan! I updated the code and integrated your comments.

sdirix (changes requested):
Looks good ;) Found one more small detail.

sdirix (approved):
Great work! That's a great improvement to JSON Forms!

PR #1834, opened 2021-11-16
Encode '/' in prop name

max-elia:
- Encode '/' in ui schema generator
- Decode '/' and '~' in:
  - Resolvers
  - Label
  - MaterialTableControl
  - Vanilla TableArrayControl
  - Angular table.renderer
- Add test cases for uischema, resolver and path

Signed-off-by: Max Elia Schweigkofler <max_elia@hotmail.de>

### DIFF
--- a/packages/angular-material/src/other/master-detail/master.ts
+++ b/packages/angular-material/src/other/master-detail/master.ts
@@ -33,6 +33,7 @@ import {
   ArrayControlProps,
   ControlElement,
   createDefaultValue,
+  decode,
   findUISchema,
   getFirstPrimitiveProp,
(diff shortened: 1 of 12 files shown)


TheZoker (changes requested):
Thanks for the PR 👍 
For most cases this works well, but as soon as there is an array involved , there seems to be no input anymore:
image

image

max-elia:
Thanks, for the feedback. Array's are now also work fine and tests have been added.

TheZoker (approved):
LGTM 👍 Array example works now

sdirix (approved):
LGTM!

PR #1794, opened 2021-08-03
Disable group elements on group disable rule

max-elia:
- Pass enabled prop to individual group elements
Closes #1586

**To Reproduce:**
```
export const schema = {
  type: 'object',
  properties: {
    name: {
      type: 'string',
      minLength: 3,
      description: 'Please enter your name'
    },
    birthDate: {
      type: 'string',
      format: 'date'
    }
  }
};

export const uischema = {
  type: 'Group',
  label: 'My Group',
  elements: [
    {
      type: 'Control',
      label: 'Name',
      scope: '#/properties/name'
    },
    {
      type: 'Control',
      label: 'Birth Date',
      scope: '#/properties/birthDate'
    }
  ],
  options: {
    readonly: true
  },
};

```
Signed-off-by: Max Elia Schweigkofler <max_elia@hotmail.de>

PR #1835, opened 2021-11-16
add support for disabling radio inputs

GideonPARANOID:
* Added support for disabling radio inputs ([supported as per the spec](https://html.spec.whatwg.org/multipage/input.html#radio-button-state-(type=radio)), though not documented on MDN).


sdirix:
Hi @GideonPARANOID. Thank you very much for the PR! Coincidentally this was already fixed with a [previous PR](https://github.com/eclipsesource/jsonforms/pull/1794) so this one is no longer needed. However you are very welcome to fix other issues too ;)

PR #1836, opened 2021-11-16
Add uischemas to dependency array

TheZoker:
Currently the uischema prop of JSONForms was not watched for changes. So when uischemas was changed after JSONForms was rendered, the change did not get applied


sdirix (approved):
LGTM!

Issue #1840, opened 2021-12-02
Add placeholder support for React Vanilla


andresgutgon:
### Is your feature request related to a problem? Please describe.

It would be great when placeholders could be defined via the ui schema and set in the respective inputs

### Describe the solution you'd like

Allow for the optional attribute placeholder, e.g.
``` javascript
{
  "type": "Control",
  "scope": "#",
  "options": {
    "placeholder": "This is a placeholder"
  }
}
```

### Describe alternatives you've considered

We could also think about checking for a placeholder attribute in the JSON Schema

### Framework

React

### RendererSet

Vanilla

### Additional context

At least for Vue-Vanilla. When adding the possibility of defining a placeholder to the JSON Schema, then we could think about adding this to the JSON Forms core so any renderer set benefits from that.

This issue is a copy of this Vue issue 
https://github.com/eclipsesource/jsonforms/issues/1729 

I guess the implementation would similar.

sdirix:
Hi @andresgutgon, thanks for the suggestion. This could definitely be added to each renderer set. 

> we could think about adding this to the JSON Forms core so any renderer set benefits from that.

We have to think about whether it makes sense to add it to the core framework. Up until recently I would definitely have said no, as there is not much benefit of receiving a `placeholder` prop than just looking it up in the renderer via something like `uischema.options.placeholder`. It's not really a feature affecting the core.

However as we recently introduced translation support it now might make sense to also provide the user a default translated placeholder as the renderer code now not only needs to look up the placeholder but potentially also translate it.

andresgutgon:
> However as we recently introduced translation support 

Just was reading about it! [Rework i18n support in JSON Forms core](https://github.com/eclipsesource/jsonforms/pull/1825). 

Not sure how to i18n my forms though. Not seing documentation related with this

andresgutgon:
I found the way the new `i18n` prop in `jsonforms v.3` is used.

I did a hook where I can use my i18n solution. Which in my case is [react-intl](https://formatjs.io/docs/react-intl/). Then I use the hook on the `JSONForms` `i18n` prop.
The hook
``` javascript
import { useIntl } from 'react-intl'
import { UISchemaElement, Translator } from '@jsonforms/core'
import { ErrorObject } from 'ajv'
export const useTranslateError = () => {
  const intl = useIntl()
  return (
    error: ErrorObject,
    _translate: Translator,
    uischema: UISchemaElement
  ): string => {
    const params = error.params
[...]

PR #1844, opened 2021-12-05
Allow to pass as UISchema options placeholder attribute for TextCell and TextAreaCell cells

andresgutgon:
# What?
Solve this issue https://github.com/eclipsesource/jsonforms/issues/1840

### DIFF
--- a/packages/examples/src/day2.ts
+++ b/packages/examples/src/day2.ts
@@ -1,19 +1,19 @@
 /*
   The MIT License
-  
+
   Copyright (c) 2017-2019 EclipseSource Munich
   https://github.com/eclipsesource/jsonforms
-  
+
   Permission is hereby granted, free of charge, to any person obtaining a copy
   of this software and associated documentation files (the "Software"), to deal
   in the Software without restriction, including without limitation the rights
   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
   copies of the Software, and to permit persons to whom the Software is
   furnished to do so, subject to the following conditions:
-  
+
   The above copyright notice and this permission notice shall be included in
   all copies or substantial portions of the Software.
-  
+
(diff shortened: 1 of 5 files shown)


packages/examples/src/day2.ts
```diff
@@ -32,7 +32,10 @@ export const uischema = {
   elements: [
     {
       type: 'Control',
-      scope: '#/properties/name'
+      scope: '#/properties/name',
+      options: {
+        placeholder: 'Placeholder for name'
+      }
```
sdirix, line 38: We should not modify the `day2` example. It is referred to [by a blog series](https://eclipsesource.com/blogs/2016/12/27/json-forms-day-2-introducing-the-ui-schema/).

packages/vanilla/src/cells/TextAreaCell.tsx
```diff
@@ -44,6 +44,7 @@ export const TextAreaCell = (props: CellProps & VanillaRendererProps) => {
       id={id}
       disabled={!enabled}
       autoFocus={uischema.options && uischema.options.focus}
+      placeholder={uischema.options && uischema.options.placeholder}
```
sdirix, line 47: As you're changing the code here anyway: We should use the same `appliedUiSchemaOptions` approach like in `TextCell.tsx`.

sdirix (changes requested):
Thanks for the PR! Looks good and thanks for adding the test cases!

I only have minor comments.

packages/vanilla/src/cells/TextAreaCell.tsx
andresgutgon: Changing it now...

andresgutgon:
Comments fixed : )

sdirix (approved):
LGTM! Thanks for the contribution!

andresgutgon:
yay! thanks to you : )

andresgutgon:
When something is merged is pushed to npm or do you have release days?

PR #1841, opened 2021-12-03
Add Live-Playground link to README file

mirismaili:
I developed this tool to the order of the company I work in. Other users can use this useful tool too.

Issue #1765, opened 2021-06-19
Support for propertyNames key/value-objects etc


runar-indico:
I am trying to set up a form from an object that is of type `Record<string, string[]>`. I'm not sure if this is already supported. 

I've tried this schema (showing only the relevant part):

```json
{
	"clients": {
		"additionalProperties": {
			"additionalItems": true,
			"description": "One or more valid password that can be used for client-authentication",
			"items": {
				"type": "string"
			},
			"minItems": 1,
			"title": "The clientId schema",
			"type": "array",
			"uniqueItems": true
		},
		"description": "The clients will authenticate via Basic Auth",
		"examples": [
			{
				"clientId": [
					"password1",
					"password2"
				]
			}
		],
		"propertyNames": {
			"maxLength": 100,
			"minLength": 3
		},
		"title": "As an alternative to other authentication, a simple map of client with clientSecrets can be supplied",
		"type": "object"
	}
}
```

Perhaps this only needs a custom renderer? I'm thinking it should act sort of like the Material-renderer does  already with arrays, only that one has to provide the propertyName as well.

sdirix:
Hi @runar-indico, we don't support "dynamic" schemas out of the box, however you can definitely write a custom renderer which can handle this use case.

PR #1845, opened 2021-12-06
Fix type-imports according to the release notes of TypeScript 3.8

mirismaili:
See: [Type-Only Imports and Export](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export)

### DIFF
--- a/packages/core/src/reducers/reducers.ts
+++ b/packages/core/src/reducers/reducers.ts
@@ -32,9 +32,9 @@ import {
 import { defaultDataReducer } from './default-data';
 import { rendererReducer } from './renderers';
 import { JsonFormsState } from '../store';
+import type { JsonFormsUISchemaRegistryEntry } from './uischemas';
 import {
   findMatchingUISchema,
-  JsonFormsUISchemaRegistryEntry,
   uischemaRegistryReducer,
 } from './uischemas';
 import {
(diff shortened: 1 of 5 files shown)


sdirix:
Hi @mirismaili, thanks for the PR. Can you tell about the motivation? Do you run into problems when not refactoring those specific type imports/exports?

mirismaili:
Hi @sdirix 

Your welcome.

I was trying to add some unsupported features to the library (`additionalProperties` and `patternProperties` / #1765) for our business. So I created a dev-project that I can test my changes into the library, quickly (using hot-reload). 

My different webpack/typescript configuration (at the dev-project) causes some issues. I fixed them (there) to match JSONForms. But this special case was an issue in JSONForms. So I decided to fix it here.
***
I also created a playground for JSONForms (#1841).

mirismaili:
I think the below configuration can be changed after this fix (be applied to all packages):

https://github.com/eclipsesource/jsonforms/blob/6bcdbb171846ba19121aa3f93d087778611dde7e/webpack/webpack.base.js#L25

But I'm not sure.

sdirix:
@mirismaili The PR conflicts now, can you rebase on latest master?

mirismaili:
@sdirix 
Rebased.

sdirix (approved):
LGTM!

PR #1848, opened 2021-12-09
Use default calculated i18n keys

sdirix:
If there is no 'i18n' key in schema or ui schema a default
calculated key based on the data path of the respective control
is used.

Additionally fixes the issue that the 'required' message of
localized errors were overwritten by JSON Forms.

Also removed redundant exports in JSON Forms core and
cleans up imports in test cases which lead to errors when
executing ava.


AlexandraBuzila (approved):
Looks good to me and works as expected!

PR #1858, opened 2021-12-22
Build ESM and CommonJS modules 

sdirix:
Use Rollup to bundle ESM and CommonJS modules for all non-Angular
packages. Angular builds are done via ngc.

ESM bundles target 'esnext' while CommonJS bundles continue to
target ES5.

Also adjusts Vue dependencies so that Vue and the respective
template compiler are of the same version.

Also removes the React Material extended renderer set


AlexandraBuzila (approved):
I did not test the changes, but the code looks good to me!

Issue #1849, opened 2021-12-11
Is "dot-separated" path safe?


mirismaili:
### Describe the bug

As you know there is a `path` prop that is passed to the renderer that indicates the location of `data` (that should be rendered by that renderer) in the root `data` object (e.g. `'foo.bar'`).

It's a **dot-separated string**:

https://github.com/eclipsesource/jsonforms/blob/81f1f19dac49b1864e712e9abbc5c5c6355a2b91/packages/core/src/util/path.ts#L82-L90

```js
compose(path, segments.join('.'))
```

_**Is this safe? Then `'foo.bar'` points to `data.foo.bar` or `data['foo.bar']`?**_

### Expected behavior

The library should can distinguish between these paths. For example, using arrays (instead of strings):
```js
path1 = ['foo', 'bar']
path2 = ['foo.bar']
```

### Steps to reproduce the issue

 1. Go to [Playground](https://mirismaili.github.io/jsonforms-playground)
 
2. Set:
    `JSON-Schema`:
    ```json
    {
      "type": "object",
      "properties": {
        "my.name": {
          "type": "string"
        }
      }
    }
    ```
    `UI-Schema`:
    ```json
    false
    ```
    `Data`:
    ```json
    {
      "my.name": "S. Mahdi",
      "my": {
        "name": "MirIsmaili"
      }
    }
    ```

As you can see in the [playground](https://mirismaili.github.io/jsonforms-playground/?hideinfo), the synchronization process flows along with **the green arrows** (see below screenshot), but NOT along with **the red arrows**. _(When you edit the form, `data['my.name']` will be updated (only) and when you edit `data.my.name` in the editor (only), the form will be updated)._

### Screenshots

![image](https://user-images.githubusercontent.com/15800189/145674684-4b53db88-2031-4e0e-9f65-4bb8ae348c95.png)

### In which browser are you experiencing the issue?

Google Chrome v96.0.4664.93 (Official Build) (64-bit)

sdirix:
Hi @mirismaili, thanks for the report. The issue is [already reported](https://github.com/eclipsesource/jsonforms/issues/1631) and we're already on a fix, see #1831. However the change is non trivial as using an array as a prop triggers rerenders by default in all bindings. Therefore we want to make sure that we don't introduce any regressions.

mirismaili:
Hi @sdirix, 

[...]

PR #1862, opened 2021-12-29
Fix "dot-separated" paths by making paths type `string[]` (instead of `string`)

mirismaili:
Fix "dot-separated" paths by making paths type `string[]` (instead of `string`) for react (in these packages: **"core"**, **"react"** and **"material-renderers"**)

eclipsesource/jsonforms#1849
eclipsesource/jsonforms#1831

### DIFF
--- a/packages/core/src/actions/actions.ts
+++ b/packages/core/src/actions/actions.ts
@@ -72,7 +72,7 @@ export type CoreActions =
 
 export interface UpdateAction {
   type: 'jsonforms/UPDATE';
-  path: string;
+  path: string[];
   updater(existingData?: any): any;
 }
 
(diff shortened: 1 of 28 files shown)


mirismaili:
I know this can't be merged now. Because of conflicts with unfixed packages (angular, vue, ...). But those can be fixed, as well.

sdirix:
Hi @mirismaili, thanks for the PR. We're already working on this, see https://github.com/eclipsesource/jsonforms/pull/1831

The main problem with the array approach is to not lose performance. Just replacing strings with arrays here will lead to massive degradation as the memoization of React will no longer work and therefore the whole form will rerender with every change.

lepffm:
Hi, Has the "dot-seperated" issue been resolved yet?

sdirix:
We came up with a better solution, documented it in #2153 and will start developing it soon. Therefore I'll close this PR as it's outdated.

PR #1865, opened 2022-01-10
Add special className to style boolean cells (checkbox) in Vanilla package

andresgutgon:
# What?
Hi I was using the lib and I noticed Vanilla React has the same styles for `input type='text'` and `input type='checkbox'`. Would be possible to have different styles? This way is easier to style it

## Order
Also noticed when is a checkbox boolean the position is the same as with `type=text`
Now:
``` html
<label>My label</label>
<input type='checkbox' />
``` 
Wouldn't be better to put the cell inside the label? 

``` html
<label>
  <input type='checkbox' />
  <div>My label</div>
</label>
```
I think is a common practice for checkboxes 

Changing the order checkbox can look like this
image

Thanks!

### DIFF
--- a/packages/vanilla/src/cells/BooleanCell.tsx
+++ b/packages/vanilla/src/cells/BooleanCell.tsx
@@ -1,19 +1,19 @@
 /*
   The MIT License
-  
+
   Copyright (c) 2017-2019 EclipseSource Munich
   https://github.com/eclipsesource/jsonforms
-  
+
   Permission is hereby granted, free of charge, to any person obtaining a copy
   of this software and associated documentation files (the "Software"), to deal
   in the Software without restriction, including without limitation the rights
   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
   copies of the Software, and to permit persons to whom the Software is
   furnished to do so, subject to the following conditions:
-  
+
   The above copyright notice and this permission notice shall be included in
   all copies or substantial portions of the Software.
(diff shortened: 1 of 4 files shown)


sdirix:
Thanks for the contribution, we'll take a look very soon! :heart:

andresgutgon:
Nice! Thanks @sdirix what do you think about the order of `label/input` when is a boolean/checkbox?

sdirix:
> Nice! Thanks @sdirix what do you think about the order of `label/input` when is a boolean/checkbox?
> 
> At the moment only added the class to style the checkbox. But the order remains the same as with `input`

Can you point me to some occurrences where the input is defined within a label? I'm not aware of that pattern. The [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox) also doesn't showcase it.

sdirix (changes requested):
Can you take a look at the failing test cases?

andresgutgon:
> Can you point me to some occurrences where the input is defined within a label?

Uhmm maybe is just me. I always do checkbox/radio this way. But maybe is not something standard. Feel free to ignore.

andresgutgon:
I think I fixed the specs, let me know if it works. Thanks!

sdirix (approved):
LGTM!

sdirix:
Thanks again for the contribution ;)

PR #1866, opened 2022-01-10
Use validationError classes defined in styles. This way this can be ovewriten by users

andresgutgon:
# What? 
This one is pretty easy. In text inputs errors can be styled. 

### Example in my code
image

But Radio buttons have a hardcoded string so is not possible. 

### Solution 
Just use the prop we have for this clase

### DIFF
--- a/packages/vanilla/src/controls/RadioGroup.tsx
+++ b/packages/vanilla/src/controls/RadioGroup.tsx
@@ -58,7 +58,9 @@ export const RadioGroup = ({
   const radioInput = useMemo(() => findStyleAsClassName(contextStyles)('control.radio.input'), [contextStyles]);
   const radioLabel = useMemo(() => findStyleAsClassName(contextStyles)('control.radio.label'), [contextStyles]);
   const isValid = errors.length === 0;
-  const divClassNames = `validation  ${isValid ? classNames.description : 'validation_error' }`;
+  const divClassNames = [classNames.validation]
+    .concat(isValid ? classNames.description : classNames.validationError)
+    .join(' ');
   const appliedUiSchemaOptions = merge({}, config, uischema.options);
   const showDescription = !isDescriptionHidden(
     visible,


packages/vanilla/src/controls/RadioGroup.tsx
```diff
@@ -58,7 +58,7 @@ export const RadioGroup = ({
   const radioInput = useMemo(() => findStyleAsClassName(contextStyles)('control.radio.input'), [contextStyles]);
   const radioLabel = useMemo(() => findStyleAsClassName(contextStyles)('control.radio.label'), [contextStyles]);
   const isValid = errors.length === 0;
-  const divClassNames = `validation  ${isValid ? classNames.description : 'validation_error' }`;
+  const divClassNames = `validation  ${isValid ? classNames.description : classNames.validationError }`;
```
andresgutgon, line 61: Do the same as in [the input control](https://github.com/eclipsesource/jsonforms/blob/master/packages/vanilla/src/controls/InputControl.tsx#L70)

packages/vanilla/src/controls/RadioGroup.tsx
```diff
@@ -58,7 +58,9 @@ export const RadioGroup = ({
   const radioInput = useMemo(() => findStyleAsClassName(contextStyles)('control.radio.input'), [contextStyles]);
   const radioLabel = useMemo(() => findStyleAsClassName(contextStyles)('control.radio.label'), [contextStyles]);
   const isValid = errors.length === 0;
-  const divClassNames = `validation  ${isValid ? classNames.description : 'validation_error' }`;
+  const divClassNames = [classNames.validation]
+    .concat(isValid ? classNames.description : classNames.validationError)
+    .join(' ');
```
andresgutgon, line 63: Do the same as in [the input control](https://github.com/eclipsesource/jsonforms/blob/master/packages/vanilla/src/controls/InputControl.tsx#L70)

sdirix:
Thanks for this contribution too :wink:   We'll take a look very soon!

sdirix (approved):
Agreed! Thanks for the contribution!

PR #1869, opened 2022-01-14
Update Angular packages

sdirix:
- Fixes the module declaration in @jsonforms/angular so the ESM
  build can be picked up reliably
- Exports all functions used in Angular components to avoid
  name clashes by the ngc transpiler
- Removes the redundant zonejs peer dependency
- Indicate compatibility with Angular 13
- Adapts i18n change detection to allow updating i18n via the
  JsonFormsAngularService independent from the root component
  props

### DIFF
--- a/packages/angular-material/package.json
+++ b/packages/angular-material/package.json
@@ -54,20 +54,19 @@
     ]
   },
   "peerDependencies": {
-    "@angular/animations": "^12.0.0",
-    "@angular/cdk": "^12.0.0",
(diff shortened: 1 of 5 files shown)


ttr7usb (approved):

sdirix:
Hi @ttr7usb! Thanks for taking a look. Did you test the changes? What exactly did you test?

AlexandraBuzila (approved):
The code changes LGTM. I did not test them however.

PR #1876, opened 2022-01-25
Add release CI with changelog generator

TheZoker:
Add github action, that creates a changelog automatically on every release.

Example changelog (packages are missing for most PRs, since the they don't have any labels yet):
https://github.com/TheZoker/jsonforms/releases/tag/v3.0.0

Note: It id recommended, that every PR since the last release is tagged with the fitting lable. So for changes in the core, the PR should get the label `core` and so on...
This will lead to the desired output of: "[core] PR Title (#pr number)"


sdirix:
Will this also add release notes to our prereleases? How can we differentiate between prerelase and regular release?

PR #1877, opened 2022-01-25
Enhance example app

TheZoker:
Enhance the example app with the following features:
- Show current uischema (with code highlighting)
- Show current schema (with code highlighting)
- Option to show/hide data/schema/uischema panel
- Allow to link to a certain example
- Use same example on page reload
- Enhance example design

Screenshot:
image

### DIFF
--- a/packages/example/src/App.tsx
+++ b/packages/example/src/App.tsx
@@ -33,13 +33,20 @@ import {
 } from '@jsonforms/core';
 import { resolveRefs } from 'json-refs';
 import { Tab, Tabs, TabList, TabPanel } from 'react-tabs';
+import Highlight from 'react-highlight';
+import 'highlight.js/styles/default.css';
 import './App.css';
 
 type AppProps = {
   examples: ExampleDescription[],
   cells: JsonFormsCellRendererRegistryEntry[],
   renderers: JsonFormsRendererRegistryEntry[],
-}
+};
+
+type Action = {
+  label: string,
+  apply: any
+};
 
 const ResolvedJsonForms = (
   props: JsonFormsInitStateProps & JsonFormsReactProps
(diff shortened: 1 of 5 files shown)


packages/example/src/App.tsx
```diff
@@ -83,63 +91,145 @@ const App = ({ examples, cells, renderers}: AppProps) => {
...
+  }, [showPanel]);
+
   const changeData = (data: any) => {
     setDataAsString(JSON.stringify(data, null, 2));
   };
 
+  const togglePanelHeight = () => {
+    var panel = document.getElementsByClassName('panel-wrapper')[0];
+    panel.classList.toggle('full-height');
+  }
```
sdirix, line 151: This could also be done by handing over different strings to `className` of the panel instead of direct DOM manipulation, i.e.
```
<div className={`someStaticClass ${fullPanelHeight ? 'full-height' : ''}`>
```
It requires managing an additional state variable but that's fine.

sdirix (changes requested):
Thanks for these changes! In general I like them very much!

 - Please remove the `Example` title over the form, it's not needed
 - Can't we just always expand the left schema/uischema/data box? If not, then please replace the expand link with a button
 - I would prefer the form not having a background color

packages/example/src/App.tsx
```diff
@@ -101,45 +126,79 @@ const App = ({ examples, cells, renderers}: AppProps) => {
...
+                    label={optionValue.label}
+                    key={index}
+                  >
+                    {optionValue.label}
+                  </option>
+                ))}
+              </select>
+            </div>
+            <div className='toggle-panel'>
+              <input type='checkbox' id='panel' name='panel' checked={showPanel} onChange={() => setShowPanel(!showPanel)} />
```
sdirix, line 154: Whenever a state update depends on the previous value of the state, the functional update should be used. This avoids problems with outdated variables/props.

```suggestion
              <input type='checkbox' id='panel' name='panel' checked={showPanel} onChange={() => setShowPanel(prevShow => !prevShow)} />
```

sdirix (approved):
Looks good! Thanks for the improvements!

PR #1879, opened 2022-01-26
Fix useMemo() for translation props

TheZoker:
Until now select options where not translated after changing the translation, this should fix this


sdirix (approved):
Thanks for the changes!

PR #1909, opened 2022-03-31
Add support for conditional schema compositions

### LINKED ISSUE #1907: Control path resolution fails for "dynamically" added properties via conditional schema compositions
### Describe the bug

Imagine having the following two schemas (i used the vue3 seeder repository as a basis):
```JavaScript
const schema = {
    type: 'object',
    properties: {
        name: {
            type: 'string',
            minLength: 1,
            description: 'The task\'s name',
        },
        recurrence: {
            type: 'string',
            enum: ['Never', 'Daily', 'Weekly', 'Monthly'],
        },
    },
    anyOf: [
        {
            if: {
                properties: {
                    recurrence: {
                        const: 'Never'
                    }
                }
            },
            then: {
                properties: {
                    lastname: {
                        type: 'string'
                    }
                }
            }
        },
    ]
};

const uischema = {
    type: 'HorizontalLayout',
    elements: [
        {
            type: 'VerticalLayout',
            elements: [
                {
                    type: 'Control',
                    scope: '#/properties/name',
                },
                {
                    type: 'Control',
                    scope: '#/properties/recurrence',
                },
                {
                    type: 'Control',
                    scope: '#/anyOf/0/then/properties/lastname',
                    rule: {
                        effect: 'SHOW',
                        condition: {
                            scope: '#/properties/recurrence',
                            schema: {
                                const: 'Never'
                            }
                        }
                    }
                },
            ],
        },
    ],
};
```
Just for reference: I used the following json-schema documentation to build such a construct [json-schema.org/...](https://json-schema.org/understanding-json-schema/reference/conditionals.html). See the second example block underneath `If-Then-Else`. My intention is to add an arbitrary set of properties (here just: `lastname`) to a schema if some condition is met (here: if `recurrence` equals `'Never'`).

The actual issue is the following: Every Control-Renderer receives a property called `control` which contains the necessary information for my current form element. This `control` also holds a property `path` which is derived from the `scope`-property of the ui-schema and is further used as the key for the data object to be created via the form. 
E.g.: `scope: '#/properties/name'` -> `control.path: 'name'` -> `{ name: 'Jon', ... }`.
For the property which is only added if a certain condition is met, my `scope`-property is the following: `#/anyOf/0/then/properties/lastname`. If I now inpsect via vue-devtools what my `control.path` holds for my actually rendered control of `#/anyOf/0/then/properties/lastname` it equals: `control.path: 'properties'`.

### Expected behavior

`control.path` holds the correctly transformed value: `lastname`.

1. Checkout and run the vue3 [seed repo](https://github.com/eclipsesource/jsonforms-vue-seed)
2. Replace the two sample schemas with the ones from above
3. Inspect the `lastname` field via vue-devtools
4. See the wrong value of `control.path`

### Screenshots

_No response_

Google Chrome

### Framework

Core, Vue 3

### RendererSet

Vanilla

### Additional context

_No response_

nmaier95:
If properties of a schema were added conditionally inside
an anyof|oneof|allof list their paths were not resolved
correctly.
For reference see second example in:
https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else

resolves #1907


sdirix:
Thanks for the PR. We'll have a look soon!

sdirix (commented):
Hi @nmaier95!

Thanks a lot for the PR, it makes a lot of sense to me!

Note that we also support skipping `oneOf/allOf/anyOf` within the UI Schema scopes, see [here](https://github.com/eclipsesource/jsonforms/blob/7e5ac137297c9ff16cb2edfad26f63d90ebb0dba/packages/core/src/util/resolvers.ts#L132-L150). Would you be interested in something similar for the `if/then/else` parts? If yes, would you be willing to also contribute this?

nmaier95:
Hi @sdirix,

thank you for your feedback! I'll add an example.

Furthermore I'll have a look at the support of skipping the `if/then/else` part inside the `scope` of an ui-schema element.

packages/core/src/util/resolvers.ts
```diff
@@ -134,13 +134,29 @@ export const resolveSchema = (
       const schemas = [].concat(
         resultSchema?.oneOf ?? [],
         resultSchema?.allOf ?? [],
-        resultSchema?.anyOf ?? []
+        resultSchema?.anyOf ?? [],
+        // also add root level schema composition entries
+        schema?.oneOf ?? [],
+        schema?.allOf ?? [],
+        schema?.anyOf ?? []
```
sdirix, line 141: Can you explain the use cases where this is necessary?

sdirix (commented):
In general this looks very good and I would like to merge this soon. However there is one detail I would like to have explained.

packages/core/src/util/resolvers.ts
nmaier95: I'll try to explain this in englisch. I hope it is comprehensible.

The `resultSchema` variable already holds the first resolved path entry. Which mostyl is `properties`.
If we take for example, the example given in the corresponding test-file `resolvers.test.ts` there I defined an `anyOf` in the root of the schema. The problem was that this code would now search further inside `properties` for the scopes defined inside `anyOf` - where it would not find those. That is why we need to also add the root level compositions here.

sdirix (changes requested):
Looking at the code in more detail I think we could do better by generalizing the code a bit. Also the resolving could easier handle all cases by transforming it to a depth-first search.

sdirix:
@nmaier95 I have a fix for my requests ready. Please allow [edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) so I can add them to this PR.

Otherwise I'll have to close this one and reopen a new PR.

nmaier95:
> @nmaier95 I have a fix for my requests ready. Please allow [edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) so I can add them to this PR.
> 
> Otherwise I'll have to close this one and reopen a new PR.

It should be allowed now.

packages/core/test/util/resolvers.test.ts
```diff
@@ -25,7 +25,7 @@
 import { resolveSchema } from '../../src/util/resolvers';
 import test from 'ava';
 
-test('resolveSchema - resolves schema with any ', t => {
+test.only('resolveSchema - resolves schema with any ', t => {
```
AlexandraBuzila, line 28: `only` should be removed from here :smile:

AlexandraBuzila (changes requested):
Thank you for the changes, they look good to me. 

There's only one small comment in one of the tests that should be addressed.

packages/core/test/util/resolvers.test.ts
sdirix: oops ;)

AlexandraBuzila (approved):
Thanks, changes look good now!

sdirix (approved):

PR #1919, opened 2022-04-15
(vue) add binding for one enum cell elements

jdwit:
The control bindings for Vue where missing the one for cells with a `oneOf` schema; no `options` prop was generated. For control bindings this is present, this PR adds the binding for cells. If you have any feedback please let me know, I would like to contribute more in the future since we are investing in this project for our application that will soon be in production.

### DIFF
--- a/packages/vue/vue/src/jsonFormsCompositions.ts
+++ b/packages/vue/vue/src/jsonFormsCompositions.ts
@@ -396,6 +396,22 @@ export const useJsonFormsEnumCell = (props: ControlProps) => {
 };
 
 /**
+ * Provides bindings for 'oneOf' enum cell elements. Cells are meant to show simple inputs,
+ * for example without error validation, within a larger structure like tables.
+ *
+ * Access bindings via the provided reactive 'cell' object.
+ * Dispatch changes via the provided `handleChange` method.
+ */
+export const useJsonFormsOneOfEnumCell = (props: ControlProps) => {
+  const { control, ...other } = useControl(
+      props,
+      mapStateToOneOfEnumControlProps,
+      mapDispatchToControlProps
+  );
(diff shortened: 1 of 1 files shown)


yaffol:
Just a note in case it's helpful to you - there is also a Vue/Vuetify renderer set in development (though usable now) over at https://github.com/eclipsesource/jsonforms-vuetify-renderers

packages/vue/vue/src/jsonFormsCompositions.ts
```diff
@@ -395,6 +395,22 @@ export const useJsonFormsEnumCell = (props: ControlProps) => {
...
+ * Provides bindings for 'oneOf' enum cell elements. Cells are meant to show simple inputs,
+ * for example without error validation, within a larger structure like tables.
+ *
+ * Access bindings via the provided reactive 'cell' object.
+ * Dispatch changes via the provided `handleChange` method.
+ */
+export const useJsonFormsOneOfEnumCell = (props: ControlProps) => {
+  const { control, ...other } = useControl(
+      props,
+      mapStateToOneOfEnumControlProps,
```
sdirix, line 408: There is a different mapping for cells, so this should be `mapStateToOneOfEnumCellProps`

sdirix (changes requested):
Hi @jdwit,

Thank you very much for the PR! In general we're definitely interested in any improvements to our bindings and renderers. If you want to tackle something which requires more effort and you want to be certain that we'll actually merge it, I would like to recommend to open a feature request first, describe your planned implementation and wait for our feedback.

However things like this which is just a missing binding certainly don't require a previous discussion ;)

sdirix (approved):
LGTM! Thanks!

PR #1935, opened 2022-05-15
Add additional interfaces for Scoped, Labelable and Labeled ui elements.

wienczny:
Make Scopable scope optional and use Scoped with required scope where required.
Add type predicates for Scopeable, Scoped, Labelable and Labeled.
Duplicate old functions handling Scopeable to have Scopeable and Scoped where required.
Update toDataPathSegments to allow nullable schemaPath and return an empty array if this happens.

### DIFF
--- a/packages/core/src/models/uischema.ts
+++ b/packages/core/src/models/uischema.ts
@@ -26,14 +26,39 @@
 import { JsonSchema } from './jsonSchema';
 
 /**
+ * Interface for describing an UI schema element that may reference
+ * a subschema. The value of the scope may be a JSON Pointer or null.
+ */
+export interface Scopable {
+  /**
+   * The scope that determines to which part this element may be bound to.
+   */
+  scope?: string;
+}
+
+/**
  * Interface for describing an UI schema element that is referencing
  * a subschema. The value of the scope must be a JSON Pointer.
  */
-export interface Scopable {
+export interface Scoped extends Scopable {
   /**
    * The scope that determines to which part this element should be bound to.
    */
   scope: string;
+}
+
+/**
+ * Interface for describing an UI schema element that may be labeled.
+ */
+export interface Lableable<T = string> {
+  label?: string|T;
+}
+
+/**
+ * Interface for describing an UI schema element that is labeled.
+ */
+export interface Labeled<T = string> extends Lableable<T> {
+  label: string|T;
 }
 
 /**
@@ -87,7 +112,7 @@ export interface Condition {
 /**
  * A leaf condition.
  */
-export interface LeafCondition extends Condition, Scopable {
+export interface LeafCondition extends Condition, Scoped {
   type: 'LEAF';
 
   /**
@@ -96,7 +121,7 @@ export interface LeafCondition extends Condition, Scopable {
   expectedValue: any;
 }
 
-export interface SchemaBasedCondition extends Condition, Scopable {
+export interface SchemaBasedCondition extends Condition, Scoped {
   schema: JsonSchema;
 }
 
@@ -170,12 +195,8 @@ export interface HorizontalLayout extends Layout {
  * A group resembles a vertical layout, but additionally might have a label.
  * This layout is useful when grouping different elements by a certain criteria.
  */
-export interface GroupLayout extends Layout {
+export interface GroupLayout extends Layout, Lableable {
   type: 'Group';
-  /**
-   * The label of this group layout.
-   */
-  label?: string;
 }
 
 /**
@@ -207,23 +228,15 @@ export interface LabelElement extends UISchemaElement {
  * A control element. The scope property of the control determines
  * to which part of the schema the control should be bound.
  */
-export interface ControlElement extends UISchemaElement, Scopable {
+export interface ControlElement extends UISchemaElement, Scopable, Lableable<string | boolean | LabelDescription> {
   type: 'Control';
-  /**
-   * An optional label that will be associated with the control
-   */
-  label?: string | boolean | LabelDescription;
 }
 
 /**
  * The category layout.
  */
-export interface Category extends Layout {
+export interface Category extends Layout, Labeled {
   type: 'Category';
-  /**
-   * The label associated with this category layout.
-   */
-  label: string;
 }
 
 /**
@@ -231,12 +244,8 @@ export interface Category extends Layout {
  * A child element may either be itself a Categorization or a Category, hence
  * the categorization element can be used to represent recursive structures like trees.
  */
-export interface Categorization extends UISchemaElement {
+export interface Categorization extends UISchemaElement, Labeled {
   type: 'Categorization';
-  /**
-   * The label of this categorization.
-   */
-  label: string;
   /**
    * The child elements of this categorization which are either of type
    * {@link Category} or {@link Categorization}.
@@ -249,3 +258,15 @@ export const isGroup = (layout: Layout): layout is GroupLayout =>
 
 export const isLayout = (uischema: UISchemaElement): uischema is Layout =>
   (uischema as Layout).elements !== undefined;
+
+export const isScopeable = (obj: object): obj is Scopable =>
+  obj !== undefined && obj.hasOwnProperty('scope');
+
+export const isScoped = (obj: object): obj is Scoped =>
+  isScopeable(obj) && obj.scope !== null;
+
+export const isLabelable = (obj: object): obj is Lableable =>
+  obj !== undefined && obj.hasOwnProperty('label');
+
+export const isLabeled = (obj: object): obj is Labeled =>
+  isLabelable(obj) && obj.label !== null;
--- a/packages/core/src/util/path.ts
+++ b/packages/core/src/util/path.ts
@@ -52,10 +52,14 @@ export { compose as composePaths };
  * and de-referencing the single segments to obtain a new object.
  *
  *
- * @param {string} schemaPath the schema path to be converted
+ * @param {string?} schemaPath the schema path to be converted
  * @returns {string[]} an array containing only non-schema-specific segments
  */
-export const toDataPathSegments = (schemaPath: string): string[] => {
+export const toDataPathSegments = (schemaPath?: string): string[] => {
+  if (schemaPath === null) {
+    return [];
+  }
+
   const s = schemaPath
     .replace(/anyOf\/[\d]\//g, '')
     .replace(/allOf\/[\d]\//g, '')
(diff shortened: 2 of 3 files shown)


packages/core/src/models/uischema.ts
```diff
@@ -25,17 +25,42 @@
...
+ * Interface for describing an UI schema element that may reference
+ * a subschema. The value of the scope may be a JSON Pointer or null.
+ */
+export interface Scopable {
+  /**
+   * The scope that determines to which part this element may be bound to.
+   */
+  scope?: string;
+}
+
```
sdirix, line 38: If `null` is a foreseen value then this should be reflected in the typings, i.e. `scope?: string | null`. However we don't have any other nullebale types in JSON Forms, so I'm a bit hesitant to introduce this now. Is there a specific use case you have in mind?

This new interface with the optional scope is only used in one place: The new replacement for `Scopable` called `Scoped`. What is the use case of an element with optional scope?

packages/core/src/models/uischema.ts
```diff
@@ -249,3 +258,15 @@ export const isGroup = (layout: Layout): layout is GroupLayout =>
 
 export const isLayout = (uischema: UISchemaElement): uischema is Layout =>
   (uischema as Layout).elements !== undefined;
+
+export const isScopeable = (obj: object): obj is Scopable =>
+  obj !== undefined && obj.hasOwnProperty('scope');
```
sdirix, line 263: This does not fit. In `Scopable` the `scope` is optional. So why check for it here?

packages/core/src/models/uischema.ts
```diff
@@ -249,3 +258,15 @@ export const isGroup = (layout: Layout): layout is GroupLayout =>
...
+
+export const isScoped = (obj: object): obj is Scoped =>
+  isScopeable(obj) && obj.scope !== null;
```
sdirix, line 266: As mentioned above, the documentation says that `null` is an expected value, so it should not be excluded here.

packages/core/src/models/uischema.ts
```diff
@@ -249,3 +258,15 @@ export const isGroup = (layout: Layout): layout is GroupLayout =>
...
+
+export const isLabelable = (obj: object): obj is Lableable =>
+  obj !== undefined && obj.hasOwnProperty('label');
```
sdirix, line 269: Same here. `label` is optional and should therefore not be checked.

packages/core/src/models/uischema.ts
```diff
@@ -249,3 +258,15 @@ export const isGroup = (layout: Layout): layout is GroupLayout =>
...
+
+export const isLabeled = (obj: object): obj is Labeled =>
+  isLabelable(obj) && obj.label !== null;
```
sdirix, line 272: Here the inconsistent becomes apparent. Why allow `null` values for `scope` but not for `label`.

packages/core/src/util/path.ts
```diff
@@ -52,10 +52,14 @@ export { compose as composePaths };
...
  *
- * @param {string} schemaPath the schema path to be converted
+ * @param {string?} schemaPath the schema path to be converted
  * @returns {string[]} an array containing only non-schema-specific segments
  */
-export const toDataPathSegments = (schemaPath: string): string[] => {
+export const toDataPathSegments = (schemaPath?: string): string[] => {
+  if (schemaPath === null) {
+    return [];
+  }
```
sdirix, line 61: You can see that the typings don't match. The parameter is typed ad `string | undefined` but now we also check for `null`. This should be avoided.

packages/core/src/models/uischema.ts
```diff
@@ -249,3 +258,15 @@ export const isGroup = (layout: Layout): layout is GroupLayout =>
 
 export const isLayout = (uischema: UISchemaElement): uischema is Layout =>
   (uischema as Layout).elements !== undefined;
+
```
sdirix, line 261: These new type guards are not used anywhere.

sdirix (changes requested):
Thanks for the pull request!

In general it definitely makes sense to me to improve our typings whenever possible.

Can you explain your motivation for this change? Especially refactoring a long standing existing interface like `Scopable` which could introduce backward compatibility problems.

packages/core/src/models/uischema.ts
wienczny: The case for this change would be the option to add scopes to existing classes without breaking the API in the future. This allows adding the interface to Categorization and Category in a second step to enable relative scopes in their children.

packages/core/src/models/uischema.ts
wienczny: You are right it. The check is only useful if was

interface Scopable {
scope: string|null;
}

packages/core/src/models/uischema.ts
wienczny: See above answer. This might be obj.scope !== undefined then

packages/core/src/models/uischema.ts
wienczny: See above

packages/core/src/models/uischema.ts
wienczny: It should be checked against undefined here too.

packages/core/src/util/path.ts
wienczny: You are right

wienczny (commented):
Did you have time to check my changes? Are there still open issues?

sdirix (changes requested):
I talked with the team. From our point of view it makes sense to support `Scopable, `Scoped`, `Labelable` and `Labeled`. Including the backward compatibility break of changing the `Scopable` interface. We're shortly before the 3.0 release, so that's the perfect point to get this change still in.

If you clean up the PR to only include the interface changes and the type guards we're going to merge it :+1:

packages/core/src/models/uischema.ts
```diff
@@ -36,6 +36,20 @@ export interface Scopable {
   scope: string;
 }
 
+/**
+ * Interface for describing an UI schema element that may be labeled.
+ */
+export interface Lableable<T = string> {
```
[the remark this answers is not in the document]
wienczny: I tried `void` before. This results in errors like this:

`
Error: jsonforms/packages/vanilla/src/complex/categorization/CategorizationList.tsx(56)
 Type 'string | void' is not assignable to type 'Key'.
  Type 'void' is not assignable to type 'Key'.
Error: jsonforms/packages/vanilla/src/complex/categorization/CategorizationList.tsx(59)
 Type 'string | void' is not assignable to type 'ReactNode'.
  Type 'void' is not assignable to type 'ReactNode'.
Error: jsonforms/packages/vanilla/src/complex/categorization/CategorizationList.tsx(73)
 Type 'string | void' is not assignable to type 'Key'.
Error: jsonforms/packages/vanilla/src/complex/categorization/CategorizationList.tsx(77)
 Type 'string | void' is not assignable to type 'ReactNode'.
Error: jsonforms/packages/vanilla/src/layouts/GroupLayout.tsx(70)
 Type 'string | void' is not assignable to type 'ReactNode'.
`

So `void` must not be inside the union type. To have at least `string` inside the union I had to use the `string | string`which results in `string`.

packages/core/src/util/path.ts
```diff
@@ -52,10 +52,15 @@ export { compose as composePaths };
...
- * @param {string} schemaPath the schema path to be converted
+ * @param {string?} schemaPath the schema path to be converted
  * @returns {string[]} an array containing only non-schema-specific segments
  */
-export const toDataPathSegments = (schemaPath: string): string[] => {
+export const toDataPathSegments = (schemaPath?: string): string[] => {
+  if (!schemaPath) {
+    // handle undefined, null and ''
+    return [];
+  }
```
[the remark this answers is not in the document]
wienczny: This change was done to keep the old API for external users. This way the function can be used with Scoped and Scopeable without any changes to the caller. You are right that might hide problems - this depends on how the results are used and if empty arrays are correctly handled there. The question is are there many external users?!?

For me this is not an issue. You can always do a typecheck before calling ;-)

packages/core/src/util/path.ts
```diff
@@ -86,6 +81,10 @@ export const toDataPath = (schemaPath: string): string => {
     };
 
 export const composeWithUi = (scopableUi: Scopable, path: string): string => {
+  if (!isScoped(scopableUi)) {
+    return '';
+  }
```
sdirix, line 86: Seems weird to return `''` here when later in the code we return `path` when the scope was empty.

The could could be clearer by
```
export const composeWithUi = (scopableUi: Scopable, path: string): string => {
  if (!isScoped(scopableUi)) {
    return path ?? '';
  }

  const segments = toDataPathSegments(scopableUi.scope);

  if (isEmpty(segments)) {
    return path ?? '';
  }

  return compose(path, segments.join('.'));
};
```

packages/core/src/models/uischema.ts
```diff
@@ -249,7 +260,13 @@ export const isLayout = (uischema: UISchemaElement): uischema is Layout =>
...
 
 export const isScopeable = (obj: object): obj is Scopable =>
-  obj !== undefined && obj.hasOwnProperty('scope') && (obj as Scopable).scope !== undefined;
+  obj !== undefined && obj.hasOwnProperty('scope');
+
+export const isScoped = (obj: object): obj is Scoped =>
+  isScopeable(obj) && obj.scope !== undefined;
+
+export const isLabelable = (obj: object): obj is Lableable =>
+  obj !== undefined && obj.hasOwnProperty('label');
```
sdirix, line 269: This check should be adapted similarly to `Scopable`

sdirix (changes requested):

sdirix (approved):

MIGRATION.md
```diff
@@ -104,6 +104,11 @@ There should not be any behavior changes.
 All React Material class components were refactored to functional components.
 Please check whether you extended any of our base renderers in your adaptation.
 
+### Scopable interface change
+
+The `scope` attribute in `Scopable` is now optional.
+Use `Scoped` instead for non optional scopes.
+
```
sdirix, line 111:
```suggestion
The utility function `fromScopable` was renamed to `fromScoped` accordingly.
```

MIGRATION.md
```diff
@@ -104,6 +104,11 @@ There should not be any behavior changes.
 All React Material class components were refactored to functional components.
 Please check whether you extended any of our base renderers in your adaptation.
 
+### Scopable interface change
+
+The `scope` attribute in `Scopable` is now optional.
+Use `Scoped` instead for non optional scopes.
+The utility function `fromScopable` was renamed to `fromScoped` accordingly.
```
sdirix, line 111:
```suggestion
The utility function `fromScopable` was renamed to `fromScoped` accordingly.

```

PR #1937, opened 2022-05-19
Integrate external errors in form

### LINKED ISSUE #1926: Integrate external errors in form

There are scenarios in which validations are executed in the backend but still need to be shown within a form. Currently this can only be reasonably supported using custom controls to consume the errors provided by the backend. It would be cleaner overall to reuse the default error handling of JSON Forms for this.

JSON Forms should offer a way to consume external errors and mix them in with its usual error handling. We can assume that the errors arrive in the same format as the ones we receive from AJV.

Note that we need to think about the interaction of the external errors with the already supported validation modes.

Support this functionality with the more generic "middleware" support, however this is definitely more effort than simply supporting this use case.

### Framework

Core, React, Angular, Vue 3, Vue 2

### RendererSet

### Additional context

_No response_

AlexandraBuzila:
Add support for specifying external errors via `additionalErrors:
ajv.ErrorObject[]` prop.

The additional errors are merged with the validation errors and are
supplied to the affected controls. The additional errors are not
affected by the validationMode like the validation errors are and it is
the responsability of the framework user to update the prop when
applicable (e.g. on ValidationMode changes).

Fixes #1926

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -65,6 +65,7 @@ export interface JsonFormsCore {
   schema: JsonSchema;
   uischema: UISchemaElement;
   errors?: ErrorObject[];
+  additionalErrors?: ErrorObject[];
   validator?: ValidateFunction;
   ajv?: Ajv;
   validationMode?: ValidationMode;
@@ -78,6 +79,7 @@ const initState: JsonFormsCore = {
   validator: undefined,
   ajv: undefined,
   validationMode: 'ValidateAndShow',
+  additionalErrors: []
 };
 
 const reuseAjvForSchema = (ajv: Ajv, schema: JsonSchema): Ajv => {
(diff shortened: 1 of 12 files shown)


packages/core/src/reducers/core.ts
```diff
@@ -391,8 +415,11 @@ const getErrorsAt = (
   instancePath: string,
   schema: JsonSchema,
   matchPath: (path: string) => boolean
-) => (state: JsonFormsCore): ErrorObject[] =>
-  errorsAt(instancePath, schema, matchPath)(state.validationMode === 'ValidateAndHide' ? [] : state.errors);
+) => (state: JsonFormsCore): ErrorObject[] => {
+  const errors = state.errors ?? [];
+  const additionalErrors = state.additionalErrors ?? [];
+  return errorsAt(instancePath, schema, matchPath)(state.validationMode === 'ValidateAndHide' ? [...additionalErrors] : [...errors, ...additionalErrors]);
```
sdirix, line 421:
```suggestion
  return errorsAt(instancePath, schema, matchPath)(state.validationMode === 'ValidateAndHide' ? additionalErrors : [...errors, ...additionalErrors]);
```

sdirix (changes requested):
Looks good already!

Can you add an example to the example app which adds/removes multiple additional errors on the click of a button?

AlexandraBuzila:
Thank you for the review, Stefan! I updated the code and added examples for React (material and vanilla) and vue-vanilla. The contributed example should work with the angular examples as well when they'll support displaying buttons for the contributed actions like the React app (it's currently not the case).

sdirix (approved):
Great! Thank you!

PR #1939, opened 2022-05-20
add errors parameter to translation function

LukasBoll:
Signed-off-by: Lukas Boll lukas-bool@web.de

### DIFF
--- a/packages/core/src/i18n/i18nUtil.ts
+++ b/packages/core/src/i18n/i18nUtil.ts
@@ -54,26 +54,26 @@ export const defaultErrorTranslator: ErrorTranslator = (error, t, uischema) => {
     getControlPath(error),
     `error.${error.keyword}`
   );
-  const specializedKeywordMessage = t(i18nKey, undefined);
+  const specializedKeywordMessage = t(i18nKey, undefined, { error } );
   if (specializedKeywordMessage !== undefined) {
     return specializedKeywordMessage;
   }
 
   // check whether there is a generic keyword message
-  const genericKeywordMessage = t(`error.${error.keyword}`, undefined);
+  const genericKeywordMessage = t(`error.${error.keyword}`, undefined, { error });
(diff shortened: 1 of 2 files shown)


packages/core/src/i18n/i18nUtil.ts
```diff
@@ -54,26 +54,26 @@ export const defaultErrorTranslator: ErrorTranslator = (error, t, uischema) => {
     getControlPath(error),
     `error.${error.keyword}`
   );
-  const specializedKeywordMessage = t(i18nKey, undefined);
+  const specializedKeywordMessage = t(i18nKey, undefined, [error]);
```
sdirix, line 57:
```suggestion
  const specializedKeywordMessage = t(i18nKey, undefined, { error } );
```

packages/core/src/util/renderer.ts
```diff
@@ -465,8 +465,8 @@ export const mapStateToControlProps = (
   const schema = resolvedSchema ?? rootSchema;
   const t = getTranslator()(state);
   const te = getErrorTranslator()(state);
-  const i18nLabel = t(getI18nKey(schema, uischema, path, 'label'), label);
-  const i18nDescription = t(getI18nKey(schema, uischema, path, 'description'), description);
+  const i18nLabel = t(getI18nKey(schema, uischema, path, 'label'), label, errors.length > 0 ? errors : undefined);
```
sdirix, line 468: We can hand over even more context for label, description etc.
```suggestion
  const i18nLabel = t(getI18nKey(schema, uischema, path, 'label'), label, {schema, uischema, path, errors} );
```

sdirix (changes requested):
I would prefer if we always hand over an object for the 3rd parameter. Then we can easily add additional parameters later on and users don't need to type check.

sdirix (approved):
LGTM

PR #1944, opened 2022-05-24
Move i18n key specification from UI Schema options to ControlElement

### LINKED ISSUE #1942: Polish translation feature

The `i18n` keys are currently definable within UI Schema options. This doesn't make much sense as merging in a default `config` with an `i18n` key is rarely a desired use case.

The `i18n` key should be placed at the root level of the respective UI Schema element.

None

### Framework

Core

### RendererSet

### Additional context

_No response_

lucas-koehler:
* The i18n base key can no longer be specified as part of the UI Schema `options`
* Add optional `i18n` property to `ControlElement`
* Adapt i18n key resolution
* Add explicit tests for i18n key extraction from UI Schema and Json Schema
* Add `Internationalizable` interface and `isInternationalized` type guard to clearly define internationalizable UISchema elements.

Fix #1942

### DIFF
--- a/packages/core/src/i18n/i18nUtil.ts
+++ b/packages/core/src/i18n/i18nUtil.ts
@@ -1,5 +1,5 @@
 import { ErrorObject } from 'ajv';
-import { UISchemaElement } from '../models';
+import { isControlElement, UISchemaElement } from '../models';
 import { getControlPath } from '../reducers';
 import { formatErrorMessage } from '../util';
 import { i18nJsonSchema, ErrorTranslator, Translator } from './i18nTypes';
@@ -8,7 +8,10 @@ export const getI18nKeyPrefixBySchema = (
   schema: i18nJsonSchema | undefined,
   uischema: UISchemaElement | undefined
 ): string | undefined => {
-  return uischema?.options?.i18n ?? schema?.i18n ?? undefined;
+  if (uischema && isControlElement(uischema) && uischema.i18n) {
+    return uischema.i18n;
+  }
+  return schema?.i18n ?? undefined;
 };
 
 /**
--- a/packages/core/src/i18n/i18nUtil.ts
+++ b/packages/core/src/i18n/i18nUtil.ts
@@ -1,5 +1,5 @@
 import { ErrorObject } from 'ajv';
-import { isControlElement, UISchemaElement } from '../models';
+import { isInternationalized, UISchemaElement } from '../models';
 import { getControlPath } from '../reducers';
 import { formatErrorMessage } from '../util';
 import { i18nJsonSchema, ErrorTranslator, Translator } from './i18nTypes';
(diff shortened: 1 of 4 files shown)


packages/core/src/i18n/i18nUtil.ts
```diff
@@ -8,7 +8,10 @@ export const getI18nKeyPrefixBySchema = (
   schema: i18nJsonSchema | undefined,
   uischema: UISchemaElement | undefined
 ): string | undefined => {
-  return uischema?.options?.i18n ?? schema?.i18n ?? undefined;
+  if (uischema && isControlElement(uischema) && uischema.i18n) {
```
sdirix, line 11: At the moment we only support it for `ControlElement`s but for this implementation here I would generalize this a bit. For example via an interface `Internationalizable` which is extended by `ControlElement`.

The function would then work unchanged for all UI Schema elements of the future instead of only for `Control`s.

packages/core/src/models/uischema.ts
```diff
@@ -213,6 +213,11 @@ export interface ControlElement extends UISchemaElement, Scopable {
    * An optional label that will be associated with the control
    */
   label?: string | boolean | LabelDescription;
+  /**
+   * The i18n key for the control. It is used to identify the string that needs to be translated.
+   * It is suffixed with `.label`, `.description` and `.error.<keyword>` to derive the corresponding message keys for the control.
```
sdirix, line 218: There can be many more suffixes depending on each control. I would only give them as examples.

packages/core/src/models/uischema.ts
```diff
@@ -244,6 +249,9 @@ export interface Categorization extends UISchemaElement {
   elements: (Category | Categorization)[];
 }
 
+export const isControlElement = (element: UISchemaElement): element is ControlElement =>
```
sdirix, line 252: We already have a `isControl` type guard.

sdirix (changes requested):

lucas-koehler:
@sdirix Thanks for the review. I rebased the changes onto master and added a second commit that provides a new `Internationalizable` interface and a typeguard checking whether its `i18n` property is set.

packages/core/src/models/uischema.ts
```diff
@@ -244,6 +253,10 @@ export interface Categorization extends UISchemaElement {
   elements: (Category | Categorization)[];
 }
 
+export const isInternationalizable = (element: unknown): element is Required<Internationalizable> => {
```
sdirix, line 256: I would change the name a bit to better fit the interface name, i.e. `..able` -> optional prop and `...ed` -> required prop.

```suggestion
export const isInternationalized = (element: unknown): element is Required<Internationalizable> => {
```

sdirix (changes requested):

lucas-koehler:
@sdirix Thanks for the suggestion, that indeed fits better :) I adapted the commit to name the new type guard `isInternationalized`.

sdirix (approved):

PR #1949, opened 2022-06-02
Pass through original values for invalid dates

sunnysingh:
This is a follow-up to the following community discussion: [Showing errors for badly formatted date and time fields](https://jsonforms.discourse.group/t/showing-errors-for-badly-formatted-date-and-time-fields/790).

The current behavior attempts to format date input values with the dayjs library, and if the formatted string returns `Invalid Date` then it will call `handleChange` with an `undefined` value. This not only causes dates to not show up in the final `data` object, but it also makes it impossible to show validation errors for optional fields.

This PR introduces a bugfix to pass through the original `keyboardInputValue`, which triggers validation:

Screen Shot 2022-06-02 at 11 43 03 AM

This is my first contribution so please let me know if I'm missing anything.


sdirix:
Thank you very much. We're gonna take a look very soon :heart:

packages/material/src/util/datejs.ts
```diff
@@ -8,13 +8,13 @@ export const createOnChangeHandler = (
   path: string,
   handleChange: (path: string, value: any) => void,
   saveFormat: string | undefined
-) => (time: dayjs.Dayjs) => {
+) => (time: dayjs.Dayjs, keyboardInputValue: string) => {
```
sdirix, line 11: looks good, I would just rename the parameter
```suggestion
) => (time: dayjs.Dayjs, textInputValue: string) => {
```

sunnysingh:
> The PR improves the situation but there are still problems. It seems that `dayjs` is very very lenient in regards of format and only reports `Invalid Date` when it can't match at all.
> 
> For example `aaabbb1sad3asdasd00` in the time picker is parsed to `01:03:00` which might be unexpected for the user.
> 
> I would like to suggest to add an `onBlur` handler to the date, time and date-time renderers which makes sure that once the user leaves the input, the text-input is reset to reflect the actual data.
> 
> Do you agree? Can you add this?

Thank you for reviewing. I noticed this issue too and wasn't sure how to tackle it, but I like the "set value to parsed data on blur" idea, will add this in.

packages/material/src/util/datejs.ts
sunnysingh: This name comes from MUI, but they're not always great at names and `textInputValue` makes more sense. Will update this.

sunnysingh:
@sdirix I got a chance to try and implement the "sync input values with parsed data on blur" improvement. However, it doesn't seem quite right especially with the failing unit tests (some of the errors are expected, but others seem to provide very different actual values from the expected values). When you get a chance, do you mind taking a look at this?

packages/material/src/util/datejs.ts
```diff
@@ -30,3 +31,38 @@ export const getData = (
...
+}) => {
+  const [value, setValue] = useState(props.data);
+
+  const onBlur = useMemo(
+    () => (event: DateInputFormEvent) => {
+      setValue(props.data);
+      if (props.onBlur) props.onBlur(event);
+    },
+    [props.data, props.onBlur]
+  );
```
sdirix, line 58: We can use `onBlur` directly however it's already used for `focused`. So instead of injecting ourselves into `onBlur` we can just check the value of `focused`.

LukasBoll:
Hi, 
I also reviewed the pull request. Code looks good to me and everything behaves like expected! :+1:

sdirix (approved):
LGTM

sdirix:
@sunnysingh I merged this now in. Please let me know if there is something which could be improved from your point of view.

PR #1951, opened 2022-06-03
Pass all props in the MaterialArrayLayoutRenderer

eneufeld:
Currently the MaterialArrayLayoutRenderer lists the props explicitly,
this way the config was forgotten and thus the
move up/down buttons cannot be enabled globally using the config object.

- add test case for config object being passed down
- pass all props down to the MaterialArrayLayout

### DIFF
--- a/packages/material/src/layouts/MaterialArrayLayoutRenderer.tsx
+++ b/packages/material/src/layouts/MaterialArrayLayoutRenderer.tsx
@@ -36,19 +36,8 @@ import { withJsonFormsArrayLayoutProps } from '@jsonforms/react';
 
 export const MaterialArrayLayoutRenderer = ({
   visible,
-  enabled,
-  id,
-  uischema,
-  schema,
-  label,
-  rootSchema,
-  renderers,
-  cells,
-  data,
-  path,
-  errors,
-  uischemas,
-  addItem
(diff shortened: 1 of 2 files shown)


packages/material/src/layouts/MaterialArrayLayoutRenderer.tsx
```diff
@@ -36,40 +36,18 @@ import { withJsonFormsArrayLayoutProps } from '@jsonforms/react';
...
-        errors={errors}
-        enabled={enabled}
         visible={visible}
-        data={data}
-        path={path}
         addItem={addItemCb}
-        renderers={renderers}
-        cells={cells}
-        uischemas={uischemas}
+        {...props}
```
sdirix, line 50: The props need to be applied *before* we set our manual props, as otherwise they will override `addItem` again.

sdirix (changes requested):

packages/material/src/layouts/MaterialArrayLayoutRenderer.tsx
sdirix: Actually that's wrong because we already extracted `addItem`

sdirix (approved):

PR #1960, opened 2022-06-21
Add default views for material date/time controls

lucas-koehler:
When no views to show for date and/or time controls are specified in the UI Schema, explicitly hand in the default ones.

This ensures consitent behavior if the internal default of the material controls changes and avoids type errors on undefined views arrays.

### DIFF
--- a/packages/material/src/controls/MaterialDateControl.tsx
+++ b/packages/material/src/controls/MaterialDateControl.tsx
@@ -68,6 +68,8 @@ export const MaterialDateControl = (props: ControlProps)=> {
   const format = appliedUiSchemaOptions.dateFormat ?? 'YYYY-MM-DD';
   const saveFormat = appliedUiSchemaOptions.dateSaveFormat ?? 'YYYY-MM-DD';
 
+  const views = appliedUiSchemaOptions.views ?? ['year', 'day'];
+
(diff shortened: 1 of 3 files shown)


packages/material/src/controls/MaterialDateControl.tsx
```diff
@@ -68,6 +68,8 @@ export const MaterialDateControl = (props: ControlProps)=> {
   const format = appliedUiSchemaOptions.dateFormat ?? 'YYYY-MM-DD';
   const saveFormat = appliedUiSchemaOptions.dateSaveFormat ?? 'YYYY-MM-DD';
 
+  const views = appliedUiSchemaOptions.views ?? ['year', 'day'];
```
sdirix, line 71: we should probably add month here too, right ;)

lucas-koehler: Actually no. This is the default in @mui/x-date-pickers, too. The initial view is the same for both and looks like this:
image

The only difference is that after clicking on the year and the selecting a year, right now it jumps to the screen shown above (with the changed year ofc). Adding the `'month'` view adds a month selection after selecting the year. IMO this is mostly not needed and just forces the user to select the month again

PR #1961, opened 2022-06-21
React Vanilla enum renderer removes empty data

### LINKED ISSUE #1959: Vanilla enum renderer should store undefined instead of empty string
### Describe the bug

When selecting the empty option in React Vanilla's enum renderer, an empty string is stored instead of `undefined`. This is unexpected and also leads to validation errors.

### Expected behavior

Undefined should be stored instead of the empty string when an enum is "unselected".

1. Run the `person` example
2. Select a `nationality`
3. Unselect `nationality`

### Screenshots

_No response_

102.0.5005.115

### Framework

React

### RendererSet

Vanilla

### Additional context

_No response_

sdirix:
Instead of the empty string the React Vanilla enum renderer will now hand over
'undefined' to handleChange and therefore delete the corresponding attribute.

Fixes #1959


LukasBoll:
Looks good to me!

Issue #1884, opened 2022-02-06
mapStateToJsonFormsRendererProps should map enabled property just like react own version of the method


kchobantonov:
### Describe the bug

Currently when using different render set than react the enabled property is not preserved.

For example if we use the vuetify render set in this example (https://jsonforms-vuetify-renderers.netlify.app/#/example/main)  - the "Address for Shipping T-Shirt" section is never disabled per committer checkbox.

Here is the relevant part of the react renderer set that does not use the core version of mapStateToJsonFormsRendererProps 
https://github.com/eclipsesource/jsonforms/blob/7532a3e16558d9f0934bd5c602ccbe3a35ed9dab/packages/react/src/JsonForms.tsx#L141-L153

Here is the relevant discussion on discourse https://jsonforms.discourse.group/t/jsonforms-should-propagate-enabled-property-when-mapping-render-properties/489

### Expected behavior

Expected behavior is that the enabled property is preserved in its own properties so that it can be passed to lower components and be able to disable/enable them

### Steps to reproduce the issue

1. Go to 'https://jsonforms-vuetify-renderers.netlify.app/#/example/main'
2. Click on 'Committer' checkbox - make sure that it is not checked
3. See  "Address for Shipping T-Shirt" never disabled

### In which browser are you experiencing the issue?

any

### Framework

Core, Vue 2

### RendererSet

Other (please specify in the Additional context field)

### Additional context

https://github.com/eclipsesource/jsonforms-vuetify-renderers

Potentially we could need the visibility to be mapped as well perhaps

sdirix:
Hi @kchobantonov, as mentioned in the discussion. I agree, this should definitely be improved upon.

sdirix:
Fixed with #1968

PR #1968, opened 2022-07-01
Issue 1884

kchobantonov:
This is fixing the issue https://github.com/eclipsesource/jsonforms/issues/1884

To check that run vue-vanilla renderer set  (using cd packages/vue/vue-vanilla && npm run serve)
and then select the newly added example - Issue 1884 and check that when the committer checkbox is selected only then the address text fields are enabled.


kchobantonov:
@sdirix can you please review.

sdirix:
Thanks for the PR. We'll take a look soon!

sdirix (commented):
Coincidentally we performed the same change in [another PR](https://github.com/eclipsesource/jsonforms/pull/1975) ;) 

Still the example app improvement is very appreciated!

PR #1974, opened 2022-07-07
Extend `ExampleDescription` with optional i18n configuration

lucas-koehler:
* Extend the ExampleDescription to allow configuring i18n for examples
* Adapt example apps to use a provided i18n config
* Add i18n configuration to the `categorization` and `Person (i18n)` examples


sdirix (approved):
LGTM

Issue #1970, opened 2022-07-05
Hand over more context to the testers


sdirix:
### Is your feature request related to a problem? Please describe.

Testers currently only receive `schema`, `uischema` and the root `schema` which was recently introduced with the resolving refactoring efforts.

### Describe the solution you'd like

Instead of only the root schema we should hand over an object which can contain an arbitrary number of variables. One of them should be the `rootSchema` but the global `config` object would also be useful. The testers could then not only check the local ui schema options but also the global ones aligning better to the overall expectation of how such a feature should work.

Note that we should not hand over the whole JSON Forms state as we can then no longer cache the tester evaluations.

So for this ticket we should:
 - Introduce the context parameter in the testers
 - Hand over `rootSchema` and `config` within that context
 - Adapt current use of `rootSchema` to look into the context object
 - Adapt current UI Schema tester helpers which check the UI Schema option to also check the context `config` object
 - Adapt test cases
 - Adapt migration document

### Describe alternatives you've considered

Instead of a single context object we could add more and more parameters. However a single object seems to be more clean going forward.

### Framework

Core

### RendererSet

Other (please specify in the Additional context field)

### Additional context

Should be done before 3.0 as this is a breaking change and an adaption of another 3.0 change.

PR #1975, opened 2022-07-08
Provide additional tester context

lucas-koehler:
### core:
* Change the third parameter of Tester and RankedTester from just the root schema to a new TesterContext object. It contains the root schema and the global config.
* Remove UI Schema generation from mapStateToJsonFormsRendererProps. Instead, it is taken from the state.

### react:
* Rename ctxToJsonFormsDispatchProps to ctxToJsonFormsRendererProps
* Add HOC withJsonFormsRendererProps that injects state props

### vue:
* Upgrade @vue/test-utils to the latest stable version
* Add config as an optional renderer prop

Fixes https://github.com/eclipsesource/jsonforms/issues/1970

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -16,15 +16,21 @@ Therefore JSON Forms was not able to properly run the testers on schemas contain
 The workaround for this was to resolve the JSON Schema by hand before handing it over to JSON Forms.
 Only the React renderers did this automatically but we removed this functionality, see the next section for more information.
 
-We now added an additional parameter to the testers, the `rootSchema`.
+We now added an additional parameter to the testers, the new `TesterContext`.
 
 ```ts
-type Tester = (uischema: UISchemaElement, schema: JsonSchema, rootSchema: JsonSchema) => boolean;
-type RankedTester = (uischema: UISchemaElement, schema: JsonSchema, rootSchema: JsonSchema) => number;
+interface TesterContext {
+  rootSchema: JsonSchema;
+  config: any;
+}
+
+type Tester = (uischema: UISchemaElement, schema: JsonSchema, context: TesterContext) => boolean;
+type RankedTester = (uischema: UISchemaElement, schema: JsonSchema, context: TesterContext) => number;
 ```
 
-This allows the testers to resolve any `$ref` they might encounter in their handed over `schema`.
+This allows the testers to resolve any `$ref` they might encounter in their handed over `schema` by using the context's `rootSchema`.
 Therefore the manual resolving of JSON Schemas before handing them over to JSON Forms does not need to be performed in those cases.
+In addition, testers can now access the global `config` to consider default UI Schema options.
 
 ### Removal of JSON Schema $Ref Parser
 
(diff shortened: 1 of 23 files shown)


packages/react/src/DispatchCell.tsx
```diff
@@ -31,9 +31,19 @@ import { withJsonFormsDispatchCellProps } from './JsonFormsContext';
...
+      rootSchema: rootSchema,
+      config: config
+    }),
+    [rootSchema, config]
+  );
+  const cell = useMemo(
+    () => maxBy(cells, r => r.tester(uischema, schema, testerContext)),
+    [cells, uischema, schema, testerContext]
+  );
+  if (cell === undefined || cell.tester(uischema, schema, testerContext) === -1) {
```
sdirix, line 46: Personally I would have done one `useMemo` as the standalone `testerContext` `useMemo` is not really necessary but that's just a minor issue.

packages/react/src/JsonForms.tsx
```diff
@@ -102,14 +103,22 @@ const TestAndRender = React.memo(
...
-      () => maxBy(props.renderers, r => r.tester(props.uischema, props.schema, props.rootSchema)),
-      [props.renderers, props.uischema, props.schema]
+      () => maxBy(props.renderers, r => r.tester(props.uischema, props.schema, testerContext)),
+      [props.renderers, props.uischema, props.schema, testerContext]
     );
     if (
       renderer === undefined ||
-      renderer.tester(props.uischema, props.schema, props.rootSchema) === -1
+      renderer.tester(props.uischema, props.schema, testerContext) === -1
     ) {
```
sdirix, line 122: same here

MIGRATION.md
```diff
@@ -16,15 +16,21 @@ Therefore JSON Forms was not able to properly run the testers on schemas contain
...
+}
+
+type Tester = (uischema: UISchemaElement, schema: JsonSchema, context: TesterContext) => boolean;
+type RankedTester = (uischema: UISchemaElement, schema: JsonSchema, context: TesterContext) => number;
 ```
 
-This allows the testers to resolve any `$ref` they might encounter in their handed over `schema`.
+This allows the testers to resolve any `$ref` they might encounter in their handed over `schema` by using the context's `rootSchema`.
 Therefore the manual resolving of JSON Schemas before handing them over to JSON Forms does not need to be performed in those cases.
+In addition, testers can now access the global `config` to consider default UI Schema options.
```
sdirix, line 33: There are also other use cases so I would just add that the `config` is now available. Especially as we don't use the config in this PR.

sdirix, line 33: Instead of `global` I usually use the term "form wide" to avoid confusion

packages/core/src/testers/testers.ts
```diff
@@ -49,17 +49,27 @@ export const NOT_APPLICABLE = -1;
...
+  context: TesterContext
 ) => number;
 
+/**
+ * Additional context given to a tester in addition to UISchema and JsonSchema.
+ */
+export interface TesterContext {
+  /** The root JsonSchema of the form. Can be used to resolve references. */
+  rootSchema: JsonSchema;
+  /** The global configuration object given to JsonForms. Can be used to derive default UISchema options. */
```
sdirix, line 69: same here

sdirix (changes requested):
Works great. Just very minor comments.

sdirix (approved):
LGTM

PR #1979, opened 2022-07-15
Mark compatible with Angular 14

TheZoker:
Mark JSONForms compatible with Angular 14. 

Was tested within the angular-seed and everything worked.


sdirix (changes requested):
Please also mark the Angular dependencies in other packages (especially `@jsonforms/angular-material`) as compatible to Angular 14

sdirix (approved):
LGTM

PR #1982, opened 2022-07-19
change MuiAutocomplete from Input to Textfield

### LINKED ISSUE #1965: MUI Autocomplete has no endAdornment arrow

The autocomplete renderer does not have a a downwards pointing arrow to indicate that clicking it will open a dropdown.

A downward arrow at the left of the autocomplete, as is the default behaviour for the MUI autocomplete.

Creating a custom autocomplete renderer (this seems like overkill)

### Framework

React

### RendererSet

Material

### Additional context

_No response_

LukasBoll:
closes #1965

### DIFF
--- a/packages/material/src/controls/MaterialEnumControl.tsx
+++ b/packages/material/src/controls/MaterialEnumControl.tsx
@@ -37,12 +37,18 @@ import { MaterialInputControl } from './MaterialInputControl';
 import { MuiAutocomplete, WithOptionLabel } from '../mui-controls/MuiAutocomplete';
 
 export const MaterialEnumControl = (props: ControlProps & OwnPropsOfEnum & WithOptionLabel) => {
-  const {config, uischema} = props;
+  const {config, uischema, errors} = props;
   const appliedUiSchemaOptions = merge({}, config, uischema.options);
+  const isValid = errors.length === 0;
   return (
+    appliedUiSchemaOptions.autocomplete === false ?
     <MaterialInputControl
       {...props}
-      input={appliedUiSchemaOptions.autocomplete === false ? MuiSelect : MuiAutocomplete}
+      input={MuiSelect}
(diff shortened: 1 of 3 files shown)


packages/material/src/controls/MaterialEnumControl.tsx
```diff
@@ -37,12 +37,19 @@ import { MaterialInputControl } from './MaterialInputControl';
...
+    appliedUiSchemaOptions.autocomplete === false ?
     <MaterialInputControl
       {...props}
-      input={appliedUiSchemaOptions.autocomplete === false ? MuiSelect : MuiAutocomplete}
+      input={MuiSelect}
+    />:
+    <MuiAutocomplete
+      {...props}
+      isValid={isValid}
+      visible={visible}
```
sdirix, line 52: * `visible` is already handed over via the `...props` so we don't need to do this explicitely.
 * There is no need to determine and hand over `isValid` here. Just do the calculation within the `MuiAutocomplete`.

packages/material/src/controls/MaterialOneOfEnumControl.tsx
```diff
@@ -37,12 +37,20 @@ import { MaterialInputControl } from '../controls/MaterialInputControl';
...
+    appliedUiSchemaOptions.autocomplete === false ?
     <MaterialInputControl
       {...props}
-      input={appliedUiSchemaOptions.autocomplete === false ? MuiSelect : MuiAutocomplete}
+      input={MuiSelect}
+    />:
+    <MuiAutocomplete
+      {...props}
+      isValid={isValid}
+      visible={visible}
```
sdirix, line 53: same here

sdirix (changes requested):

sdirix (approved):
LGTM!

PR #1984, opened 2022-07-21
Deleting Values

### LINKED ISSUE #1983: Empty inputs should result in attribute being deleted

We often save empty strings in the data which is unintuitive behavior and very likely not the effect most users want.

- React Material String
- React Material Select (autocomplete:off)
- React Vanilla String
- React Vanilla Select
- React Vanilla Number/Integer

Keeping status quo

### Framework

React

### RendererSet

Material, Vanilla

### Additional context

_No response_

LukasBoll:
add empty option for enum
empty input now deletes attributes.

closes #1983

### DIFF
--- a/packages/core/src/util/cell.ts
+++ b/packages/core/src/util/cell.ts
@@ -55,7 +55,7 @@ import {
 } from './renderer';
 import { JsonFormsState } from '../store';
 import { JsonSchema } from '../models';
-import { getI18nKey, getI18nKeyPrefix } from '../i18n';
+import { getI18nKeyPrefix } from '../i18n';
 
 export type { JsonFormsCellRendererRegistryEntry };
 
@@ -212,11 +212,9 @@ export const defaultMapStateToEnumCellProps = (
         getI18nKeyPrefix(props.schema, props.uischema, props.path)
       )
     ]);
-  const emptyLabel = getEmptyLabel(props,state);
   return {
     ...props,
-    options,
-    emptyLabel
+    options
   };
 };
 
@@ -240,11 +238,9 @@ export const mapStateToOneOfEnumCellProps = (
         getI18nKeyPrefix(props.schema, props.uischema, props.path)
       )
     );
-  const emptyLabel = getEmptyLabel(props,state);
   return {
     ...props,
-    options,
-    emptyLabel
+    options
   };
 };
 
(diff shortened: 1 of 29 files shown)


packages/core/src/util/cell.ts
```diff
@@ -267,3 +271,8 @@ export const defaultMapDispatchToControlProps =
       handleChange: ownProps.handleChange || handleChange
     };
   };
+
+function getEmptyLabel(props: OwnPropsOfEnumCell, state: JsonFormsState) {
+  const t = getTranslator()(state);
+  return t(getI18nKey(props.schema, props.uischema, props.path, 'empty'), 'empty');
```
sdirix, line 277: I would prefer `None` over `Empty`.

packages/core/src/util/renderer.ts
```diff
@@ -260,6 +260,7 @@ export interface OwnPropsOfLabel extends OwnPropsOfRenderer {
 
 export interface OwnPropsOfEnum {
   options?: EnumOption[];
+  emptyLabel?: string;
```
sdirix, line 263: Hmm I don't think we should handle this via the props. For me this seems like more a decision on the renderer side whether they want to offer a `None` option.

packages/material/src/mui-controls/MuiSelect.tsx
```diff
@@ -49,11 +50,11 @@ export const MuiSelect = React.memo((props: EnumCellProps & WithClassname) => {
       disabled={!enabled}
       autoFocus={appliedUiSchemaOptions.focus}
       value={data !== undefined ? data : ''}
-      onChange={ev => handleChange(path, ev.target.value)}
+      onChange={ev =>handleChange(path, ev.target.value === '' ? undefined : ev.target.value)}
```
sdirix, line 53: This will delete values which are empty on purpose. We really need to check whether this was our `None` entry.

sdirix (changes requested):

packages/material/src/mui-controls/MuiSelect.tsx
sdirix: On further inspection it seems it's not straightforward to get this information out of Material UI's `Select`.

sdirix (approved):

PR #2001, opened 2022-08-18
resolve Ref in resolveSchema

LukasBoll:


sdirix (changes requested):
I understand the concept you were going for but I would keep the changes much more narrow in scope. Instead of modifying the core resolving functionality I would like to suggest to just trigger it again for the currently failing case (similar to depth = 1).

This would mean just triggering the resolve again [in this line](https://github.com/eclipsesource/jsonforms/blob/d04e851f56537bcf6bda9240ab7e0ae5f09b4010/packages/core/src/util/renderer.ts#L970) in case the schema is just a ref. Then the other changes are not needed.

sdirix (approved):
Nice! That's how I wanted to see it! Great!

PR #2013, opened 2022-08-30
(core) Categorization interface does not have a 'label' property

jdwit:
I may be missing something but I think the `Categorization` interface should not extend `Labeled`; the top level element of a Category layout (`type: "Categorization"`) does not have a label, [the example JSON in the docs](https://jsonforms.io/docs/uischema/layouts#categorization) confirms this.


sdirix:
Hi @jdwit,

You are right! Thanks for the contribution :heart:

sdirix:
Closing this for now as indicated by the previous comment.

PR #2015, opened 2022-09-06
Update RXJS peer depenency version to include ^7.4.0

### LINKED ISSUE #2014: Update version of peer dependency RXJS

We're running into some issues trying to implement JSON forms. 

We have an Angular app running on version 13.3.0. That version of Angular has a peer dependency on RXJS, at version `^7.4.0`.

Unfortunately, this results in conflicts when trying to install `jsonforms`, because this package has a peer dependency on RXJS at version `6.4.0`. 

Would it be possible to update this dependency to `^6.4.0 || ^7.4.0`?

Using `npm install --force`, but that is not an ideal solution.

### Framework

Angular

### RendererSet

_No response_

### Additional context

_No response_

lucas-koehler:
* Include versions ^7.4.0 into the allowed versions. They are allowed by Angular from 12.2.0 upwards
* Increase the lower bound of the RXJS 6.x dependency to ^6.5.3 to match Angular's actual requirements
  * Even @angular/core 12.0.0 already requires ^6.5.3
* Updated package-lock.json generated in dev container

I tested the latest version of rxjs (7.5.6) with rc 1 in the angular seed and did not notice any differences to using the 6.x version

Fixes #2014

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5607,27 +5607,6 @@
           "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz",
           "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw=="
         },
-        "ajv": {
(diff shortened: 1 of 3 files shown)


sdirix (approved):
LGTM

PR #2016, opened 2022-09-08
Improve angular performance

### LINKED ISSUE #1946: Improve Angular performance
### Describe the bug

I analyze the code how is implemented, there are a lot of function iteration calls, the whole schema its updated even in click on select option control, on focus, if you inspect in chrome, you will see the performance issues, I don't know why all the logic of updates are not set up in services.
In comparison of React - in React it works ok.

### Expected behavior

/

/

### Screenshots

_No response_

Google Chrome 100

### Framework

Angular

### RendererSet

Material

### Additional context

_No response_

lucas-koehler:
### Calculate Angular layout renderer child props via a pure pipe
Increase rendering and on change performance for all forms including nultiple layouts:
* Avoid unnecessary recalculation of the child render props.
* In turn, avoid some unnecessary change detection cycles in children.

### Add service method to get global config

When querying the state from the `JsonFormsAngularService` via its `getState` method, the whole state is cloned.
This lead to a heavy performance penalty because the `JsonFormsAbstractControl` queried the state on every change detection cycle to get the global config. Thus, the whole state was cloned for every control on every change.
As the config is only a small part of the state, this problem can be alleviated by adding a getConfig method to the service that only clones the config.

* Extend `JsonFormsAngularService` with a `getConfig` method that returns a clone of the global config
* Use the new method in `JsonFormsAbstractControl`

### Add service method to get locale

When querying the state from the `JsonFormsAngularService` via its `getState` method, the whole state is cloned.
This could lead to performance issues with the angular material renderer set because the `NumberControlRenderer` queried the state on every change detection cycle to get the locale. Thus, the whole state was cloned for every control of these types on every change.
To fix this, the `JsonFormsAngularService` now allows to query the locale directly without cloning the state

* Add method `getLocale` to `JsonFormsAngularService` that does not clone the state
* Use new method in `NumberControlRenderer`

Fix #1946

### DIFF
--- a/packages/angular-material/src/controls/number.renderer.ts
+++ b/packages/angular-material/src/controls/number.renderer.ts
@@ -25,7 +25,6 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
 import { JsonFormsAngularService, JsonFormsControl } from '@jsonforms/angular';
 import {
-  getLocale,
   isIntegerControl,
   isNumberControl,
   or,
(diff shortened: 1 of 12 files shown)


FredUK:
Please can we get this reviewed by an approver? The angular component at the moment is very slow and laggy when we have more than a dozen components.

sdirix:
Hi @FredUK, yes will review soonish and if there are no issues this will be part of the next (pre)release.

sdirix (approved):
LGTM! Did not find any regression

PR #2020, opened 2022-09-21
Add Interface for config of Jsonforms-Angular

lambnoah99:
Add Interface for Jsonforms-Config, instead of 'any' Datatype. This applies only for the Angular-Version.

### DIFF
--- a/packages/angular/src/config.ts
+++ b/packages/angular/src/config.ts
@@ -1,0 +1,7 @@
+export interface Config {
+    hideRequiredAsterisk?: boolean,
+    readonly?: boolean,
+    restrict?: boolean,
+    trim?: boolean;
+    showUnfocusedDescription?: boolean
+}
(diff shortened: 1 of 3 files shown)


sdirix (commented):
The config can actually be `any`. It's the renderer set which determines how the `config` shall be used. Even if we wanted to restrict this to the use case of default UI Schema options then this is too limited. There can be any number of custom renderers which look up additional properties in the `config`.

lambnoah99:
> The config can actually be `any`. It's the renderer set which determines how the `config` shall be used. Even if we wanted to restrict this to the use case of default UI Schema options then this is too limited. There can be any number of custom renderers which look up additional properties in the `config`.

Maybe an extensible interface (I'd say readonly and hideRequiredAsterisk  could/should be supported every renderer)?

PR #2021, opened 2022-09-21
Added add and remove icons to table renderer fixes #1782

### LINKED ISSUE #1782: Add "Add" and "Remove" actions to the Angular Table

At the moment the Angular Table is very basic. It would be great to add "Add" and "Remove" actions, similar to the Array Layout renderer.

Add an "Add" button which is able to
 - create a new item instance (should also support primitives)
 - create the array if it not yet exists

Add a "Remove" button in each row which is able to
- remove the corresponding entry

There could also be a remove button for the whole array which not only empties the array but removes it completely. Note that we don't have such a remove button yet in any other renderer set too.

### Framework

Angular

### RendererSet

Material

### Additional context

_No response_

Tarjei400:
Added add and remove icons for Table renderer

### DIFF
--- a/packages/angular-material/src/other/table.renderer.ts
+++ b/packages/angular-material/src/other/table.renderer.ts
@@ -30,12 +30,12 @@ import {
 } from '@jsonforms/angular';
 import {
   ArrayControlProps,
-  ControlElement,
+  ControlElement, createDefaultValue,
   deriveTypes,
   encode,
   isObjectArrayControl,
   isPrimitiveArrayControl,
-  JsonSchema,
+  JsonSchema, mapDispatchToArrayControlProps,
   or,
   OwnPropsOfRenderer,
   Paths,
@@ -48,12 +48,33 @@ import {
 @Component({
   selector: 'TableRenderer',
   template: `
+
     <table
       mat-table
       [dataSource]="data"
       class="mat-elevation-z8"
       [trackBy]="trackElement"
     >
+      <ng-container matColumnDef="action">
+        <tr>
+          <th mat-header-cell *matHeaderCellDef>
+            <button mat-icon-button color="primary" (click)="add()" [disabled]="!isEnabled()" title="Add"><mat-icon>add</mat-icon></button>
+          </th>
+        </tr>
+        <tr>
+          <td mat-cell *matCellDef="let row; let i = index" style="width:20%">
+            <button
+                mat-button
+                color="warn"
+                (click)="remove(i)"
+                [disabled]="!isEnabled()"
+                matTooltipPosition="right"
+            >
+              <mat-icon>delete</mat-icon>
+            </button>
+          </td>
+        <tr>
+      </ng-container>
       <ng-container
         *ngFor="let item of items"
         matColumnDef="{{ item.property }}"
@@ -77,6 +98,8 @@ export class TableRenderer extends JsonFormsArrayControl {
   displayedColumns: string[];
   items: ColumnDescription[];
   readonly columnsToIgnore = ['array', 'object'];
+  addItem: (path: string, value: any) => () => void;
+  removeItems: (path: string, toDelete: number[]) => () => void;
 
   constructor(jsonformsService: JsonFormsAngularService) {
     super(jsonformsService);
@@ -87,6 +110,9 @@ export class TableRenderer extends JsonFormsArrayControl {
   mapAdditionalProps(props: ArrayControlProps) {
     this.items = this.generateCells(props.schema, props.path);
     this.displayedColumns = this.items.map(item => item.property);
+    if (this.isEnabled()) {
+      this.displayedColumns.push('action');
+    }
   }
   getProps(index: number, props: OwnPropsOfRenderer): OwnPropsOfRenderer {
     const rowPath = Paths.compose(props.path, `${index}`);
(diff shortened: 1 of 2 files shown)


sdirix:
Thanks for the contribution :heart:. We will take a look right after the 3.0 release.

Tarjei400:
@sdirix let me know if I can be of any assistance

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -48,12 +48,33 @@ import {
 @Component({
   selector: 'TableRenderer',
   template: `
+
```
lucas-koehler, line 51: Please remove this superfluous empty line

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -48,12 +48,33 @@ import {
...
     <table
       mat-table
       [dataSource]="data"
       class="mat-elevation-z8"
       [trackBy]="trackElement"
     >
+      <ng-container matColumnDef="action">
+        <tr>
+          <th mat-header-cell *matHeaderCellDef>
+            <button mat-icon-button color="primary" (click)="add()" [disabled]="!isEnabled()" title="Add"><mat-icon>add</mat-icon></button>
```
lucas-koehler, line 61: Use the material tooltip instead of the plain `title`. Use `mat-button` like for the delete buttons to have the same spacing
```suggestion
            <button mat-button color="primary" (click)="add()" [disabled]="!isEnabled()" matTooltip="Add"><mat-icon>add</mat-icon></button>
```

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -48,12 +48,33 @@ import {
...
+          </th>
+        </tr>
+        <tr>
+          <td mat-cell *matCellDef="let row; let i = index" style="width:20%">
+            <button
+                mat-button
+                color="warn"
+                (click)="remove(i)"
+                [disabled]="!isEnabled()"
+                matTooltipPosition="right"
```
lucas-koehler, line 71: Add tooltip to delete buttons
```suggestion
                matTooltipPosition="right"
                matTooltip="Delete"
```

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -48,12 +48,33 @@ import {
...
       [trackBy]="trackElement"
     >
+      <ng-container matColumnDef="action">
+        <tr>
+          <th mat-header-cell *matHeaderCellDef>
+            <button mat-icon-button color="primary" (click)="add()" [disabled]="!isEnabled()" title="Add"><mat-icon>add</mat-icon></button>
+          </th>
+        </tr>
+        <tr>
+          <td mat-cell *matCellDef="let row; let i = index" style="width:20%">
```
lucas-koehler, line 65: I think we don't need the explicit width declaration here. Is there a reason that you added this?
```suggestion
          <td mat-cell *matCellDef="let row; let i = index">
```

lucas-koehler (changes requested):
@Tarjei400 Thanks again for the contribution :) The add and remove buttons already work great for me. I added some suggestions regarding the buttons itself. Could you take a look at them please?

packages/angular-material/src/other/table.renderer.ts
Tarjei400: Done

packages/angular-material/src/other/table.renderer.ts
Tarjei400: Done

packages/angular-material/src/other/table.renderer.ts
Tarjei400: @lucas-koehler is it okay to add something like this? 
```
  styles: [
      'table {width: 100%;}',
      '.cdk-column-action { width: 5%}']
      ```

I think material is putting some auto width on columns, I wanted to shrink it. so cell is taking only small fraction of a row.

File: packages/angular-material/src/other/table.renderer.ts [hunk 5]
```diff
@@ -87,6 +113,10 @@ export class TableRenderer extends JsonFormsArrayControl {
   mapAdditionalProps(props: ArrayControlProps) {
     this.items = this.generateCells(props.schema, props.path);
     this.displayedColumns = this.items.map(item => item.property);
+    console.log(this.displayedColumns, this.items );
```
lucas-koehler, line 116: Please remove this log again. I guess it was leftover from testing ;)

packages/angular-material/src/other/table.renderer.ts
```diff
@@ -54,6 +54,28 @@ import {
...
+        <tr>
+          <td mat-cell *matCellDef="let row; let i = index">
+            <button
+                mat-button
+                color="warn"
+                (click)="remove(i)"
+                [disabled]="!isEnabled()"
+                matTooltipPosition="right"
+                matTooltip="Delete"
+
```
lucas-koehler, line 72: Please remove this empty line :)

packages/angular-material/src/other/table.renderer.ts
lucas-koehler: Seems fine to me. I tested it in the example app and it seems to work well :)

lucas-koehler (commented):
Thanks for the updates. I just have two minor comments left.

packages/angular-material/src/other/table.renderer.ts
Tarjei400: Removed

packages/angular-material/src/other/table.renderer.ts
Tarjei400: Done

lucas-koehler (approved):
Thanks again for the updates. LGTM now :)

PR #2027, opened 2022-09-28
Fix category hiding in Angular Material

lucas-koehler:
Previously, the Angular Material category renderer didn't consider categories' visbilities and always showed themy.
This prefilters visible categories before rendering them in the template.

Fix unit test by moving the setting of the new UI schema in the jsonFormsService after it was reassigned in the component. With this, when the event is triggered on the subscription, the new schema is already present. This is necessary for the new changes and also fixes the previously commented out assertion.


sdirix (approved):
Works for me

PR #2035, opened 2022-10-15
fix angular material date control selecting the wrong date

TheMicDiet:
replace toISOString(), which formats dates based on the UTC time zone with Angular DatePipe, which works with local time zones

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -205,14 +205,6 @@
         "tslib": "^2.2.0"
       }
     },
-    "@angular/material-moment-adapter": {
-      "version": "12.2.13",
-      "resolved": "https://registry.npmjs.org/@angular/material-moment-adapter/-/material-moment-adapter-12.2.13.tgz",
-      "integrity": "sha512-0XhkAhBalvNvIfvi9t5qDKaTBtOuIqd1cABfotyaTecWmw57VoH0KoJtNekXjxDcx71S0LaqeYXfqTo4m+duSA==",
-      "requires": {
-        "tslib": "^2.2.0"
-      }
-    },
     "@angular/platform-browser": {
       "version": "12.2.15",
       "resolved": "https://registry.npmjs.org/@angular/platform-browser/-/platform-browser-12.2.15.tgz",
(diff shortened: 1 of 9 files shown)


lucas-koehler (changes requested):
I think this only fixes part of the problem. While the display now correctly works for timezones UTC and UTC+x, it still shows the wrong date for UTC-x timezones (e.g. in the US). As far as I saw it, the  storage of the date now works correctly but I guess there might be additional problems converting the date back when showing it.

sdirix:
@TheMicDiet Will you take a look and handle the review comment by @lucas-koehler ?

TheMicDiet:
@sdirix yes, it should work now

sdirix (changes requested):
Thanks for the update! However I would prefer to use something else than moment. Moment is very large which is why we [removed it recently from the other renderer sets](https://github.com/eclipsesource/jsonforms/pull/1792). Can we also use dayjs here?

TheMicDiet:
@sdirix  I managed to replace moment with dayjs. However, I had to increase the target level in the tsconfig to es6, otherwise the date adapter wouldn't work. Is that an issue? Also, the date format is currently hard-coded to YYYY-MM-DD. I've seen that it is possible to change the date format via the UI schema in the React implementation. I've tried to implement this for the Angular date picker, but I haven't found a nice way to configure the formatting of an individual Angular date picker component yet :/

sdirix:
Sorry for the late reply! I took a look at the new code. 

> I managed to replace moment with dayjs.

Looks good!

> However, I had to increase the target level in the tsconfig to es6, otherwise the date adapter wouldn't work. Is that an issue?

Do you know any specifics of why it didn't work? Some class problem with es5?

As we are targeting later Angular versions I think this change should be fine. @lucas-koehler what do you think?

> Also, the date format is currently hard-coded to YYYY-MM-DD. I've seen that it is possible to change the date format in the React implementation. I've tried to implement this for the Angular date picker, but I haven't found a nice way to configure the formatting of an individual Angular date picker component yet :/

We had the same problems with localizing the data adapter which then influenced the whole app outside of JSON Forms too. Sadly I'm not aware of a good local solution at the moment.

lucas-koehler:
> > However, I had to increase the target level in the tsconfig to es6, otherwise the date adapter wouldn't work. Is that an issue?
> 
> Do you know any specifics of why it didn't work? Some class problem with es5?
> 
> As we are targeting later Angular versions I think this change should be fine. @lucas-koehler what do you think?

For me, updating to ES6 is fine. It seems to be supported by every reasonably current browser anyway :) https://caniuse.com/?search=es6

TheMicDiet:
> Sorry for the late reply! I took a look at the new code.
> 
> > I managed to replace moment with dayjs.
> 
> Looks good!
> 
> > However, I had to increase the target level in the tsconfig to es6, otherwise the date adapter wouldn't work. Is that an issue?
> 
> Do you know any specifics of why it didn't work? Some class problem with es5?
> 
> As we are targeting later Angular versions I think this change should be fine. @lucas-koehler what do you think?
> 
> > Also, the date format is currently hard-coded to YYYY-MM-DD. I've seen that it is possible to change the date format in the React implementation. I've tried to implement this for the Angular date picker, but I haven't found a nice way to configure the formatting of an individual Angular date picker component yet :/
> 
> We had the same problems with localizing the data adapter which then influenced the whole app outside of JSON Forms too. Sadly I'm not aware of a good local solution at the moment.



> Do you know any specifics of why it didn't work? Some class problem with es5?
If I run the examples with es5, I get the following error message:  
```
ERROR TypeError: Class constructor NativeDateAdapter cannot be invoked without 'new'
    at new DayJsDateAdapter (dayjs-date-adapter.ts:36:1)
```
Changing the target level to es6 seems to solve this problem.

sdirix:
1. Closed as we updated to pnpm with https://github.com/eclipsesource/jsonforms/pull/2148

PR #2037, opened 2022-10-25
fix visibility of error-icon in array

LukasBoll:


lucas-koehler (approved):
Thanks for the fix @LukasBoll . I tested it before the fix and could reproduce the error in the development app. With the fix, the issue is gone :)

PR #2038, opened 2022-10-25
Add Sort buttons to Angular Material ArrayLayoutRenderer and TableRenderer

fredjohnd:
This adds Sort functionality to Angular Material package, as long as `options.showSortButtons` is True.

This is my first PR for this project, so please let me know if something is not quite right.

Test coverage added as well as example in packages/examples.

imagem

### DIFF
--- a/packages/examples/src/examples/arrays-with-sorting.ts
+++ b/packages/examples/src/examples/arrays-with-sorting.ts
@@ -123,7 +123,7 @@ const actions = [
 registerExamples([
   {
     name: 'array-with-sorting',
-    label: 'array-with-sorting',
+    label: 'Array with sorting',
     data,
     schema,
     uischema,
--- a/packages/examples/src/examples/arrays-with-sorting.ts
+++ b/packages/examples/src/examples/arrays-with-sorting.ts
@@ -63,7 +63,7 @@ export const uischema = {
       type: 'Control',
       scope: '#/properties/comments',
       options: {
-        showSortButtons: false
+        showSortButtons: true
       }
     }
   ]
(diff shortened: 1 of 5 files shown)


packages/examples/src/examples/arrays-with-sorting.ts
```diff
@@ -0,0 +1,135 @@
...
+        uischema: uischema
+      }
+    }
+  },
+];
+
+registerExamples([
+  {
+    name: 'array-with-sorting',
+    label: 'array-with-sorting',
```
lucas-koehler, line 126:
```suggestion
    label: 'Array with sorting',
```

lucas-koehler (commented):
Hi @fredjohnd , thank you very much for your contribution :) The changes generally look good to me. Would it be possible to add an example for an array with detail? You could either add it to your new example or add another one.
The reason to have this in an example is that it's currently not possible to test the sort buttons of an array with detail in the examples. For a schema and uischema, you can have a look at `packages/examples/src/examples/arrays-with-detail.ts`

packages/examples/src/examples/arrays-with-sorting.ts
fredjohnd: Hi @lucas-koehler ,

The arrays-with-detail.ts example already renders the sort buttons as the example schema already had `showSortButtons: true` .

imagem

Let me know if you still require me to add a new example or modify the existing one?

lucas-koehler: Hi @fredjohnd thanks for the hint. In that case it's fine as it is, thanks!

lucas-koehler (approved):

PR #2040, opened 2022-11-08
add readOnly functionality to array and list

### LINKED ISSUE #1901: Arrays can be modified even with readonly property set on true
### Describe the bug

The buttons for adding or removing elements from an array are visible and functional in the form even if I set it for readonly.

### Expected behavior

I should not be able to modify the form in readonly mode.

1. Create a Form
2. Set a JSON with an array in its data property
3. Set the readonly property to true

### Screenshots

image

Google Chrome Versión 99.0.4844.51

### Framework

React

### RendererSet

Material

### Additional context

Dependencies used:
        "@emotion/react": "^11.5.0",
        "@emotion/styled": "^11.3.0",
        "@jsonforms/core": "3.0.0-alpha.3",
        "@jsonforms/material-renderers": "3.0.0-alpha.3",
        "@jsonforms/react": "3.0.0-alpha.3",
        "@mui/icons-material": "^5.2.0",
        "@mui/lab": "^5.0.0-alpha.58",
        "@mui/material": "^5.2.2",
        "@mui/styles": "^5.2.3",

LukasBoll:
closes #1901


sdirix:
@LukasBoll With this PR readOnly arrays don't even show their action buttons. Alternatively we could still show them but disable them. What do prefer personally?

LukasBoll:
There are two cases:

1) The array is set to read only permanently: In this case, it would be better to don’t show any buttons at all, because otherwise the user may expect some functionalities that aren´t supported.

2) The array can be disabled and enabled: While it would be better to enable/disable the buttons to draw attention to the functionality and the current state, hiding the buttons is still an acceptable solution.

Overall, I would not show the action buttons, since 1) is probably the basic case and it is misleading for the user if buttons without functionality are shown.

sdirix:
@lucas-koehler  With this PR readOnly arrays don't even show their action buttons. Alternatively we could still show them but disable them. What do prefer personally?

lucas-koehler:
@sdirix Ideally,we would support both cases described by @LukasBoll :

- don't show them for permanent readonly
- disable them when disabled by a rule

The reason is that I would expect a disabled button to have the possibility to get enabled.

If supporting both is not feasible, I tend to agree with @LukasBoll to not show them to not confuse users in the first case.

sdirix:
@LukasBoll please rebase or merge in the master branch. The files were moved.

sdirix (approved):
Works for me!

PR #2051, opened 2022-11-14
fix duplicate error message

### LINKED ISSUE #1684: Required bubble that appears when marking an array field as required warns of two errors instead of one
**Describe the bug**
Required bubble that appears when marking an array field as required warns of two errors instead of one.

**To Reproduce**
Use an empty object for data, use no uischema (autogenerated) and use the following schema object as an example:
`schema: {
          properties: {
              test: {
                type: "array",
                items: {
                    type: "object",
                    properties: {
                        "test1": {
                            label: "test1",
                            type: "string"
                        },
                        "test2": {
                            label: "test2",
                            type: "string"
                        },
                        "test3": {
                            label: "test3",
                            type: "string"
                        }
                    }
                }
              }
          },
          required: ["test"],
          type: "object"
        }`

**Expected behavior**
I expect there to be a bubble with a single error instead of 2.

**Screenshots**
image

 - Browser: Google Chrome
 - Version: 88.0.4324.96 (Official Build) (x86_64)

 - Framework: react
 - RendererSet: material

### LINKED ISSUE #1760: Duplicate validation messages in primitive array renderer
**Describe the bug**

For primitive arrays, all validation messages are combined into 1, and the top level validation for the array (is required, min amount of items) seems to duplicate itself.

The `error` property seems to contain:
* Validation messages for all items in the array
* Validation messages for the top level array itself (min/max items validation, is required, etc.)

**Screenshots**

Validation messages for the individual items are also in the top level array renderer:

Screenshot 2021-05-28 at 11 27 35

Top level validation messages are duplicated:

Screenshot 2021-05-28 at 11 27 09

**Expected behaviour**

I would expect the error property of the Array control to only contain validation for the top level. The individual items in the array could render their own validation messages.

Additionally, I would expect the "is required" or "should NOT have fewer than X items" validation messages to not be duplicated like this.

**To Reproduce**

Array layout renderer:

```ts
export class ArrayLayoutRenderer extends JsonFormsArrayControl implements OnInit {

  private addItem: (path: string, value: any) => () => void;
  private removeItems: (path: string, toDelete: number[]) => () => void;

  constructor(jsonFormsService: JsonFormsAngularService) {
    super(jsonFormsService);
  }

  remove(index: number): void {
    this.removeItems(this.propsPath, [index])();
  }
  add(): void {
    this.addItem(this.propsPath, createDefaultValue(this.scopedSchema))();
  }

  mapToProps(state: JsonFormsState): StatePropsOfArrayLayout {
    const props = mapStateToArrayLayoutProps(state, this.getOwnProps());
    return { ...props };
  }

  ngOnInit(): void {
    super.ngOnInit();

    const { addItem, removeItems } = mapDispatchToArrayControlProps(
      this.jsonFormsService.updateCore.bind(this.jsonFormsService)
    );
    this.addItem = addItem;
    this.removeItems = removeItems;
  }

  getProps(index: number): OwnPropsOfRenderer {
    return {
      schema: this.scopedSchema,
      path: Paths.compose(this.propsPath, `${index}`),
      uischema: controlWithoutLabel(`#`)
    };
  }

  trackByFn(index: number) {
    return index;
  }

}
```

Template file:

```html
<label class="h6">{{ label }}</label>

<p class="text-hint mt-2 mb-1">{{ uischema['description'] }}</p>

<p class="caption-2 text-danger my-1">{{ error }}</p>

<button (click)="add()">Add</button>

<ng-container *ngFor="
    let item of [].constructor(data);
    let idx = index;
    trackBy: trackByFn;
    last as last
">
    <jsonforms-outlet [renderProps]="getProps(idx)"></jsonforms-outlet>
    <button (click)="remove(idx)">delete</button>
</ng-container>
```

Schema for this individual question:

```json
{
  "list-question": {
    "type": "array",
    "minItems": 1,
    "items": {
      "type": "string",
      "minLength": 1
    }
  }
}
```

 - Framework: angular
 - RendererSet: custom, but based on Angular Material

LukasBoll:
subErrorsAt now only returning errors form child elements

closes #1760, closes #1684

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -424,4 +424,4 @@ const getErrorsAt = (
 export const errorAt = (instancePath: string, schema: JsonSchema) =>
   getErrorsAt(instancePath, schema, path => path === instancePath);
 export const subErrorsAt = (instancePath: string, schema: JsonSchema) =>
-  getErrorsAt(instancePath, schema, path => path.startsWith(instancePath));
+  getErrorsAt(instancePath, schema, path => path.startsWith(instancePath + '.'));
--- a/packages/core/test/reducers/core.test.ts
+++ b/packages/core/test/reducers/core.test.ts
@@ -1175,6 +1175,42 @@ test('subErrorsAt filters array inner', t => {
   t.deepEqual(filtered[0], state.errors[1]);
 });
 
+test('subErrorsAt only returning suberrors', t => {
+  const ajv = createAjv();
+  const schema: JsonSchema = {
+    type: 'object',
+    properties: {
+      numbers: {
+        title: 'Numbers',
+        type: 'array',
+        minItems: 1,
+        items: {
+          title: 'Type',
+          type: 'string',
+          enum: ['One', 'Two', 'Three']
+        },
+      }
+    }
+  };
+  const data: { numbers: string[] } = {
+    numbers: []
+  };
+  const v = ajv.compile(schema);
+  const errors = validate(v, data);
+
+  const state: JsonFormsCore = {
+    data,
+    schema,
+    uischema: undefined,
+    errors
+  };
+  const subErrors = subErrorsAt(
+    'numbers',
+    schema.properties.numbers.items as JsonSchema
+  )(state);
+  t.is(subErrors.length, 0);
+});
+
 test('subErrorsAt filters oneOf array inner', t => {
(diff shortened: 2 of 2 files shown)


TheZoker (approved):
LGTM, seems to work with the example from the issue:
image

sdirix (approved):
Thanks!

PR #2053, opened 2022-11-23
fix allOf in array

LukasBoll:


TheZoker (approved):
LGTM, can confirm that the change work with the example from the issue:
image

sdirix (approved):
LGTM

PR #2061, opened 2022-12-07
fix toDataPath function comment explaining correct functionality

jlarrubiaq:
I found the explanation of what `toDataPath` does is not correct. The new comment reflects the actual functionality.

### DIFF
--- a/packages/core/src/util/path.ts
+++ b/packages/core/src/util/path.ts
@@ -69,12 +69,15 @@ export const toDataPathSegments = (schemaPath: string): string[] => {
 };
 
 /**
- * Remove all schema-specific keywords (e.g. 'properties') from a given path.
+ * Convert a schema path (i.e. JSON pointer) to a data path.
+ * 
+ * Data paths can be used in field change event handlers like handleChange.
+ * 
  * @example
- * toDataPath('#/properties/foo/properties/bar') === '#/foo/bar')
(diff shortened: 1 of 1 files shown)


packages/core/src/util/path.ts
```diff
@@ -69,9 +69,12 @@ export const toDataPathSegments = (schemaPath: string): string[] => {
...
+ * Convert a schema path (i.e. JSON pointer) to a data path.
+ * 
+ * Data paths can be used in field change event handlers like handleChange.
+ * 
  * @example
- * toDataPath('#/properties/foo/properties/bar') === '#/foo/bar')
+ * toDataPath('#/properties/foo/properties/bar') === 'foo.bar')
  *
  * @param {string} schemaPath the schema path to be converted
  * @returns {string} the path without schema-specific keywords
```
sdirix, line 80: We should probably also adjust the return description, right?

sdirix (changes requested):
Thanks for the contribution!

packages/core/src/util/path.ts
jlarrubiaq: That was quick! I changed it

jlarrubiaq:
@sdirix I addressed your comment. Thanks!

sdirix (approved):
Thanks for the contribution

PR #2062, opened 2022-12-14
propagate readOnly prop in Angular detail-List

### LINKED ISSUE #1479: [angular] table does not correctly propagate read-only state to children
**Describe the bug**
When setting an UI schema containing a table to readonly, its children are still enabled.

LukasBoll:
closes #1479


sdirix (approved):
lgtm

Issue #1036, opened 2018-07-13
folder names should match package names


edgarmueller:

eneufeld:
We should probably also unify the naming schema, but this would be a big breaking change, as the react packages would need to be renamed ...

PR #2064, opened 2022-12-20
Adapt README URLs  to renamed vanilla and material renderers folders

lucas-koehler:
Related to https://github.com/eclipsesource/jsonforms/issues/1036


sdirix (approved):
Thanks!

PR #2072, opened 2023-01-09
Show description or tooltip on boolean controls

brockfanning:
The intended behavior here is that if a boolean control has a description, then it will appear as a tooltip. However if showUnfocusedDescription is true, then the description appears below as text. I tried to copy as much code from existing components as possible, but I'm happy to make any changes as needed.

### DIFF
--- a/packages/material-renderers/src/controls/MaterialBooleanControl.tsx
+++ b/packages/material-renderers/src/controls/MaterialBooleanControl.tsx
@@ -23,15 +23,17 @@
   THE SOFTWARE.
 */
 import isEmpty from 'lodash/isEmpty';
+import merge from 'lodash/merge';
 import React from 'react';
 import {
   isBooleanControl,
   RankedTester,
   rankWith,
-  ControlProps
+  ControlProps,
+  isDescriptionHidden
 } from '@jsonforms/core';
 import { withJsonFormsControlProps } from '@jsonforms/react';
-import { FormControlLabel, Hidden } from '@mui/material';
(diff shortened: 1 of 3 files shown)


packages/material-renderers/src/controls/MaterialBooleanControl.tsx
```diff
@@ -46,30 +48,64 @@ export const MaterialBooleanControl = ({
...
+  const appliedUiSchemaOptions = merge({}, config, uischema.options);
+
+  const showDescription = !isDescriptionHidden(
+    visible,
+    description,
+    false,
+    appliedUiSchemaOptions.showUnfocusedDescription
+  );
+
+  const showTooltip = !showDescription && !isDescriptionHidden(
```
lucas-koehler, line 65: Maybe add a short comment why `true`, `true` are handed in as the two last parameters to make it easier to understand in the future

lucas-koehler (commented):
Hello @brockfanning, thank you very much for your contribution :)
I tested it and it seems to work well.

I have two requests:

1. Could you extend one the examples to contain a description for the boolean control. E.g. you could extend packages/examples/src/examples/control-options.ts
2. Could you also add this behavior to the [MaterialBooleanToggleControl](packages/material-renderers/src/controls/MaterialBooleanToggleControl.tsx) as well? This way both boolean controls behave consistently.

lucas-koehler (approved):
LGTM. Thanks for the updates and again for the contribution :)
CI did not run again due to a first time contributor. However, I executed the build and unit tests locally  and they still worked :)

Issue #2073, opened 2023-01-09
Angular object renderer crashes example app for some examples containing references


lucas-koehler:
### Describe the bug

At least the examples `allOf`, `Huge Test` and `AnyOf OneOf AllOf Resolve` crash the example app when opened in the Angular Material examples.

All of them use `$ref` references in their schema and the same error message appears in the log. The Angular example app does not resolve references via an external dependency (e.g. `json-refs`). The same examples work for React Material.

The following two errors appear in the log whenever the issue occurs:
<details>
<summary>Show logs</summary>
Log 1:

```
ObjectControlRenderer.html:2 ERROR 
View_ObjectControlRenderer_0	@	ObjectControlRenderer.html:2
proxyClass	@	compiler.js:25075
logError	@	core.js:32811
handleError	@	core.js:7792
(anonymous)	@	core.js:29781
_ZoneDelegate.invoke	@	zone.js:409
Zone.run	@	zone.js:169
runOutsideAngular	@	core.js:28668
tick	@	core.js:29781
(anonymous)	@	core.js:29629
_ZoneDelegate.invoke	@	zone.js:409
onInvoke	@	core.js:28769
_ZoneDelegate.invoke	@	zone.js:408
Zone.run	@	zone.js:169
run	@	core.js:28623
next	@	core.js:29628
SafeSubscriber.__tryOrUnsub	@	Subscriber.js:192
SafeSubscriber.next	@	Subscriber.js:130
Subscriber._next	@	Subscriber.js:76
Subscriber.next	@	Subscriber.js:53
Subject.next	@	Subject.js:47
emit	@	core.js:26042
checkStable	@	core.js:28691
onLeave	@	core.js:28819
onInvokeTask	@	core.js:28763
_ZoneDelegate.invokeTask	@	zone.js:442
Zone.runTask	@	zone.js:214
ZoneTask.invokeTask	@	zone.js:525
invokeTask	@	zone.js:1727
globalCallback	@	zone.js:1758
globalZoneAwareCallback	@	
```

Log2:
```
ObjectControlRenderer.html:2 ERROR CONTEXT DebugContext_ {view: {…}, nodeIndex: 1, nodeDef: {…}, elDef: {…}, elView: {…}}
View_ObjectControlRenderer_0	@	ObjectControlRenderer.html:2
proxyClass	@	compiler.js:25075
logError	@	core.js:32811
handleError	@	core.js:7797
(anonymous)	@	core.js:29781
[...]

Issue #1706, opened 2021-02-26
Improve internal example app


sdirix:
- [x] Use a unique path for each example
- [x] Offer a built version of the example app so we can build all example app variants with each JSON Forms PR and link them.
  - [x]  https://github.com/eclipsesource/jsonforms/pull/2056: Site with static index.html and links to react vanilla and react material examples. Use rollup to bundle examples
  - [x] Configure Netlify to build them for each PR
  - [x]  https://github.com/eclipsesource/jsonforms/pull/2074: Angular material example
  - [x] https://github.com/eclipsesource/jsonforms/pull/2074 Vue 3 Vanilla
  - [x] Vue 2 Vanilla: Will be handled in separate issue: https://github.com/eclipsesource/jsonforms/issues/2087
- [x] https://github.com/eclipsesource/jsonforms/pull/2101: Cleanup examples (e.g. add description, ordering)
---
Improving the webpack dev builds was descoped and is tracked in issue https://github.com/eclipsesource/jsonforms/issues/2093

lucas-koehler:
All points are resolved or extracted to their own issues.

PR #2074, opened 2023-01-10
Add bundled examples for Angular Material and Vue Vanilla

lucas-koehler:
* Add rollup config to bundle the angular material example
* Add rollup config to bundle the vue 3 vanilla example
* Add to examples app

Notes:

- Some examples crash the angular example app when selected. However, this is not an issue of the bundled app but also happens in the existing dev example and is tracked as a separate issue: #2073
- Only added example app for vue 3 because the vue 2 dev app currently does not work: https://github.com/eclipsesource/jsonforms/issues/2075

Part of #1706


sdirix (approved):
Works for me!

PR #2076, opened 2023-01-11
Use type only imports 

wienczny:
Use type only imports supported since TS 3.8+ to improve module isolation.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html


sdirix:
@wienczny Thanks for the contribution!

sdirix (approved):
Changes look good to me! Thanks!

PR #2079, opened 2023-01-11
Use aria-describedby on checkboxes with descriptions

### LINKED ISSUE #2078: Screenreader support on MUI checkboxes with descriptions

The descriptions of checkboxes are not read by screenreaders. This means that non-sighted users will not benefit from checkbox help text when using the form.

The checkbox should use [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) and reference the help text elements so that screenreaders will read them.

I'm not aware of other solutions.

### Framework

React

### RendererSet

Material

### Additional context

_No response_

brockfanning:
Fixes #2078 

I should note that the problem this solves also affects other components, like the text input. But I wanted to start with a simpler case (checkboxes) before tackling that one, to make sure the team is OK with this approach.

### DIFF
--- a/packages/material-renderers/src/controls/MaterialBooleanControl.tsx
+++ b/packages/material-renderers/src/controls/MaterialBooleanControl.tsx
@@ -83,9 +83,23 @@ export const MaterialBooleanControl = ({
     : null;
   const secondFormHelperText = showDescription && !isValid ? errors : null;
 
+  let ariaDescribedBy = '';
+  const tooltipId = `${id}-tip`;
+  const helpId1 = `${id}-help1`;
+  const helpId2 = `${id}-help2`;
+  if (showTooltip) {
+    ariaDescribedBy += tooltipId;
+  }
+  if (firstFormHelperText) {
+    ariaDescribedBy += ` ${helpId1}`;
+  }
+  if (secondFormHelperText) {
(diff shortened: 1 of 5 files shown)


sdirix:
Thanks for the contribution!

brockfanning:
@sdirix No problem! I need to make the same type of change for text inputs, but I wanted to make sure this approach (passing along the inputProps) is OK with you. Should I proceed?

sdirix (approved):
Thanks again! The code looks good.

I thought about whether it's worth to `useMemo` the inputProps objects but I actually think it's not necessary here. In almost all use cases the memo will not apply.

The only use case I could think off is when someone updates the JSON Schema at runtime with a new description. This is really an edge case, and if someone does that they probably don't care that the input is also rerendered (maybe there is even some additional memo barrier further within Material UI)

So the code looks good! However it would be nice if you could add test cases

sdirix:
@brockfanning Will you add a testcase?

brockfanning:
@sdirix I just pushed up a couple of tests - please let me know if that is not what you had in mind. I only did tests for MaterialBooleanToggleControl because I did not see any existing tests for MaterialBooleanControl. Did I miss those?

sdirix:
Thanks for the contribution :heart: ;)

PR #2082, opened 2023-01-19
Improve Translator type

sdirix:
The Translator type is an overloaded function. While the current type considers the 'defaultMessage' to be either 'string' or 'undefined' it does not handle the case where it could be 'string | undefined' while the implementation obviously handles this fine.

This change adds this additional case to the 'Translator' type and optionalizes the 'defaultMessage' parameter. This allows a more convenient use of this type by adopters.


lucas-koehler (approved):
LGTM, thanks!

PR #2095, opened 2023-02-16
Add a failWhenUndefined option to the SchemaBasedCondition

clemens-msupply:
This PR adds an option to mark the scope of a rule condition as required.

The original proposed solution was to add:
`whenUndefined: 'fail' | 'success'`
but after understanding the problem a bit better I think the main problem is that you can't define the condition scope as required and adding this option keeps it a bit closer to JSONSchema.

I added a comment about my interpretation about whats happening in the code. Let me know what you think, I might be wrong about it and AJV just does something weird... I also still like `whenUndefined` since it might actually be more user friendly.

Original discussion:
https://jsonforms.discourse.group/t/no-rule-validation-for-array-contains-value-when-array-undefined/1308/7

### DIFF
--- a/packages/core/src/models/uischema.ts
+++ b/packages/core/src/models/uischema.ts
@@ -135,6 +135,19 @@ export interface LeafCondition extends Condition, Scoped {
 
 export interface SchemaBasedCondition extends Condition, Scoped {
   schema: JsonSchema;
+
+  /**
+   * When the scope resolves to undefined and `failWhenUndefined` is set to `true`, the condition
+   * will fail. Therefore the reverse effect will be applied.
+   *
+   * Background:
+   * Most JSON Schemas will successfully validate against `undefined` data. Specifying that a
+   * condition shall fail when data is `undefined` requires to lift the scope to being able to use
+   * JSON Schema's `required`.
+   *
+   * Using `failWhenUndefined` allows to more conveniently express this condition.
+   */
+  failWhenUndefined?: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


sdirix:

I'll talk with the team in regards whether we prefer `whenUndefined` or `scopeRequired` :+1:.

What I don't understand is your comment in the discussion board where the reverse effect and condition did not work for you, i.e.
>  am pretty sure I tried the reverse rule and it didn’t work for me

Looking at the code here and the suggested improvements of `scopeRequired` only applying in the `true` case, this seems for me to indicate that reversing the effect/rule for the `undefined` case should also have worked. Can you verify what's going on?

sdirix (dismissed):
Very nice improvement to the docs! Thanks for the contribution :heart:

clemens-msupply:
I created some examples for the "range" rule and the "array contains" rule:

https://github.com/clemens-msupply/jsonforms-react-seed/tree/rule-condition-with-undefined-values

I reversed the rule by using "not" but it both behaves the same (dependent controls are showing on initial load). Is this what you meant?

sdirix:
Heads up: We'll take a look at this again soon

sdirix (approved):

FedeTommi:
> I created some examples for the "range" rule and the "array contains" rule:
> 
> https://github.com/clemens-msupply/jsonforms-react-seed/tree/rule-condition-with-undefined-values
> 
> I reversed the rule by using "not" but it both behaves the same (dependent controls are showing on initial load). Is this what you meant?

Hi there, I experienced the same issue here https://jsonforms.discourse.group/t/default-hide-for-array-contains-value-when-array-undefined/2152. Are there any updates on this?

sdirix:
I answered in the community thread :+1:

PR #2098, opened 2023-02-20
fix: issue/2090 mapStateToCellProps should use translated error messages

codefactor:


sdirix:

Please make sure to sign the CLA in a way which is recognizable by the CLA bot.

codefactor:
@sdirix ,
I've now updated the CLA so the bot should be satisfied.

codefactor:
@sdirix ,
Try as I might I could not figure out the unit test so I just rolled it back. I believe the tests will pass without covering the case where the `mapStateToCellProps` function is given a `state` which contains an error object array. The commit will be in the history of the `cell.test.ts` I believe, so it would be possible to revert the revert in the future if adding unit test coverage to those lines is important.

Unfortunately, I can't quite figure out why there is a `test/util/cell.test.ts exited due to SIGABRT` error happening. After some more debugging there does appear to be some minor issues regarding the setup and the assertions - you can see the unit test I removed here:

https://github.com/eclipsesource/jsonforms/pull/2098/commits/9828188f4d2c1b2f41107471cd0b7185bd1e72cf#diff-7d5366084c8446977a6750222516b31ab17a6b20eda83d02a9ec9848f28cd172L277-L298

I'm pretty sure the `ownProps` is missing the `schema` and the `path` props which is causing the `errorAt` functions to filter out all errors, also the `errors` array is in the wrong part of the state, I believe it should be in `state.jsonforms.core.errors` and not `state.jsonforms.errors`. But even if I correct all of those issues, the tests still crash. Under normal circumstances, this should cause the `mapStateToCellProps` function to return an empty string for the `errors` prop and then fail the test with an assertion error; however, the whole test just bombs.

Please let me know how you think we should proceed on this one.

sdirix:
@codefactor Thanks for the work and analysis. We'll take a look :+1:

codefactor:
@sdirix what are the next steps on this PR? Do you need more test coverage to get it merged?

sdirix:
Depending on our work load it can take some time until we are able to look at topics like this in detail. The good news is I was able to fix the test ;)

sdirix (approved):
Works for me!

PR #2099, opened 2023-02-20
translate array controls in material renderers

LukasBoll:
translate array controls in material renderers

### DIFF
--- a/packages/core/src/i18n/arrayTranslations.ts
+++ b/packages/core/src/i18n/arrayTranslations.ts
@@ -1,0 +1,38 @@
+export interface ArrayDefaultTranslation {
+    key: ArrayTranslationEnum,
+    default: (variable?: string) => string
+}
+
+export enum ArrayTranslationEnum{
+    addTooltip = 'addTooltip',
+    addAriaLabel = 'addAriaLabel',
+    removeTooltip = 'removeTooltip',
+    upAriaLabel = 'upAriaLabel',
+    downAriaLabel = 'downAriaLabel',
+    noSelection = 'noSelection',
+    removeAriaLabel = 'removeAriaLabel',
+    noDataMessage = 'noDataMessage',
+    deleteDialogTitle = 'deleteDialogTitle',
+    deleteDialogMessage = 'deleteDialogMessage',
+    deleteDialogAccept = 'deleteDialogAccept',
+    deleteDialogDecline = 'deleteDialogDecline'
+}
+
+export type ArrayTranslations = {
+    [key in ArrayTranslationEnum]?: string
+}
+
+export const arrayDefaultTranslations: ArrayDefaultTranslation[] = [
+    {key: ArrayTranslationEnum.addTooltip, default: (input) => input?`Add to ${input}`:'Add'},
+    {key: ArrayTranslationEnum.addAriaLabel, default: (input) => input?`Add to ${input}`:'Add'},
+    {key: ArrayTranslationEnum.removeTooltip, default: () => 'Delete'},
+    {key: ArrayTranslationEnum.upAriaLabel, default: (input) => `Move ${input} up`},
+    {key: ArrayTranslationEnum.downAriaLabel, default: (input) => `Move ${input} down`},
+    {key: ArrayTranslationEnum.removeAriaLabel, default: () => 'Delete button'},
+    {key: ArrayTranslationEnum.noDataMessage, default: () => 'No Data'},
+    {key: ArrayTranslationEnum.noSelection, default: () => 'No Selection'},
+    {key: ArrayTranslationEnum.deleteDialogTitle, default: () => 'Confirm Deletion'},
+    {key: ArrayTranslationEnum.deleteDialogMessage, default: () => 'Are you sure you want to delete the selected entry?'},
+    {key: ArrayTranslationEnum.deleteDialogAccept, default: () => 'Yes'},
+    {key: ArrayTranslationEnum.deleteDialogDecline, default: () => 'No'}
+]
--- a/packages/core/src/i18n/i18nUtil.ts
+++ b/packages/core/src/i18n/i18nUtil.ts
@@ -3,6 +3,7 @@ import { isInternationalized, Labelable, UISchemaElement } from '../models';
 import { getControlPath } from '../reducers';
 import { formatErrorMessage } from '../util';
 import type { i18nJsonSchema, ErrorTranslator, Translator } from './i18nTypes';
+import { ArrayDefaultTranslation, ArrayTranslations } from './arrayTranslations'
 
 export const getI18nKeyPrefixBySchema = (
   schema: i18nJsonSchema | undefined,
@@ -45,6 +46,13 @@ export const getI18nKey = (
   key: string
 ): string => {
   return `${getI18nKeyPrefix(schema, uischema, path)}.${key}`;
+};
+
+export const addI18nKeyToPrefix = (
+  i18nKeyPrefix: string,
+  key: string
+): string => {
+  return `${i18nKeyPrefix}.${key}`;
 };
 
 export const defaultTranslator: Translator = (_id: string, defaultMessage: string | undefined) => defaultMessage;
@@ -122,4 +130,18 @@ export const deriveLabelForUISchemaElement = (uischema: Labelable<boolean>, t: T
   const i18nKeyPrefix = getI18nKeyPrefixBySchema(undefined, uischema);
   const i18nKey = typeof i18nKeyPrefix === 'string' ? `${i18nKeyPrefix}.label` : stringifiedLabel;
   return t(i18nKey, stringifiedLabel, { uischema: uischema });
+}
+
+export const getArrayTranslations = (
+  t: Translator,
+  defaultTranslations: ArrayDefaultTranslation [],
+  i18nKeyPrefix: string,
+  label: string
+): ArrayTranslations => {
+  const translations:ArrayTranslations = {};
+  defaultTranslations.forEach((controlElement)=>{
+    const key = addI18nKeyToPrefix(i18nKeyPrefix, controlElement.key);
+    translations[controlElement.key]=t(key, controlElement.default(label));
+  })
+  return translations;
 }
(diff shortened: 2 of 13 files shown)


packages/core/src/i18n/i18nUtil.ts
```diff
@@ -123,3 +131,19 @@ export const deriveLabelForUISchemaElement = (uischema: Labelable<boolean>, t: T
...
 }
+
+export const translateElements = (
+  t: Translator,
+  controlElements: DefaultTranslation [],
+  i18nKeyPrefix: string,
+  label: string
+): ArrayTranslations => {
+  const translations:ArrayTranslations = {};
+  console.log("translating...");
```
lucas-koehler, line 142: I think this might be a leftover from testing? Please remove this.

packages/core/src/i18n/i18nUtil.ts
```diff
@@ -123,3 +131,19 @@ export const deriveLabelForUISchemaElement = (uischema: Labelable<boolean>, t: T
...
+  controlElements.forEach((controlElement)=>{
+    const key = addI18nKeyToPrefix(i18nKeyPrefix, controlElement.key);
+    translations[controlElement.key]=t(key, controlElement.default(label));
+    console.log("translations: ", translations);
```
lucas-koehler, line 146: I think this might be a leftover from testing? Please remove this.

packages/core/src/i18n/i18nUtil.ts
```diff
@@ -123,3 +131,19 @@ export const deriveLabelForUISchemaElement = (uischema: Labelable<boolean>, t: T
   const i18nKey = typeof i18nKeyPrefix === 'string' ? `${i18nKeyPrefix}.label` : stringifiedLabel;
   return t(i18nKey, stringifiedLabel, { uischema: uischema });
 }
+
+export const translateElements = (
+  t: Translator,
+  controlElements: DefaultTranslation [],
```
lucas-koehler, line 137: Please rename this to better represent what is expected here: The default translations. Thus, I suggest `defaultTranslations`

packages/core/src/i18n/arrayTranslations.ts
```diff
@@ -0,0 +1,40 @@
...
+    {key: ArrayTranslationEnum.removeTooltip, default: () => 'Delete'},
+    {key: ArrayTranslationEnum.upAriaLabel, default: (input) => `Move ${input} up`},
+    {key: ArrayTranslationEnum.downAriaLabel, default: (input) => `Move ${input} down`},
+    {key: ArrayTranslationEnum.removeAriaLabel, default: () => 'Delete button'},
+    {key: ArrayTranslationEnum.noDataMessage, default: () => 'No Data'},
+    {key: ArrayTranslationEnum.noSelection, default: () => 'No Selection'},
+    {key: ArrayTranslationEnum.deleteDialogHeader, default: () => 'Confirm Deletion'},
+    {key: ArrayTranslationEnum.deleteDialogMsg, default: () => 'Are you sure you want to delete the selected entry?'},
+    {key: ArrayTranslationEnum.deleteDialogAccept, default: () => 'YES'},
+    {key: ArrayTranslationEnum.deleteDialogDecline, default: () => 'NO'}
```
lucas-koehler, line 37: Why are these translations in all caps? As I saw it in the changes in `DeleteDialog.tsx` this was not the case before. I think we should stay with the old capitalization, i.e. `Yes` and `No`.

packages/material-renderers/src/complex/DeleteDialog.tsx
```diff
@@ -37,13 +37,17 @@ export interface DeleteDialogProps {
     onClose(): void;
     onConfirm(): void;
     onCancel(): void;
+    headerText: string,
+    msg: string,
```
lucas-koehler, line 41: No need to safe a few characters here :D Please use the full word `message`

packages/core/src/util/renderer.ts
```diff
@@ -1012,6 +1018,7 @@ export const mapStateToOneOfProps = (
 export interface StatePropsOfArrayLayout extends StatePropsOfControlWithDetail {
   data: number;
   minItems?: number;
+  translations?: ArrayTranslations;
```
lucas-koehler, line 1021:
```suggestion
  translations: ArrayTranslations;
```
I think this should not be optional as we always have default values to initialize these.
If these are optional, renderers need to consider this and have their own defaults again.
@sdirix Do you have any concerns with adding these as non-optional?

packages/material-renderers/src/complex/MaterialTableControl.tsx
```diff
@@ -65,6 +65,7 @@ import NoBorderTableCell from './NoBorderTableCell';
 import TableToolbar from './TableToolbar';
 import { ErrorObject } from 'ajv';
 import merge from 'lodash/merge';
+import { ArrayTranslations } from '@jsonforms/core/lib/i18n/arrayTranslations';
```
lucas-koehler, line 68: This should be imported directly from the jsonforms/core package. If that is not possible, please export the new definitions appropriately.
```suggestion
import { ArrayTranslations } from '@jsonforms/core';
```

packages/material-renderers/src/complex/TableToolbar.tsx
```diff
@@ -37,6 +37,7 @@ import { Grid, Typography } from '@mui/material';
 import AddIcon from '@mui/icons-material/Add';
 import ValidationIcon from './ValidationIcon';
 import NoBorderTableCell from './NoBorderTableCell';
+import { ArrayTranslations } from '@jsonforms/core/lib/i18n/arrayTranslations';
```
lucas-koehler, line 40: This should be imported directly from the jsonforms/core package. If that is not possible, please export the new definitions appropriately.
```suggestion
import { ArrayTranslations } from '@jsonforms/core';
```

packages/material-renderers/src/layouts/ArrayToolbar.tsx
```diff
@@ -8,20 +8,23 @@ import {
 import AddIcon from '@mui/icons-material/Add';
 import React from 'react';
 import ValidationIcon from '../complex/ValidationIcon';
+import { ArrayTranslations } from '@jsonforms/core/lib/i18n/arrayTranslations';
```
lucas-koehler, line 11: This should be imported directly from the jsonforms/core package. If that is not possible, please export the new definitions appropriately.
```suggestion
import { ArrayTranslations } from '@jsonforms/core';
```

packages/core/src/i18n/arrayTranslations.ts
```diff
@@ -0,0 +1,40 @@
...
+export enum ArrayTranslationEnum{
+    addTooltip = 'addTooltip',
+    addAriaLabel = 'addAriaLabel',
+    removeTooltip = 'removeTooltip',
+    upAriaLabel = 'upAriaLabel',
+    downAriaLabel = 'downAriaLabel',
+    noSelection = 'noSelection',
+    removeAriaLabel = 'removeAriaLabel',
+    noDataMessage = 'noDataMessage',
+    deleteDialogHeader = 'deleteDialogHeader',
```
lucas-koehler, line 15: I think title more precisely describes what this is for.
```suggestion
    deleteDialogTitle = 'deleteDialogTitle',
```

packages/material-renderers/src/complex/DeleteDialog.tsx
```diff
@@ -37,13 +37,17 @@ export interface DeleteDialogProps {
     onClose(): void;
     onConfirm(): void;
     onCancel(): void;
+    headerText: string,
```
lucas-koehler, line 40: I think title describes more precisely what this is used for. Please rename to `title`

lucas-koehler (changes requested):
Hi @LukasBoll , thanks for your pull request :)
I like the general approach of making the array translations available in a typed way.
Please have a look at the inline comments. They are mainly about namings to make things more clear.

Also, testing that translations are properly applied is pretty inconvenient right now. I suggest adding an additional example similar to packages/examples/src/examples/arrays.ts.
The new example could be named Array (i18n)
You can leave out the non-array property and probably do not need the actions to toggle sorting. Instead,  the sorting buttons should be shown in the new example to allow checking their aria labels.

packages/examples/src/examples/arraysI18n.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+};
+
+export const uischema = {
+  type: 'VerticalLayout',
+  elements: [
+    {
+      type: 'Control',
+      scope: '#/properties/foo'
+    },
```
lucas-koehler, line 62: This is not needed for this example => please delete

packages/examples/src/examples/arraysI18n.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+};
+export const translate: Translator = (key: string, defaultMessage: string) => {
+  console.log(key);
+  return get(translations, key) ?? defaultMessage;
+};
+
+registerExamples([
+  {
+    name: 'array (i18n)',
```
lucas-koehler, line 102: The name is used as part of the url fragment. Thus, it's better to avoid spaces and parantheses.
```suggestion
    name: 'array-i18n',
```

lucas-koehler (changes requested):
Thanks for the updates :) I just have three small comments.

LukasBoll:
Hi @lucas-koehler,
thank you for the comments and the feedback! I adjusted everything accordingly!

packages/examples/src/examples/arraysI18n.ts
```diff
@@ -0,0 +1,107 @@
...
+  comments: {
+    [ArrayTranslationEnum.noDataMessage]: 'Be the first to write a comment',
+    [ArrayTranslationEnum.addTooltip]: 'Add a Comment',
+    [ArrayTranslationEnum.deleteDialogAccept]: 'Delete!',
+    [ArrayTranslationEnum.deleteDialogDecline]: 'Cancel!',
+    [ArrayTranslationEnum.deleteDialogMessage]: 'Are you sure you want to delete this comment?'
+  }
+};
+export const translate: Translator = (key: string, defaultMessage: string) => {
+  console.log(key);
```
lucas-koehler, line 91: Sorry I forgot this earlier. Please remove this console log before merging :)

lucas-koehler (commented):
LGTM now, thanks! I just marked one console.log line to remove that I overlooked earlier.

packages/examples/src/examples/arraysI18n.ts
LukasBoll: Thank you, I just removed the console log

PR #2100, opened 2023-02-22
Allow random fields in schema

buremba:
The Json Schema spec lets us have additional properties in the schema. https://swagger.io/docs/specification/openapi-extensions/

### DIFF
--- a/packages/core/src/models/jsonSchema4.ts
+++ b/packages/core/src/models/jsonSchema4.ts
@@ -155,4 +155,5 @@ export interface JsonSchema4 {
 
   format?: string;
   const?: any;
+  [x: string | number | symbol]: unknown;
 }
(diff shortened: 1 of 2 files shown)


lucas-koehler:
Hello @buremba ,
thanks for your contribution :heart: . I'll discuss with the team whether we want to relax the type like this.
Before we could merge this, you need to sign the Eclipse CLA (see comment https://github.com/eclipsesource/jsonforms/pull/2100#issuecomment-1439329066)

Also note that `symbol` cannot be used as a key here. This currently fails the build.

buremba:
Done, thanks @lucas-koehler! This feature is useful to me primarily because I infer the uischema and custom renderer from the original json schema properties with `x-jsonform` prefix.

buremba:
@lucas-koehler It looks like manual action needed to trigger CI: https://github.com/eclipsesource/jsonforms/actions/runs/4263956482

Issue #2103, opened 2023-02-27
Vanilla Renderer Radio/Checkbox: add Option for vertical/horizontal layout


rsoika:
### Is your feature request related to a problem? Please describe.

Currently the Readio-Button has always a horizontal layout. This is because of the hard coded css style 'flexDirection=row'.

### Describe the solution you'd like

The Readio-Button Layout can be optimized in an horizontal or vertical design by introducing a new option

     orientation=vertically|horizontally

In this way the layout can be changed by the UISchema:

```yaml
{
  "type": "VerticalLayout",
  "elements": [
    {
      "type": "Group",
      "label": "Simple enum",
      "elements": [
        {
          "type": "Control",
          "scope": "#/properties/exampleRadioEnum",
          "options": {
            "format": "radio",
            "orientation": "vertical"
          }
        }
      ]
    },
    {
      "type": "Group",
      "label": "One of Enum",
      "elements": [
        {
          "type": "Control",
          "scope": "#/properties/exampleRadioOneOfEnum",
          "options": {
            "format": "radio"
          }
        }
      ]
    }
  ]
}
```

The new implementation in the file 'RadioGroup.tsx' can look like this:

```javascript
....
  let groupStyle: { [x: string]: any } = {};
  // compute flexDirection based on the optional option 'orientation=vertical|horizontal'
  if (hasRadioClass) {
    groupStyle = {
      display: 'flex',
      flexDirection: ('vertical' === uischema.options.orientation) ? 'row' : 'column'
    };
  }
....
```

Result:

![image](https://user-images.githubusercontent.com/1339124/221670127-31f18f95-1479-4d02-a381-858418dea1ff.png)

### Describe alternatives you've considered

An alternative would be to allow overwriting or providing optional layout classes. 

### Framework

React

### RendererSet

Vanilla

### Additional context

See discussion here: https://jsonforms.discourse.group/t/is-it-possible-to-arange-radio-checkbox-options-vertical/1352/6

PR #2104, opened 2023-02-27
Vanilla Renderer Radio/Checkbox: add Option for vertical/horizontal layout

rsoika:
See Issue #2103

### DIFF
--- a/packages/vanilla-renderers/src/controls/RadioGroup.tsx
+++ b/packages/vanilla-renderers/src/controls/RadioGroup.tsx
@@ -73,7 +73,7 @@ export const RadioGroup = ({
   if (hasRadioClass) {
     groupStyle = {
       display: 'flex',
-      flexDirection: ('vertical' === uischema.options.orientation) ? 'column' : 'row'
+      flexDirection: ('vertical' === appliedUiSchemaOptions.orientation) ? 'column' : 'row'
     };
   }
   return (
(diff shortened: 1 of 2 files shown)


lucas-koehler (approved):
Thanks for the contribution :) Code changes LGTM.

packages/vanilla-renderers/src/controls/RadioGroup.tsx
```diff
@@ -70,10 +70,11 @@ export const RadioGroup = ({
   );
   const hasRadioClass = !radioControl || radioControl === 'radio';
   let groupStyle: { [x: string]: any } = {};
+  // compute flexDirection based on the optional option 'orientation=vertical|horizontal'
   if (hasRadioClass) {
     groupStyle = {
       display: 'flex',
-      flexDirection: 'row'
+      flexDirection: ('vertical' === uischema.options.orientation) ? 'column' : 'row'
```
sdirix, line 77: We need to access the `appliedUiSchemaOptions` instead of the `uischema.options` directly. Then probably the test cases will also work again.
```suggestion
      flexDirection: ('vertical' === appliedUiSchemaOptions.orientation) ? 'column' : 'row'
```

sdirix (changes requested):

sdirix (approved):
Thanks for the contribution!

Issue #1551, opened 2020-02-26
Replace TSLint with ESLint and Prettier


sdirix:
TSLint is officially deprecated. Move to ESLint + Prettier: https://prettier.io/docs/en/integrating-with-linters.html

lucas-koehler:
- Integrate Prettier in ESLint so that prettier formatting is applied on linting.
- For prettier:
  - use single quotes
  - use line ending auto to be compatible with Windows development using default Git settings.
- Prefer rule recommendations over explicitly configuring many.
- Consider making the packages' configs self-contained in each package as long as there is not too much duplication

lucas-koehler:
Fixed via #2112 . Follow up tasks tracked in #2131

PR #2112, opened 2023-03-10
Replace TSLint with ESLint and Prettier

lucas-koehler:
- Configure ESLint and Prettier for each package
- Use ESLint 7.x instead of 8.x because the Angular eslint plugin for Angular 12 does not work with ESLint 8.x
- Did not apply the formatting yet

#1551


.vscode/settings.json
```diff
@@ -1,6 +1,46 @@
...
     "editor.formatOnSave": false,
-    "editor.defaultFormatter": "esbenp.prettier-vscode",
-    "prettier.jsxSingleQuote": true,
-    "prettier.singleQuote": true
+  },
+  "[javascriptreact]": {
+    "editor.codeActionsOnSave": {
+      "source.fixAll.eslint": true
+    },
+    "editor.formatOnSave": false
```
sdirix, line 21: Do we need this additional `editor.formatOnSave: false`?

sdirix (commented):
I tried the configured linter in the various packages.

- I got a lot of `Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any` warnings reported. Do you intend to fix them with this PR or shall we disable the rule for now to not make the warnings meaningless? This could be handled in a follow up PR then
- There are a lot of `Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins` errors. Do you intend to fix them with this PR?
- There are warning like `'_key' is defined but never used                                           @typescript-eslint/no-unused-vars`. Personally I prefer to specify an exception to this rule for variables starting with an underscore. Do you agree? If yes, can you add it?
- The `lint` script is also linting config files like `rollup.config.js`. This should either be avoided or they need specific rules. Personally I would not lint these files.
- There is warning when running `lint` in the react packages `Warning: React version not specified in eslint-plugin-react settings. See https://github.com/jsx-eslint/eslint-plugin-react#configuration`
- Some of the test files should probably be excluded as they do things like `require` which the linter doesn't like

Also:
 - Can you add markdownlint to the recommended extensions?
 - There is no linter yet in the vue packages

I think the PR should consist of three commits in the end:
 - A commit adding all the configurations
 - A commit applying all autofixable resolutions
 - A commit fixing all remaining errors and warnings

packages/core/.eslintrc.js
```diff
@@ -0,0 +1,21 @@
...
+module.exports = {
+  root: true,
+  parser: '@typescript-eslint/parser',
+  ignorePatterns: [
+    'dist',
+    'docs',
+    'lib',
+    'node_modules',
+    '*.d.ts',
+    'src/models/jsonSchema*',
```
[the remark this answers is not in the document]
lucas-koehler: The *.d.ts came from a template and the src/models/jsonSchema* was already ignored in the tslint config. Probably the idea was that these files were mostly copied. However, thinking about it again, I think we can include both.

.vscode/settings.json
lucas-koehler: The problem for me is that the prettier extension's format on save does not result in the same format as the eslint --fix. First, I thought this was a problem with the ESLint config. However, using the prettier cli, i.e. `prettier --write some/file.ts`, matches the output from ESLint. This means the Prettier extension does not behave the same as the cli. So far, I have not been able to figure out how to fix this. This is surprising as the the prettier extension should use the local prettier version and config. For some reason, this does not work as expected.

lucas-koehler: Nevermind, the reason was that the default formatter was overridden for some languages in my user vscode config. This then trumps the general default formatter setting

lucas-koehler:
@sdirix I improved the config. See the commit messages for the changes. I am going to squash the changes to have the three suggested commits in the end

sdirix (commented):
LGTM so far

sdirix (approved):
LGTM! Thanks!

PR #2120, opened 2023-03-21
Transform CategorizationRenderer from class to functional

### LINKED ISSUE #2097: React Vanilla Categorization Renderer should handle updated schemas
### Describe the bug

The React Vanilla Categorization renderer maintains local state about the currently selected category. This state is not properly updated when the UI Schema changes and therefore may return a previous category which no longer fits the JSON Schema.

### Expected behavior

Whenever a prop of the React Vanilla Categorization renderer changes it should update accordingly.

1. Render a categorization with React Vanilla
2. Select a category
3. Replace schema and uischema with a new Categorization but whose content is completely different
4. Still the old category of the previous uischema is used, leading for example to "No applicable cell" errors

The problematic code seems to be [here](https://github.com/eclipsesource/jsonforms/blob/386174994eabc41758450b8d551e1c859498a468/packages/vanilla-renderers/src/complex/categorization/CategorizationRenderer.tsx#L93-L97).

### Screenshots

_No response_

n/a

3.0.0

### Framework

React

### RendererSet

Vanilla

### Additional context

Reported in this [discussion](https://jsonforms.discourse.group/t/how-to-re-inizalize-a-complex-json-form/1338)

TheZoker:
This should fix #2097


sdirix (changes requested):
Please add a test case.

I don't think the error is fixed. The current category is still saved in state and is not updated on ui schema changes.

packages/vanilla-renderers/src/complex/categorization/CategorizationRenderer.tsx
```diff
@@ -22,88 +22,105 @@
...
+  useJsonForms,
   withJsonFormsLayoutProps,
   withTranslateProps,
 } from '@jsonforms/react';
 import { CategorizationList } from './CategorizationList';
 import { SingleCategory } from './SingleCategory';
-import { isCategorization } from './tester';
 import { withVanillaControlProps } from '../../util';
-import type { VanillaRendererProps } from '../../index';
+import type { AjvProps, VanillaRendererProps } from '../../index';
```
sdirix, line 37: We should not import from a parent index

TheZoker: I think the issue is, that the types are defined in that index file.
e.g.: https://github.com/eclipsesource/jsonforms/blob/master/packages/vanilla-renderers/src/index.ts#L86-L103

sdirix: That's not good. I would prefer if the `index` files only export and do nothing on their own. Can you refactor this and move the types and renderers each to an own file?

TheZoker: Done

packages/vanilla-renderers/src/complex/categorization/CategorizationList.tsx
```diff
@@ -37,7 +36,7 @@ const getCategoryClassName = (
 ): string => (selectedCategory === category ? 'selected' : '');
 
 export interface CategorizationProps {
-  categorization: Categorization;
+  filteredCategories: [Category];
```
lucas-koehler, line 39: I think this type is incorrect: `[Category]` is an array with exactly one `Category` element. `Category[]` is a normal array of `Category` elements.

Furthermore, I think handing in only the filtered categories is not enough. Child categorizations are lost with this approach. I think the type should probably be `(Category | Categorization)[]` to allow both. In fact it effectively gets an array of this type at the moment because `Categorization`s are not filtered out in `CategorizationRenderer` ;)

packages/vanilla-renderers/src/complex/categorization/CategorizationList.tsx
```diff
@@ -57,22 +56,20 @@ export const CategorizationList = ({
...
-      {categorization.elements.map((category, idx) => {
+      {filteredCategories.map((category, idx) => {
         if (isCategorization(category)) {
           return (
             <li key={categoryLabels[idx]} className={groupClassName}>
               <span>{categoryLabels[idx]}</span>
               <CategorizationList
-                categorization={category}
                 selectedCategory={selectedCategory}
+                filteredCategories={filteredCategories}
```
lucas-koehler, line 72: This is causing the endless loop and, thus, the Node out of memory error in the tests.  Basically you hand in the same categories as were handed in the current `CategorizationList`. This results in an endless loop. Instead, you need the elements contained in `category`

packages/vanilla-renderers/src/complex/categorization/CategorizationRenderer.tsx
```diff
@@ -22,88 +22,105 @@
...
+  t,
+  visible,
+  getStyleAsClassName,
+  onChange,
+}: LayoutProps &
+  VanillaRendererProps &
+  TranslateProps &
+  CategorizationProps &
+  AjvProps) => {
+  const ajv = useJsonForms().core.ajv;
```
lucas-koehler, line 63: You should not need this because `ajv` is already given as part of the props as indicated by the `AjvProps` type. To actually get this handed in, make sure to wrap the export at the end of this file with `withAjvProps`

packages/vanilla-renderers/src/complex/categorization/CategorizationRenderer.tsx
```diff
@@ -22,88 +22,105 @@
...
+  const categorization = uischema as Categorization;
+  const filteredCategories = useMemo(
+    () =>
+      categorization.elements.filter((category: Category) =>
+        isVisible(category, data, undefined, ajv)
+      ),
+    [categorization, data, ajv]
+  ) as [Category];
```
lucas-koehler, line 71: This is not what the useMemo returns because `[Category]` is an array (or tuple) with exactly one `Category` element.

packages/vanilla-renderers/src/complex/categorization/CategorizationRenderer.tsx
```diff
@@ -22,88 +22,105 @@
...
+}: LayoutProps &
+  VanillaRendererProps &
+  TranslateProps &
+  CategorizationProps &
+  AjvProps) => {
+  const ajv = useJsonForms().core.ajv;
+  const categorization = uischema as Categorization;
+  const filteredCategories = useMemo(
+    () =>
+      categorization.elements.filter((category: Category) =>
```
lucas-koehler, line 67: The type used in the method argument is not correct because `categorization.elements` is of type `(Category | Categorization)[]`. IF you want to get rid of the sub-categorizations, they need to be filtered out here. However, I think the `CategorizationList` rendered below also needs to know of child `Catagorization` elements. Otherwise nested categorizations can no longer be rendered. Or am I missing something?

lucas-koehler (changes requested):
Hi @TheZoker , thanks for the contribution.
I left some comments inline. Please have a look :) The two main issues atm seem to be:

- In `CategorizationList`, and endless loop is created by handing down the same `filteredCategories`. This causes the out of memory error during tests. More details are in the comment inline
- The new `filteredCategories` prop is of the wrong type and does not consider `Categorization` elements despite them effectively being handed down. Handing them down seems to be correct because otherwise we could no longer handle nested `Categorization`s. The type and potentially the prop name should be fixed accordingly.

packages/vanilla-renderers/src/complex/categorization/CategorizationRenderer.tsx
```diff
@@ -22,89 +22,105 @@
...
+  getStyleAsClassName,
+  onChange,
+  ajv,
+}: LayoutProps &
+  VanillaRendererProps &
+  TranslateProps &
+  CategorizationProps &
+  AjvProps) => {
+  const categorization = uischema as Categorization;
+  const elements = useMemo(
```
lucas-koehler, line 63: This memo is superfluous because `categorization.elements` does not change. Thus, the `useMemo` should be removed

packages/vanilla-renderers/src/complex/categorization/CategorizationList.tsx
```diff
@@ -24,55 +24,65 @@
...
+  data,
   depth,
   onSelect,
   subcategoriesClassName,
   groupClassName,
   t,
-}: CategorizationProps) => {
+  ajv,
+}: CategorizationProps & AjvProps) => {
+  const filteredElements = elements.filter(
```
lucas-koehler, line 63: This should be memoed via a `useMemo`. Otherwise, the memo for the `categoryLabels` is useless because `filteredElements` is recalculated on each render and, thus, breaks the memo of `categoryLabels`.

lucas-koehler (changes requested):
Thanks for the updates. Looks better now :) I have two remarks inline.
Furthermore, the vanilla example app and some tests for the input control and the categorization renderer are failing.
I have a suspicion regarding the example app and input control: `packages/vanilla-renderers/src/util/renderer.ts` being in the `util` folder creates a circular dependency. Thus, `renderer.ts` should be moved up one folder to be directly contained in the `src` folder.
Please also have a look at the categorization renderer tests.

lucas-koehler:
@TheZoker Thanks for the updates :)
The test case Categorization renderer › reset selected category on schema change is still failing. Please have a look.

packages/vanilla-renderers/src/renderer.ts
sdirix, line 1: This file should be called `renderers` as we typically use plurals.

packages/vanilla-renderers/src/renderer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+import {
+  GroupLayout,
+  groupTester,
+  HorizontalLayout,
+  horizontalLayoutTester,
+  VerticalLayout,
+  verticalLayoutTester,
+} from './layouts';
+import DateTimeCell from './cells/DateTimeCell';
```
sdirix, line 78: Can you align this import with the others?

sdirix (commented):
Looks good! Can you squash this into 1-2 nicely formatted commits?

sdirix (approved):

PR #2125, opened 2023-04-05
Fix reference type in schema and support angular 15

aregulator:
* if a json schema element has a link to reference type and `title` or `description`, these
elements overwrite `title` and `description`
from reference type.
* jsonforms supports angular 15, in order we can safe upgrade angular in frontend.

[REG-35516](https://jira.regnology.net/browse/REG-35516) [REG-35573](https://jira.regnology.net/browse/REG-35573)


aregulator:
wrong target branch

PR #2129, opened 2023-04-11
array translatins for vanilla, angular-material and vue

### LINKED ISSUE #1826: Translate hard coded strings in existing renderer sets

The goal is to allow a full translation of currently hard coded strings in the existing renderer sets.

With https://github.com/eclipsesource/jsonforms/issues/1751 JSON Forms now has the capabilities to localize any UI element. The new translation functions now need to be integrated in the existing React Material, React Vanilla, Angular Material, Vue 2/3 Vanilla and Vue 2 Vuetify renderer sets.

The alternative is for users to provide custom renderers.

### Framework

React, Angular, Vue 3, Vue 2

### RendererSet

### Additional context

Vuetify

LukasBoll:
Add translation-options for array-control elements for the vanilla, angular-material and vue renderers.

closes #1826

### DIFF
--- a/packages/angular-material/src/layouts/array-layout.renderer.ts
+++ b/packages/angular-material/src/layouts/array-layout.renderer.ts
@@ -31,6 +31,7 @@ import {
 import { JsonFormsAngularService, JsonFormsAbstractControl } from '@jsonforms/angular';
 import {
   ArrayLayoutProps,
+  ArrayTranslations,
   createDefaultValue,
   findUISchema,
   isObjectArrayWithNesting,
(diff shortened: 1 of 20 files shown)


sdirix:
@lucas-koehler Do you have any idea why the Angular Material tests are failing

LukasBoll:
Thank you @lucas-koehler, @sdirix,
I looked at the tests again this afternoon and I just found the issues; A module-import was missing in the testing Module! And the aria-label attribute needs to be assigned another way, then I did. I will update the PR today!

LukasBoll:
> Code changes in the renderers look already pretty good to me, thanks :)
> 
> There are various test cases failing. They are failing consistently on both Windows and Ubuntu. Note that the MacOS build succeeds because it does not execute tests.
> 
> It seems that the tests fail because elements are not found. This is likely caused by the changes to the aria labels in `packages/core/src/i18n/arrayTranslations.ts` because the tests use aria label attributes to find the buttons. Thus, you need to adapt the tests to the changed aria label texts.

Thank you, I just realized that as well, but now everything should work properly!

lucas-koehler (approved):
LGTM now, thanks for the fixes :)

PR #2130, opened 2023-04-12
WIP: Angular 15 and more

JonasDev17:
I updated lots of packages which will require some testing. I updated angular to angular 14.2 & 15, dropping support for 12 & 13 & 14.0.0, Typescript to 4.8.4.

I couldn't get the Vue vanilla build to pass because I never used Vue.

I am not sure if what I did here is any good, so I am open to feedback!


JonasDev17:
Not sure why these checks fail with a dependency issue as I have no dependency issues locally.

JonasDev17:
@sdirix Before I put in much more time, is this something you guys are interested in? My changes probably drop support for quite a few versions of the frameworks just like with angular where typescript >= 4.8.4 is required for angular 15, hence dropping support for 12, 13, and 14.0.0
If you guys don't wanna update yet or drop support for older versions then there is no point in me putting in work for all these other framworks and for the dev environment with lerna and typedoc and all that and I'll just create an angular-only fork because to be honest, that's all I need for my project...

sdirix:
Hi @JonasDev17! Thanks for your efforts!

If this is properly finished then we would definitely merge and use it going forward. We try to stay up to date, so if Angular 12/13/14 are then out of the question then that's fine. Users of these Angular versions can then just use an older version of JSON Forms.

My feeling is that this PR exploded a bit in scope. We would be perfectly fine just with the Angular update which is also easier to review than a huge PR touching rollup, typedoc, eslint, Material UI, even rimraf, etc. etc.

My suggestion would be to dial this a bit back and focus on the Angular packages. I understand that a Typescript update is needed for them which seemed to motivate all the other changes too. If you polish the PR so that `@jsonforms/core`, `@jsonforms/angular`, `@jsonforms/angular-material` and `@jsonforms/angular-test` work (you can ignore the other packages for now by removing them from lerna), then we can have a look at the remaining open issues, i.e. updating the remaining packages and maybe typedoc.

JonasDev17:
@sdirix 

Thanks for the feedback. I removed the other packages and finally got it to build but the examples page is empty - not sure what the problem is here.
I will probably have to create a new branch and open a new PR to clean up the PR.
But as far as I know, typedoc has to be updated as well because its part of every package and the lerna build process, no?

PR #2133, opened 2023-04-24
Do not filter out errors for oneOf enums

AlexandraBuzila:
The `errorsAt` function, used to filter the errors that belong to a control, ignores the errors AJV produces for oneOf schemas. Most of these errors should not be shown in the UI (e.g. when the oneOf property
does not match exactly one subschema), however the errors produced directly for oneOf enums are an exception.

Validation errors for oneOf enum controls are valid and should be shown in the UI.

### How to test in the UI:

- the `enums` example now displays sample controls with errors - [preview](https://deploy-preview-2133--jsonforms-examples.netlify.app/react-material/#enum)
- the `arrays` example now displays an item with a oneOf enum with errors - [preview](https://deploy-preview-2133--jsonforms-examples.netlify.app/react-material/#array)

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -45,7 +45,7 @@ import {
   UPDATE_CORE,
   UpdateCoreAction,
 } from '../actions';
-import { createAjv, Reducer } from '../util';
+import { createAjv, isOneOfEnumSchema, Reducer } from '../util';
 import type { JsonSchema, UISchemaElement } from '../models';
 
 export const validate = (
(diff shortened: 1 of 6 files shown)


eneufeld:
@AlexandraBuzila what am I missing, https://deploy-preview-2133--jsonforms-examples.netlify.app/react-material/#enum does not show any errors.
Could you please also improve the commit message so that it states, what is broken atm and how it is fixed?

AlexandraBuzila:
> @AlexandraBuzila what am I missing, https://deploy-preview-2133--jsonforms-examples.netlify.app/react-material/#enum does not show any errors. Could you please also improve the commit message so that it states, what is broken atm and how it is fixed?

Hi @eneufeld 
I did not adapt the `enum` example. I don't think it the error cases belong necessarily to that example.
To test the changes you need to follow the steps from the PR description, locally.

I'll improve the commit message.

eneufeld:
why do you think this case does not belong into the example?

sdirix:
I agree with Eugen, would be nice to have this in the enum example

AlexandraBuzila:
I updated the example, it now contains sample controls for enums with errors as well.

I think I overlooked the array case however and the errors for oneOf enums in array elements are  still filtered out. I'll update the PR ASAP.

AlexandraBuzila:
@eneufeld I fixed the code, errors for oneOf enums in arrays are now also considered.

eneufeld (approved):
Thank you!

Issue #2131, opened 2023-04-21
Fix ignored ESLint issues


lucas-koehler:
### Is your feature request related to a problem? Please describe.

This is a follow up of #1551 

The following ESLint violations were ignored in the initial PR #2112 but should be fixed, too:
[ ] `import/no-named-as-default`: Currently ignored in packages vanilla-renderers and material-renderers
[ ] `no-prototype-builtins`: Currently ignored in all packages
[ ] `@typescript-eslint/no-explicit-any`: Currently ignored in all packages
[ ] `@typescript-eslint/ban-types`: Search repository for `TODO fix @typescript-eslint/ban-types` to find places to fix. The rule was not universally ignored.

Potentially also investigate:
[ ] Ignored violation of `@angular-eslint/no-conflicting-lifecycle` in packages/angular/src/jsonforms-root.component.ts

### Describe the solution you'd like

-

### Describe alternatives you've considered

-

PR #2134, opened 2023-05-04
Enable ESLint no-prototype-builtins rule and fix errors

lucas-koehler:
Use `Object.prototype.<method>.call(<object>, <property>)` syntax instead of directly calling the methods on the objects. This is the suggested fix in the [rule documentation](https://eslint.org/docs/latest/rules/no-prototype-builtins).

Part of #2131


sdirix (approved):
LGTM

PR #2147, opened 2023-06-13
fix: properly match AJV paths with slashes

### LINKED ISSUE #2146: Errors for properties containing slashes are not properly matched
### Describe the bug

Consider the following subschema:

```json
"/work/email": {
  "title": "Email with slash",
  "description": "Email address With Slash",
  "type": "string",
  "format": "email"
}
```

The `'email'` format error is not successfully matched to the property and therefore does not show up in the `errors` prop.

### Expected behavior

The error should be matched to the property.

1. Use the property posted above
2. Observe that the UI will not show the error

### Screenshots

_No response_

n/a

3.0.0

### Framework

Core

### RendererSet

_No response_

### Additional context

Reported in the community: https://jsonforms.discourse.group/t/ajv-issues-with-slashes-in-json-pointer/1418

A good fix for the error could be in the [`getControlPath` method](https://github.com/eclipsesource/jsonforms/blob/a5fdca6fb6afa9820eabe8051401785675f53ad7/packages/core/src/reducers/core.ts#L358-L378). The method is used to transform the JSON Pointer based path from AJV into the dot-notation of JSON Forms. However it does not decode escaped JSON Pointer sequences which is why the matching fails: `'~1work~1email' === '/work/email'`

sdirix:
AJV error paths are JSON Pointers. These paths are converted to dot-notation in JSON Forms. However escaped JSON Pointer characters were not decoded. Therefore errors for properties containing one of the escaped characters were not properly matched.

Background information: In JSON Pointers two characters need to be escaped: `'~'` is converted to `'~0'` and `'/'` is converted to `'~1'`. Therefore properties containing either `'~'` or `'/'` were not properly matched to errors before this fix.

Fixes #2146

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -45,7 +45,7 @@ import {
   UPDATE_CORE,
   UpdateCoreAction,
 } from '../actions';
-import { createAjv, isOneOfEnumSchema, Reducer } from '../util';
+import { createAjv, decode, isOneOfEnumSchema, Reducer } from '../util';
 import type { JsonSchema, UISchemaElement } from '../models';
 
 export const validate = (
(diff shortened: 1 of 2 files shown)


LukasBoll (approved):
I tested the changes and they look good to me.

PR #2148, opened 2023-06-13
Upgrade build stack to pnpm, Node 16, Lerna 6

lucas-koehler:
- Move to Node 16.13+
- Use PNPM 8.5+ as package manager instead of NPM
- Upgrade to Lerna 6
- Explicitly add all needed dependencies to packages as PNPM does not hoist  application dependencies
- Use explicit PNPM workspace references for jsonforms packages
- Increase various dependencies to reflect the actually needed versions
- Migrate to Jest 27
- Increase Typedoc version to ~0.21.9 to be compatible with Node 16
- Increase base lib version from ES6 to ES2016 
- Improve stability of prepare example apps script
- **BREAKING: Remove vue2** to avoid migration cost


lucas-koehler:
@sdirix The netlify build fails because node 16 and pnpm are not available there, yet.

lucas-koehler:
@sdirix I rebased the changes. Netlify still fails as it does not have pnpm and node 16 configured.

.devcontainer/devcontainer.json
```diff
@@ -20,7 +20,7 @@
     ]
   },
   // Use 'postCreateCommand' to run commands after the container is created.
-  "postCreateCommand": "npm ci && npm run init && npm run build",
+  "postCreateCommand": "pnpm i --frozen-lockfile && pnpm run build",
   // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
   "remoteUser": "node"
-}
+}
```
sdirix, line 26: Your editor should add an empty line at the end

.github/workflows/ci.yaml
```diff
@@ -16,29 +16,43 @@ jobs:
...
         if: matrix.os == 'ubuntu-latest'
-        run: npm run lint
+        run: pnpm run lint
+
       - name: Test
         if: matrix.os == 'windows-latest'
         run: |
-          npm run test
+          set NODE_OPTIONS=--max_old_space_size=4096
+          pnpm run test
```
sdirix, line 46: We should try running tests on Mac again.

.vscode/settings.json
```diff
@@ -45,5 +45,6 @@
     "**/docs": true,
     "**/node_modules": true,
     "**/lib": true
-  }
-}
+  },
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
```
sdirix, line 50: empty line

README.md
```diff
@@ -32,10 +32,10 @@ In addition, EclipseSource also offers [professional support](https://jsonforms.
 
 ### First time setup
 
-- Install [node.js](https://nodejs.org/) (only Node 14 and npm 6 is currently supported)
+- Install [node.js](https://nodejs.org/) (only Node v16.13+ < 17 is currently supported)
+- Install pnpm: <https://pnpm.io/installation>
```
sdirix, line 36: Should we specify which version of pnpm or are they pretty compatible?

package.json
```diff
@@ -1,11 +1,14 @@
...
+    "pnpm": ">=8.5"
+  },
   "scripts": {
     "lerna": "lerna",
-    "preparePublish": "git clean -dfx && npm ci && npm run init && npm run clean && npm run build && npm run test",
+    "preparePublish": "git clean -dfx && pnpm i --frozen-lockfile && pnpm run clean && pnpm run build && pnpm run test",
     "merge-report": "mkdir -p coverage && lcov-result-merger 'packages/**/coverage/lcov.info' 'coverage/lcov.info'",
     "check-format": "echo 'temporarily disabled'",
-    "init": "lerna bootstrap --hoist",
     "build": "lerna run build && lerna run doc",
```
sdirix, line 12: I would like to remove the doc step from the build step

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - 'packages/*'
+  - 'packages/vue/*'
```
sdirix, line 3: Not as part of this PR but a follow up should move the vue packages up again

tsconfig.base.json
```diff
@@ -17,5 +20,7 @@
     "esModuleInterop": true,
     "importHelpers": true
   },
-  "types": ["jest"]
-}
+  "types": [
+    "jest"
+  ]
+}
```
sdirix, line 26: empty line

sdirix (commented):
I had a first look and only very minor comments. As a next step I will actually test it.

pnpm-workspace.yaml
lucas-koehler: I created https://github.com/eclipsesource/jsonforms/issues/2158 as a follow up so it doesn't get lost :)

README.md
lucas-koehler: It's hard to say but I'll add a minimum version. This is also specified in the package.json's `engines` prop and will fail when the version constrained is violated.

sdirix:
Seems the mac tests still fail. So let's exclude them again and create a follow up issue to fix them

lucas-koehler:
@sdirix I applied your suggestions and rebased the changes. The latest 2 commits are new.
I also tested running tests on macOS CI but unfortunately they still fail: https://github.com/eclipsesource/jsonforms/actions/runs/5517905609/jobs/10061162445?pr=2148

lucas-koehler:
Follow up issue to fix the macOS tests: https://github.com/eclipsesource/jsonforms/issues/2159

sdirix (approved):
Works great for me!

PR #2151, opened 2023-06-16
Add Object and OneOf renderer to vue3-vanilla

butzist:
We ported the Object and OneOf renderers from vue2 to vue3 - would be nice to have them upstream

### DIFF
--- a/packages/vue/vue-vanilla/src/complex/ObjectRenderer.vue
+++ b/packages/vue/vue-vanilla/src/complex/ObjectRenderer.vue
@@ -1,0 +1,84 @@
+<template>
+  <div v-if="control.visible">
+    <dispatch-renderer
+      :visible="control.visible"
+      :enabled="control.enabled"
+      :schema="control.schema"
+      :uischema="detailUiSchema"
+      :path="control.path"
+      :renderers="control.renderers"
+      :cells="control.cells"
+    />
+  </div>
+</template>
+
+<script lang="ts">
+import {
+  JsonFormsRendererRegistryEntry,
+  rankWith,
+  ControlElement,
+  Generate,
+  GroupLayout,
+  UISchemaElement,
+  findUISchema,
+  isObjectControl,
+} from '@jsonforms/core';
+import { defineComponent } from 'vue';
+import {
+  DispatchRenderer,
+  rendererProps,
+  RendererProps,
+  useJsonFormsControlWithDetail,
+} from '../../config/jsonforms';
+import { useVanillaControl } from '../util';
+import { isEmpty } from 'lodash';
+
+const controlRenderer = defineComponent({
+  name: 'ObjectRenderer',
+  components: {
+    DispatchRenderer,
+  },
+  props: {
+    ...rendererProps<ControlElement>(),
+  },
+  setup(props: RendererProps<ControlElement>) {
+    const control = useVanillaControl(useJsonFormsControlWithDetail(props));
+    return {
+      ...control,
+      input: control,
+    };
+  },
+  computed: {
+    detailUiSchema(): UISchemaElement {
+      const uiSchemaGenerator = () => {
+        const uiSchema = Generate.uiSchema(this.control.schema, 'Group');
+        if (isEmpty(this.control.path)) {
+          uiSchema.type = 'VerticalLayout';
+        } else {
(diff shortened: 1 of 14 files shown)


sdirix:

In general we are definitely interested in these renderers :+1: . Please make sure that the build succeeds (there seem to be some Typescript errors) and sign the CLA. Afterwards we'll take a look. Thanks!

butzist:
Should I try to copy over the relevant unit tests from vue2 - or what is required to get the check passing?

Also I probably missed in the docs how to run the build/test for a specific package. Can you help me there?

butzist:
Update: added some tests for the new renderers, but was not able to test the dialog in the OneOfRenderer (showModal() is not a function). It worked in manual testing though. Let me know if there is anything missing.

packages/vue/vue-vanilla/src/complex/ObjectRenderer.vue
```diff
@@ -0,0 +1,96 @@
...
+
+      if (this.nested.level > 0) {
+        result = cloneDeep(result);
+        result.options = {
+          ...result.options,
+          bare: true,
+          alignLeft:
+            this.nested.level >= 4 || this.nested.parentElement === 'array',
+        };
+      }
```
sdirix, line 83: Do you actually use this information in your renderers? This was introduced in the Vue Vuetify renderer set because of its specifics, I don't think we necessarily need the `useNested` in the Vanilla renderers. I would have expected a simple implementation [like in the React Material renderers](https://github.com/eclipsesource/jsonforms/blob/7e0115feeced7711b0768d325566aaa6b054c32b/packages/material-renderers/src/complex/MaterialObjectRenderer.tsx).

packages/vue/vue-vanilla/src/complex/OneOfRenderer.vue
```diff
@@ -0,0 +1,203 @@
...
+      :uischema="indexedOneOfRenderInfos[selectedIndex].uischema"
+      :path="control.path"
+      :renderers="control.renderers"
+      :cells="control.cells"
+      :enabled="control.enabled"
+    />
+
+    <dialog ref="dialog" :class="styles.dialog.root">
+      <h1 :class="styles.dialog.title">
+        {{ t('form.clear.title', 'Clear form?') }}
```
sdirix, line 47: These translations don't work as the rest of them do in the code base. We don't use a hard coded key but use a prefix for each control.

See [`mapStateToArrayControlProps`](https://github.com/eclipsesource/jsonforms/blob/7e0115feeced7711b0768d325566aaa6b054c32b/packages/core/src/util/renderer.ts#L755-L760) for an example in the core library.

We don't provide the one of specific translations in the core framework yet. For now I would like to suggest to at least extract that into an own `useOneOfTranslations` here and then access the translations via a `translations` object. The keys should be determined similar to the core library.

sdirix (commented):
Looks good in general! Can you take a look at my comments?

packages/vue/vue-vanilla/src/complex/OneOfRenderer.vue
butzist: Added combinatorTranslations to core

packages/vue/vue-vanilla/src/complex/ObjectRenderer.vue
butzist: removed

butzist:
Now also the build is passing for me - sorry for the inconvenience, but had some trouble getting the build system to work locally

LukasBoll (commented):
Thank you for the great contribution @butzist!
Everything looks good to me. I just found one issue, which I addressed in my comment below.

packages/vue/vue-vanilla/src/complex/OneOfRenderer.vue
```diff
@@ -0,0 +1,196 @@
...
+      this.newSelection();
+      this.closeDialog();
+    },
+    onCancel(): void {
+      this.newSelectedIndex = this.selectedIndex;
+      this.closeDialog();
+    },
+    newSelection(): void {
+      this.handleChange(
+        this.path,
```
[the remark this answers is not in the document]
butzist: Danke @LukasBoll , I pushed the updated code

LukasBoll (approved):
Looks good to me!

PR #2154, opened 2023-07-04
Vue: Disable ArrayListRenderer add if control.data.length exceeds or is equal to maxItems

Droxx:
The included ArrayListRenderer in Vue-Vanilla would allow more items to be added to an array than the schema allowed.

This PR uses the maxItems value from the schema (if present) and disables the add button if the control.data array is an equal, or greater, length to the maxItems value in the schema


packages/vue/vue-vanilla/src/array/ArrayListRenderer.vue
```diff
@@ -5,7 +5,12 @@
         :class="styles.arrayList.addButton"
         type="button"
         @click="addButtonClick"
-      >
+        :disabled="!control.enabled ||
+            (arraySchema !== undefined &&
+            arraySchema.maxItems !== undefined &&
+            control.data !== undefined &&
+            control.data.length >= arraySchema.maxItems)"
+        >
```
sdirix, line 13: I would prefer moving the calculation of this into a `computed`. Also we should only disable the add button if `appliedOptions.restrict` is `true`. By default we don't interfere with the user doing something which invalidates the schema.
The same could be implemented for the delete button and checking the `minItems` if you want to.

packages/vue/vue-vanilla/src/array/ArrayListRenderer.vue
```diff
@@ -77,6 +84,12 @@ const controlRenderer = defineComponent({
     noData(): boolean {
       return !this.control.data || this.control.data.length === 0;
     },
+    arraySchema(): JsonSchema | undefined {
+      return Resolve.schema(
+          this.control.rootSchema,
```
sdirix, line 89: The first parameter should be the `props.schema` otherwise this will fail for nested objects/arrays.

sdirix (changes requested):
Thanks for the contribution :heart:. Much appreciated.

Please check the comments and add tests for the new use case(s). Also the build must succeed, it seems the linter is complaining.

packages/vue/vue-vanilla/src/array/ArrayListRenderer.vue
Droxx: Thanks @sdirix I've addressed this, plus your other feedback. But I had two comments:

* I did initially add appliedOptions.restrict. But I removed it due to the following
   * I cannot see this check being performed on any other renderers in the project (I do see it in use in the vuetify renderers). So I left it out in order to remain consistent with the other renderers
   * I cannot find documentation on how to set this option when using json-forms in vue. If you could point me in this direction, it would be helpful!
* Secondly, I'm afraid I didn't implement a minItems delete button check. In my use-case (vanilla-vue) I cannot see the delete button render. So perhaps this need to first be addressed as a separate issue. I'll post some screenshots below and maybe you can determine if this is due to my usage, or an issue in the package itself

Droxx:
Thanks for the feedback @sdirix. I have also implemented minItems checking on the delete button.

I've left one comment on your review. But besides this it should be ready for re-evaluation

sdirix:
> I cannot see this check being performed on any other renderers in the project (I do see it in use in the vuetify renderers). So I left it out in order to remain consistent with the other renderers

Besides the Vuetify renderers ([ListWithDetail](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/additional/ListWithDetailRenderer.vue#L28), [ArrayControlRenderer](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/complex/ArrayControlRenderer.vue#L26), [AdditionalProperties](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/complex/components/AdditionalProperties.vue#L322), [AnyOfStringOrEnumControlRenderer](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/controls/AnyOfStringOrEnumControlRenderer.vue#L23), [MultiStringControlRenderer](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/controls/MultiStringControlRenderer.vue#L23), [PasswordControlRenderer](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/controls/PasswordControlRenderer.vue#L24), [StringControlRenderer](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/controls/StringControlRenderer.vue#L23), [StringMaskControlRenderer](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/controls/StringMaskControlRenderer.vue#L22), [ArrayLayoutRenderer](https://github.com/eclipsesource/jsonforms-vuetify-renderers/blob/1ca0fc47848e49b714887abb6c4de622c96f6369/vue2-vuetify/src/layouts/ArrayLayoutRenderer.vue#L159)), restrict is used in various string and number renderers, see [MaterialAnyOfStringOrEnumControl](https://github.com/eclipsesource/jsonforms/blob/f74e9ed35edc970dbfc601e362e730bb10ab042b/packages/material-renderers/src/controls/MaterialAnyOfStringOrEnumControl.tsx#L73), [MuiInputNumberFormat](https://github.com/eclipsesource/jsonforms/blob/f74e9ed35edc970dbfc601e362e730bb10ab042b/packages/material-renderers/src/mui-controls/MuiInputNumberFormat.tsx#L48), [MuiInputText](https://github.com/eclipsesource/jsonforms/blob/f74e9ed35edc970dbfc601e362e730bb10ab042b/packages/material-renderers/src/mui-controls/MuiInputText.tsx#L68), [NumberFormatCell](https://github.com/eclipsesource/jsonforms/blob/f74e9ed35edc970dbfc601e362e730bb10ab042b/packages/vanilla-renderers/src/cells/NumberFormatCell.tsx#L60), [TextCell](https://github.com/eclipsesource/jsonforms/blob/f74e9ed35edc970dbfc601e362e730bb10ab042b/packages/vanilla-renderers/src/cells/TextCell.tsx#L63).

> I cannot find documentation on how to set this option when using json-forms in vue. If you could point me in this direction, it would be helpful!

You can set it in each Ui Schema options, e.g.

```
{
  type: 'Control',
  scope: '#/properties/name',
  options: {
    restrict: true
  }
}
```

Or globally via the config

```
<json-forms config="config"/>

setup(){
  return ({
    config: { restrict: true}
  });
}
```

packages/vue/vue-vanilla/src/array/ArrayListRenderer.vue
```diff
@@ -4,6 +4,7 @@
       <button
         :class="styles.arrayList.addButton"
         type="button"
+        :disabled="!control.enabled || maxItemsReached"
```
sdirix, line 7: good catch to also check `control.enabled`. Can you also add the check to all the other buttons?

package-lock.json
sdirix, line 1: Please don't commit any package-lock changes

Droxx:
@sdirix Thanks for the comments about the appliedOptions.restrict. I have added this to the control. 

I have also addressed your latest points

sdirix (approved):

Issue #1790, opened 2021-07-27
Difference in mapAdditionalProps props.data and global data when setting a prop as undefined via handleChange


JBBianchi:
### Describe the bug

When implementing `DispatchPropsOfControl.handleChange()` in a custom control and calling `handleChange('somePath', undefined)`, the property at `somePath` of the global `data` object is removed but in `mapAdditionalProps(props)`, the property still exists with an `undefined` value in `props.data`.

This is useful when handling `additionalProperties` for instance.

### Expected behavior

For the property to be remove of `props.data`

### Steps to reproduce the issue

https://github.com/JBBianchi/jsonforms-angular-seed/blob/master/src/app/edit-object.renderer.ts#L105

1. Clone & run the above project
2. Test and debug the "object editor" sample (not the "dictionary" sample)

### In which browser are you experiencing the issue?

any

### Framework

Angular

### RendererSet

Material

PR #2163, opened 2023-07-12
feat: remove state slice when set as undefined

JBBianchi:
Handles setting a slice state to `undefined` as removing the said slice.

Related to #1790

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -25,6 +25,7 @@
 
 import cloneDeep from 'lodash/cloneDeep';
 import setFp from 'lodash/fp/set';
+import unsetFp from 'lodash/fp/unset';
 import get from 'lodash/get';
 import filter from 'lodash/filter';
 import isEqual from 'lodash/isEqual';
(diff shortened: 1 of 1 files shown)


sdirix (changes requested):

Please add test cases for the new behavior and make sure that the builds are green.

JBBianchi:
Sorry, I did hammer a bit the CI, I don't have the environment installed locally so I could test before committing.

I hope it's good enough.

packages/core/test/reducers/core.test.ts
```diff
@@ -562,6 +562,44 @@ test('core reducer - update - should update errors', (t) => {
...
+  const after = coreReducer(
+    before,
+    update('foo', (_) => {
+      return undefined;
+    })
+  );
+
+  t.not(before, after);
+  t.not(before.data, after.data);
+  t.like(Object.keys(after.data), ['fizz']);
```
lucas-koehler, line 600: `t.like` must not be used. It leads to an out of memory exception of NodeJS. Interestingly enough, in VSCode the `like` method is not even recognized at all. Why this leads to out of memory and not just fails the test execution, I don't know.
To fix this, you can use `t.deepEqual` to verify the `after` object has the correct keys. See the suggested fix below. 
```suggestion
  t.deepEqual(Object.keys(after.data), ['fizz']);
```

JBBianchi: Thanks for your input, included in [041db92](https://github.com/eclipsesource/jsonforms/pull/2163/commits/041db924a68ea824abe81aa265e5b2fe3a97869b)

JBBianchi: All green, thanks!

sdirix:
We can do an alpha release soon, so you can consume the changes :+1:

JBBianchi:
> We can do an alpha release soon, so you can consume the changes 👍

That'd be great :)

sdirix:
@JBBianchi `3.1.1-alpha.0` is released

JBBianchi:
This feature seems to work as expected but this alpha-0 ships the bug reported (and fixed) in #2166 so I'll wait next release to actually use it.

PR #2165, opened 2023-07-18
Add a test for combinators and pass resolved subschema to createLabel

### LINKED ISSUE #2164: AnyOf subschemas defined in refs don't get correct labels from createCombinatorRenderInfos
### Describe the bug

Consider this schema:
```
{
  type: 'object',
  properties: {
    widget: {
      anyOf: [
        {
          $ref: '#/definitions/Dua',
        },
        {
          $ref: '#/definitions/Lipa',
        },
      ],
    },
  },
  definitions: {
    Dua: {
      title: 'Dua',
      type: 'object',
      properties: { name: { type: 'string' } },
    },
    Lipa: {
      title: 'Lipa',
      type: 'object',
      properties: { name: { type: 'string' } },
    },
  },
};
```

When calling createCombinatorRenderInfos, the title of the subschemas is not reflected in the label for individual renderinfos.

### Expected behavior

The label for individual renderinfos is informed by the title of the subschema.

I've got a test case in a PR I'm about to open, which fails without the change I'm introducing.

### Screenshots

_No response_

Node

v3.1.0

### Framework

Core

### RendererSet

### Additional context

_No response_

DrewHoo:
Resolves #2164 
- Passes resolved subschema to createLabel so that label can actually use the title of the subschema
- Adds a test for combinators that fails if the resolved schema isn't passed

### DIFF
--- a/packages/core/src/util/combinators.ts
+++ b/packages/core/src/util/combinators.ts
@@ -70,6 +70,6 @@ export const createCombinatorRenderInfos = (
         control,
         rootSchema
       ),
-      label: createLabel(subSchema, subSchemaIndex, keyword),
+      label: createLabel(schema, subSchemaIndex, keyword),
     };
   });
--- a/packages/core/src/util/combinators.ts
+++ b/packages/core/src/util/combinators.ts
@@ -23,7 +23,6 @@
   THE SOFTWARE.
 */
 
-import merge from 'lodash/merge';
 import type { ControlElement, JsonSchema, UISchemaElement } from '../models';
 import { findUISchema, JsonFormsUISchemaRegistryEntry } from '../reducers';
 import { Resolve } from './util';
--- a/packages/core/src/util/combinators.ts
+++ b/packages/core/src/util/combinators.ts
@@ -23,6 +23,7 @@
   THE SOFTWARE.
 */
 
+import merge from 'lodash/merge';
 import type { ControlElement, JsonSchema, UISchemaElement } from '../models';
 import { findUISchema, JsonFormsUISchemaRegistryEntry } from '../reducers';
 import { Resolve } from './util';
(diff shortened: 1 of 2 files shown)


DrewHoo:
Hi @sdirix do I need to do anything additional to get this PR reviewed?

sdirix (commented):
Hi @DrewHoo,

Thanks for the contribution!

While this fixes the use case of the test case, another use case is now broken:

```
{
  type: 'object',
  properties: {
    widget: {
      anyOf: [
        {
          title: 'DuaOverride',
          $ref: '#/definitions/Dua',
        },
        {
          title: 'LipaOverride',
          $ref: '#/definitions/Lipa',
        },
      ],
    },
  },
  definitions: {
    Dua: {
      title: 'Dua',
      type: 'object',
      properties: { name: { type: 'string' } },
    },
    Lipa: {
      title: 'Lipa',
      type: 'object',
      properties: { name: { type: 'string' } },
    },
  },
};
```
`DuaOverride` and `LipaOverride` are no longer considered.

We should check both `title` and use the outer one in case both exist.

DrewHoo:
@sdirix Thanks for the suggestion! I added the test case you described and merged the two schemas before passing them to `createLabel`, giving the outer properties preference -- not sure if merged schema would also be appropriate for other places the sub-schema is used within the combinators function, so I only modified what's passed to `createLabel`.

packages/core/src/util/combinators.ts
```diff
@@ -70,6 +72,10 @@ export const createCombinatorRenderInfos = (
         control,
         rootSchema
       ),
-      label: createLabel(subSchema, subSchemaIndex, keyword),
+      label: createLabel(
+        merge({}, resolvedSubSchema, subSchema),
+        subSchemaIndex,
+        keyword
+      ),
```
sdirix, line 79: I think the `merge` is a bit too much ;) I would like to suggest to be a bit more explicit here and just remove the unnecessary `createLabel` method.

```suggestion
      label: subSchema.title ?? resolvedSubSchema.title ?? `${keyword}-${subSchemaIndex}`
```

sdirix (commented):
looks good! one more small suggestion ;)

sdirix (approved):

sdirix:
@DrewHoo Thanks a lot for the contribution :heart:

PR #2169, opened 2023-08-09
fix createRenderInfo TypeError in default title case

DrewHoo:
I realized there's a possible error here, added a test & fix for it in last commit


DrewHoo:
@sdirix see test case here--I think there's a bug if there's neither a resolved schema or a title.

lucas-koehler (approved):
Hi @DrewHoo 
thank you very much for the contribution <3
The fix and the test case look good to me :)

PR #2171, opened 2023-08-11
ci: Enable and fix tests on MacOS

### LINKED ISSUE #2159: Fix tests on macOS and execute during CI

Tests currently don't work on MacOS and, thus, are not executed during CI.

The tests should be fixed and run during CI.

lucas-koehler:
For ava tests, register ts-node with adapted typescript config adaptions via a javascript file. Before, the typescript config adaptions were loaded via an environment variable not working on mac and Windows via one configuration.

With this, the `cross-env` dev dependency is obsolete in `core` and `angular` packages. This removes the deps and regenerates the lock file to update dependencies.

Fixes #2159


sdirix (approved):
Very nice!

PR #2175, opened 2023-09-11
reset data in anyOfRenderer, if data types dont match

### LINKED ISSUE #2173: invalid output from allOf/anyOf combinator
### Describe the bug

Here is my code
```

import { JsonForms } from "@jsonforms/react";
import { useState } from "react";
import {
  materialRenderers,
  materialCells,
} from "@jsonforms/material-renderers";

const Example = () => {
  const [data] = useState({});
  const schema = {
    anyOf: [
      {
        type: "object",
        properties: {
          firstName: {
            type: "object",
            properties: {
              A: {
                type: "number",
                maxLength: 1,
              },
            },
          },
        },
      },
      {
        type: "string",
      },
    ],
  };
  return (
    <JsonForms
      schema={schema}
      uischema={undefined}
      data={data}
      onChange={(d: any) => {
        console.log("getData ", d);
      }}
      renderers={materialRenderers}
      cells={materialCells}
    />
  );
};
```

I got this output: 

image

I cant paste by code because it is a String object. When I copy this object it turns to: 
`{
    "data": "1",
    "errors": []
}`.
Obviously it is not a valid json object.

### Expected behavior

I expected output: 
`
{
    "data": {
        "firstName": {
            "A": 1
        }
    },
    "errors": []
}
` 

1. Start this example and render:
image
2. select anyof-1 tab & input any string
image
3. select anyof-0 tab & input any string
image
4.  see error on console

### Screenshots

_No response_

Google Chrome  116.0.5845.96

v3.1.0

### Framework

React

### RendererSet

Material

### Additional context

_No response_

LukasBoll:
closes #2173

### DIFF
--- a/packages/material-renderers/src/complex/MaterialAnyOfRenderer.tsx
+++ b/packages/material-renderers/src/complex/MaterialAnyOfRenderer.tsx
@@ -25,18 +25,22 @@
 import React, { useCallback, useState } from 'react';
 
 import {
+  CombinatorRendererProps,
   createCombinatorRenderInfos,
+  createDefaultValue,
   isAnyOfControl,
   JsonSchema,
   RankedTester,
   rankWith,
-  StatePropsOfCombinator,
 } from '@jsonforms/core';
 import { JsonFormsDispatch, withJsonFormsAnyOfProps } from '@jsonforms/react';
 import { Hidden, Tab, Tabs } from '@mui/material';
 import CombinatorProperties from './CombinatorProperties';
+import isEmpty from 'lodash/isEmpty';
+import { TabSwitchConfirmDialog } from './TabSwitchConfirmDialog';
 
 export const MaterialAnyOfRenderer = ({
+  handleChange,
   schema,
   rootSchema,
   indexOfFittingSchema,
@@ -46,12 +50,44 @@ export const MaterialAnyOfRenderer = ({
   cells,
   uischema,
   uischemas,
-}: StatePropsOfCombinator) => {
+  id,
+  data,
+}: CombinatorRendererProps) => {
   const [selectedAnyOf, setSelectedAnyOf] = useState(indexOfFittingSchema || 0);
-  const handleChange = useCallback(
-    (_ev: any, value: number) => setSelectedAnyOf(value),
-    [setSelectedAnyOf]
+  const [confirmDialogOpen, setConfirmDialogOpen] = useState(false);
+  const [newSelectedIndex, setNewSelectedIndex] = useState(0);
+
+  const handleClose = useCallback(
+    () => setConfirmDialogOpen(false),
+    [setConfirmDialogOpen]
   );
+
+  const handleTabChange = useCallback(
+    (_event: any, newIndex: number) => {
+      if (
+        isEmpty(data) ||
+        typeof data ===
+          typeof createDefaultValue(anyOfRenderInfos[newIndex].schema)
+      ) {
+        setSelectedAnyOf(newIndex);
+      } else {
+        setNewSelectedIndex(newIndex);
+        setConfirmDialogOpen(true);
+      }
+    },
+    [setConfirmDialogOpen, setSelectedAnyOf, data]
+  );
+
+  const openNewTab = (newIndex: number) => {
+    handleChange(path, createDefaultValue(anyOfRenderInfos[newIndex].schema));
+    setSelectedAnyOf(newIndex);
+  };
+
+  const confirm = useCallback(() => {
+    openNewTab(newSelectedIndex);
+    setConfirmDialogOpen(false);
+  }, [handleChange, createDefaultValue, newSelectedIndex]);
+
   const anyOf = 'anyOf';
   const anyOfRenderInfos = createCombinatorRenderInfos(
     (schema as JsonSchema).anyOf,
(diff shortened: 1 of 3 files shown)


packages/material-renderers/src/complex/ConfirmDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={cancel} color='primary'>
+          No
+        </Button>
+        <Button
+          onClick={confirm}
+          color='primary'
+          autoFocus
+          id={`oneOf-${id}-confirm-yes`}
```
lucas-koehler, line 49: The `oneOf` part of the id is not universal and incorrect when used in the any of renderer. Instead, you could hand in an id prefix property that contains the oneOf/anyOf string together with the id prop.

packages/material-renderers/src/complex/ConfirmDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+  DialogContentText,
+  DialogTitle,
+} from '@mui/material';
+
+export interface ConfirmDialogProps {
+  open: boolean;
+  handleClose: () => void;
+  confirm: () => void;
+  cancel: () => void;
+  id: any;
```
lucas-koehler, line 17: I would expect this to be of type `string`. Is there a reason for using `any`?

packages/material-renderers/src/complex/ConfirmDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export const ConfirmDialog = ({
```
lucas-koehler, line 20: Could we name this in a way that immediately makes clear what kind of confirmation this dialog is for? E.g. `TabSwitchConfirmDialog`.

packages/material-renderers/src/complex/MaterialAnyOfRenderer.tsx
```diff
@@ -46,12 +50,41 @@ export const MaterialAnyOfRenderer = ({
...
+  const [newSelectedIndex, setNewSelectedIndex] = useState(0);
+
+  const handleClose = useCallback(() => setOpen(false), [setOpen]);
+
+  const handleTabChange = useCallback(
+    (_event: any, newIndex: number) => {
+      if (
+        isEmpty(data) ||
+        typeof data ===
+          typeof createDefaultValue(anyOfRenderInfos[newIndex].schema)
```
lucas-koehler, line 67: What is the purpose of this? Is this to avoid unnecessary confirmations in case the data is "the same"?
If yes, I see a risk of skipping the  confirmation if two anyOfs have the same data type. This could easily happen if they are both objects.
You can test this in the deployed example at https://deploy-preview-2175--jsonforms-examples.netlify.app/react-material/#anyOf
When you change some data and then switch the tab, no confirm dialog is opened.

packages/material-renderers/src/complex/MaterialAnyOfRenderer.tsx
```diff
@@ -46,12 +50,41 @@ export const MaterialAnyOfRenderer = ({
...
   uischemas,
-}: StatePropsOfCombinator) => {
+  id,
+  data,
+}: CombinatorRendererProps) => {
   const [selectedAnyOf, setSelectedAnyOf] = useState(indexOfFittingSchema || 0);
-  const handleChange = useCallback(
-    (_ev: any, value: number) => setSelectedAnyOf(value),
-    [setSelectedAnyOf]
+  const [open, setOpen] = useState(false);
```
lucas-koehler, line 57: Minor: This could be renamed to `confirmDialogOpen` or someting similar to immediately make clear what this is about :)

lucas-koehler (changes requested):
Hi @LukasBoll 
Thanks for the pull request :heart: 
I have some comments inline. Pease have a look.

packages/material-renderers/src/complex/MaterialAnyOfRenderer.tsx
LukasBoll: The purpose was to fix #2173, where unfitting data types would lead to a crash. For example, if you switch between an object and an array. I would argue that the current behaviour is correct. AnyOf means that several of the subschemas can be fulfilled. Resetting the data every time would represent the oneOf option.

LukasBoll:
Thank you very much for the review and your comments! I adressed all of your comments!

packages/material-renderers/src/complex/TabSwitchConfirmDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={cancel} color='primary'>
+          No
+        </Button>
+        <Button
+          onClick={confirm}
+          color='primary'
+          autoFocus
+          id={id + -'confirm-yes'}
```
lucas-koehler, line 49: The `+ -'confirm-yes'` looks suspicious to me. Testing something like this in the browser console `-'avc'` is converted to `NaN`. Instead you can simply to it together via a template string.
```suggestion
          id={`${id}-confirm-yes`}
```

packages/material-renderers/src/complex/MaterialAnyOfRenderer.tsx
lucas-koehler: That makes sense. Thanks for the explanation :)

lucas-koehler (changes requested):
Hi @LukasBoll , thanks for the updates. I found one small bug inline and added a suggestion to fix it.

lucas-koehler (approved):
@LukasBoll I merged the minor suggested fix. Thanks again for the contribution :)

PR #2181, opened 2023-09-29
core: fix removeItems for list renderers

### LINKED ISSUE #2180: List renderer HOC: removeItems does not handle indices above 10 properly
### Describe the bug

The `removeItems` function exposed by the array HOC does not work well with indexes above 10.

For instance, given a custom list renderer that holds 12 items, `props.removeItems('', [11, 2])()` won't give the expected result: only the item at index 2 will be removed.

The issue is with the usage of `sort` on an array of number in the implementation: `[1, 2, 10].sort() === [1, 10, 2]`.

### Expected behavior

The `removeItems` function exposed by the array HOC should work well with indexes above 10.

In this example, item at index 11 should have been removed.

Here is a test that reproduces the issue:

```js
test('mapDispatchToArrayControlProps should remove items from array with more than 10 items', (t) => {
  const data = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11'];
  const schema: JsonSchema = {
    type: 'array',
    items: {
      type: 'string',
    },
  };
  const uischema: ControlElement = {
    type: 'Control',
    scope: '#',
  };
  const initCore: JsonFormsCore = {
    uischema,
    schema,
    data,
    errors: [] as ErrorObject[],
  };
  const [getCore, dispatch] = mockDispatch(initCore);
  dispatch(init(data, schema, uischema));
  const props = mapDispatchToArrayControlProps(dispatch);
  props.removeItems('', [11, 2])();
  t.is(getCore().data.length, 10); // fails, actual value is 11
});
```

### Screenshots

_No response_

any

master

### Framework

core

### RendererSet

_No response_

### Additional context

Custom renderer

xou816:
Fixes #2180

### DIFF
--- a/packages/core/src/util/renderer.ts
+++ b/packages/core/src/util/renderer.ts
@@ -799,7 +799,7 @@ export const mapDispatchToArrayControlProps = (
     dispatch(
       update(path, (array) => {
         toDelete
-          .sort()
+          .sort((a, b) => a - b)
           .reverse()
           .forEach((s) => array.splice(s, 1));
         return array;
--- a/packages/core/test/util/renderer.test.ts
+++ b/packages/core/test/util/renderer.test.ts
@@ -625,6 +625,31 @@ test('mapDispatchToArrayControlProps should remove items from array', (t) => {
   t.is(getCore().data[0], 'quux');
 });
 
+test('mapDispatchToArrayControlProps should remove items from array with more than 10 items', (t) => {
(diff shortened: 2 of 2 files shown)


lucas-koehler (approved):
Hi @xou816 ,
thanks for contributing this fix and opening the corresponding issue :heart: 
The changes look good to me!

xou816:
You're welcome! :)

PR #2186, opened 2023-10-06
update  peer dependency for mui/icons in material-renderers

IhsenBen:
self explanatory minor change

### DIFF
--- a/packages/material-renderers/package.json
+++ b/packages/material-renderers/package.json
@@ -87,7 +87,7 @@
     "@emotion/styled": "^11.3.0",
     "@jsonforms/core": "3.1.1-alpha.0",
     "@jsonforms/react": "3.1.1-alpha.0",
-    "@mui/icons-material": "~5.11.16",
+    "@mui/icons-material": "^5.11.16 || ^6.0.0",
     "@mui/material": "~5.13.0",
     "@mui/x-date-pickers": "^6.0.0",
     "react": "^16.12.0 || ^17.0.0 || ^18.0.0"
--- a/packages/material-renderers/package.json
+++ b/packages/material-renderers/package.json
@@ -87,7 +87,7 @@
     "@emotion/styled": "^11.3.0",
     "@jsonforms/core": "3.1.1-alpha.0",
     "@jsonforms/react": "3.1.1-alpha.0",
-    "@mui/icons-material": "^5.11.16 || ^6.0.0",
+    "@mui/icons-material": "^5.11.16",
     "@mui/material": "~5.13.0",
     "@mui/x-date-pickers": "^6.0.0",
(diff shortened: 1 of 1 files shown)


packages/material-renderers/package.json
```diff
@@ -87,7 +87,7 @@
     "@emotion/styled": "^11.3.0",
     "@jsonforms/core": "3.1.1-alpha.0",
     "@jsonforms/react": "3.1.1-alpha.0",
-    "@mui/icons-material": "~5.11.16",
+    "@mui/icons-material": ">=5.11.16",
```
lucas-koehler, line 90: `>=` seems a bit too lax to me because it also allows all higher major versions (e.g. 6.x, 7.x). I think we should change it to `^5.11.16` to allow all `5.x` versions but no other major versions. If it's known to work with 6.x it could then be adapted to `^5.11.16 || ^6.0.0`

lucas-koehler (changes requested):
I added a comment inline. Please have a look.

packages/material-renderers/package.json
IhsenBen: will do, thx

IhsenBen: it's done :D

packages/material-renderers/package.json
```diff
@@ -87,7 +87,7 @@
     "@emotion/styled": "^11.3.0",
     "@jsonforms/core": "3.1.1-alpha.0",
     "@jsonforms/react": "3.1.1-alpha.0",
-    "@mui/icons-material": "~5.11.16",
+    "@mui/icons-material": "^5.11.16 || ^6.0.0",
```
lucas-koehler, line 90: I quickly looked it up now and saw that there is no version 6.x of [@mui/icons-material](https://www.npmjs.com/package/@mui/icons-material?activeTab=versions) at the moment. Thus, I would not include it because we don't know if it will break.

```suggestion
    "@mui/icons-material": "^5.11.16",
```

lucas-koehler (changes requested):

lucas-koehler (approved):
Hi @IhsenBen ,
I changed it to `^5.11.16` because v6 does not exist yet and we don't know whether it will be compatible. Thanks for the issue and the PR :)

lucas-koehler:
@IhsenBen could you sign the [contributor license agreement](https://github.com/eclipsesource/jsonforms/pull/2186#issuecomment-1750547684) please? Otherwise I cannot merge this. You won't need to do this again in case you open further PRs in the future :)

IhsenBen:
> @IhsenBen could you sign the [contributor license agreement](https://github.com/eclipsesource/jsonforms/pull/2186#issuecomment-1750547684) please? Otherwise I cannot merge this. You won't need to do this again in case you open further PRs in the future :)

done

PR #2189, opened 2023-10-10
generate a control for enums in generated uischema

### LINKED ISSUE #2177: Enum without "type" can't be rendered
### Describe the bug

Enum definition without "type", like:

```json
{
  "enum": ["A", "B"]
}
```

, is valid JSON Schema, but can't be rendered.

### Expected behavior

Rendered as normal enum.

The following schema does not render any fields:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "properties": {
    "foo": {
      "enum": ["A"]
    }
  }
}
```

image

And the following schema causes runtime error, `options is undefined` at [this line](https://github.com/eclipsesource/jsonforms/blob/1e44159b4e9e7aaf4f5324b6f67636e5e35bd777/packages/material-renderers/src/mui-controls/MuiAutocomplete.tsx#L85):

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "properties": {
    "foo": {
      "type": "string",
      "anyOf": [
        {
          "enum": ["A", "B"]
        },
        {
          "enum": ["C", "D"]
        }
      ]
    }
  }
}
```

image

### Screenshots

_No response_

Firefox 117.0 (64-bit Windows)

v3.1.0

### Framework

React

### RendererSet

Material

### Additional context

I don't give any UI Schema(using default behavior).

LukasBoll:
closes #2177

### DIFF
--- a/packages/core/src/util/util.ts
+++ b/packages/core/src/util/util.ts
@@ -77,7 +77,17 @@ export const deriveTypes = (jsonSchema: JsonSchema): string[] => {
   if (!isEmpty(jsonSchema.items)) {
     return ['array'];
   }
-
+  if (!isEmpty(jsonSchema.enum)) {
+    const types: Set<string> = new Set();
+    jsonSchema.enum.forEach((enumElement) => {
+      if (typeof enumElement === 'string') {
+        types.add('string');
+      } else {
+        deriveTypes(enumElement).forEach((type) => types.add(type));
+      }
+    });
+    return Array.from(types);
+  }
   if (!isEmpty(jsonSchema.allOf)) {
     const allOfType = find(
       jsonSchema.allOf,
--- a/packages/core/test/util/derivetype.test.ts
+++ b/packages/core/test/util/derivetype.test.ts
@@ -65,6 +65,23 @@ test('derive type w/o type - items array', (t) => {
   t.is(deriveTypes(schema)[0], 'array');
 });
 
+test('derive type w/o type - enum', (t) => {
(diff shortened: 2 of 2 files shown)


packages/core/src/util/util.ts
```diff
@@ -77,7 +77,17 @@ export const deriveTypes = (jsonSchema: JsonSchema): string[] => {
...
     return ['array'];
   }
-
+  if (!isEmpty(jsonSchema.enum)) {
+    const types: Set<string> = new Set();
+    jsonSchema.enum.forEach((enumElement) => {
+      if (typeof enumElement === 'string') {
+        types.add('string');
+      } else {
+        deriveTypes(enumElement).forEach((type) => types.add(type));
```
sdirix, line 86: We need to check whether the array has a single item. If yes, then we should return that directly instead of an array

lucas-koehler (approved):
LGTM thanks :)

packages/core/src/util/util.ts
lucas-koehler: If we want to do that we should adapt the type definition of the deriveTypes function as it only declares string[] as a return type. Furthermore, we also return single value arrays further above. So IMO we can leave it like that.

lucas-koehler: @sdirix Or am I missing something?

sdirix: @lucas-koehler you are right. I was too fast when reviewing and got the use case of this function confused. The code is fine.

PR #2196, opened 2023-10-24
WIP: introduce controlled & uncontrolled states for data/onChange management

### LINKED ISSUE #2174: Data loss when using React controlled style
### Describe the bug

`<JsonForms/>` seems to do asynchronous `onChange`, which breaks controlled style.

### Expected behavior

`onChange` is synchronous, thereby allowing use in controlled style.

 Here is a very simple reproduction:

```tsx
const schema = {
  type: "object",
  properties: {
    value: {type: "number"},
  },
};

function App() {
  const [data, setData] = useState({value: 0});

  useEffect(() => {
    const interval = setInterval(() => {
      // This is a stand-in for some external data change.
      setData(data => ({...data}));
    }, 10);

    return () => clearInterval(interval);
  }, []);

  return (
    <JsonForms 
      data={data}
      renderers={vanillaRenderers} 
      cells={vanillaCells}
      schema={schema}
      onChange={({data}) => setData(data)}/>
  );
}
```

https://codesandbox.io/p/sandbox/intelligent-snyder-y2p7m4?file=%2Fsrc%2FApp.tsx%3A18%2C10

You can easily see that data is being lost when rapidly typing a few digits into the input field.

### Screenshots

_No response_

Any

v3.1.0

### Framework

React

### RendererSet

Vanilla

I didn't test other RendererSets, but I assume they would have the same issue in React.

### Additional context

It seems that the problem might be caused, at least in part, by the [debounce in `JsonFormsContext.tsx`](https://github.com/eclipsesource/jsonforms/blob/c1686c1db33332bd7e820d53fdc29c27439acf1e/packages/react/src/JsonFormsContext.tsx#L216C6-L216C6), which, at first glance, seems to be fundamentally incompatible with controlled style.

dholmes2409:
Resolves #2174 

- Introduces 2 new modes which can be configured via a new 'mode' prop
- Missing tests as pending review from JsonForms team to be sure we're on the right track
- Missing configuration for CustomRenders as still pending discussion on #2174

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "npmClient": "pnpm",
   "useWorkspaces": true
 }
--- a/packages/angular-material/package.json
+++ b/packages/angular-material/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/angular-material",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "description": "Material Renderer Set for Angular module of JSON Forms",
   "repository": "https://github.com/eclipsesource/jsonforms",
   "bugs": "https://github.com/eclipsesource/jsonforms/issues",
@@ -69,8 +69,8 @@
     "@angular/material": "^12.0.0 || ^13.0.0 || ^14.0.0",
     "@angular/platform-browser": "^12.0.0 || ^13.0.0 || ^14.0.0",
     "@angular/router": "^12.0.0 || ^13.0.0 || ^14.0.0",
-    "@jsonforms/angular": "3.2.0-alpha.1",
-    "@jsonforms/core": "3.2.0-alpha.1",
+    "@jsonforms/angular": "3.2.0-alpha.3",
+    "@jsonforms/core": "3.2.0-alpha.3",
     "core-js": "^2.5.3",
     "rxjs": "^6.5.3 || ^7.4.0"
   },
--- a/packages/angular-test/package.json
+++ b/packages/angular-test/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/angular-test",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "private": true,
   "main": "./lib/index.js",
   "dependencies": {
--- a/packages/angular/package.json
+++ b/packages/angular/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/angular",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "description": "Angular module of JSON Forms",
   "repository": "https://github.com/eclipsesource/jsonforms",
   "bugs": "https://github.com/eclipsesource/jsonforms/issues",
@@ -71,7 +71,7 @@
   "peerDependencies": {
     "@angular/core": "^12.0.0 || ^13.0.0 || ^14.0.0",
     "@angular/forms": "^12.0.0 || ^13.0.0 || ^14.0.0",
-    "@jsonforms/core": "3.2.0-alpha.1",
+    "@jsonforms/core": "3.2.0-alpha.3",
     "rxjs": "^6.5.3 || ^7.4.0"
   },
   "devDependencies": {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/core",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "description": "Core module of JSON Forms",
   "repository": "https://github.com/eclipsesource/jsonforms",
   "bugs": "https://github.com/eclipsesource/jsonforms/issues",
--- a/packages/examples-app/package.json
+++ b/packages/examples-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/examples-app",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "private": true,
   "license": "MIT",
   "type": "module",
--- a/packages/examples-react/package.json
+++ b/packages/examples-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/examples-react",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "private": true,
   "dependencies": {
     "@jsonforms/core": "workspace:*",
--- a/packages/examples/package.json
+++ b/packages/examples/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/examples",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "description": "JSON Forms Example Data",
   "repository": "https://github.com/eclipsesource/jsonforms",
   "bugs": "https://github.com/eclipsesource/jsonforms/issues",
@@ -47,7 +47,7 @@
     "lodash": "^4.17.21"
   },
   "peerDependencies": {
-    "@jsonforms/core": "3.2.0-alpha.1"
+    "@jsonforms/core": "3.2.0-alpha.3"
   },
   "devDependencies": {
     "@jsonforms/core": "workspace:*",
--- a/packages/material-renderers/package.json
+++ b/packages/material-renderers/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/material-renderers",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "description": "Material Renderer Set for JSON Forms",
   "repository": "https://github.com/eclipsesource/jsonforms",
   "bugs": "https://github.com/eclipsesource/jsonforms/issues",
@@ -85,8 +85,8 @@
   "peerDependencies": {
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
-    "@jsonforms/core": "3.2.0-alpha.1",
-    "@jsonforms/react": "3.2.0-alpha.1",
+    "@jsonforms/core": "3.2.0-alpha.3",
+    "@jsonforms/react": "3.2.0-alpha.3",
     "@mui/icons-material": "^5.11.16",
     "@mui/material": "~5.13.0",
     "@mui/x-date-pickers": "^6.0.0",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/react",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "description": "React module of JSON Forms",
   "repository": "https://github.com/eclipsesource/jsonforms",
   "bugs": "https://github.com/eclipsesource/jsonforms/issues",
@@ -86,7 +86,7 @@
     "lodash": "^4.17.21"
   },
   "peerDependencies": {
-    "@jsonforms/core": "3.2.0-alpha.1",
+    "@jsonforms/core": "3.2.0-alpha.3",
     "react": "^16.12.0 || ^17.0.0 || ^18.0.0"
   },
   "optionalPeerDependencies": {
--- a/packages/vanilla-renderers/package.json
+++ b/packages/vanilla-renderers/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/vanilla-renderers",
-  "version": "3.2.0-alpha.1",
+  "version": "3.2.0-alpha.3",
   "description": "Default Renderer Set for JSON Forms",
   "repository": "https://github.com/eclipsesource/jsonforms",
   "bugs": "https://github.com/eclipsesource/jsonforms/issues",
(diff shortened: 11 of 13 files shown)


sdirix:
Thanks for the contribution :hear_no_evil: ! I'll take a detailed look once I have some cycles available, probably at the end of the week / start of next week.

packages/react/src/JsonFormsContext.tsx
```diff
@@ -126,11 +129,53 @@ const useEffectAfterFirstRender = (
...
+  useEffect(() => {
+    onChangeRef.current = onChange;
+  }, [onChange]);
+
+  const StateProvider =
+    initState.config.mode === 'controlled'
+      ? JsonFormsControlledStateProvider
+      : initState.config.mode === 'uncontrolled'
+      ? JsonFormsUncontrolledStateProvider
+      : JsonFormsMixedStateProvider;
```
sdirix, line 165: We should probably put the `StateProvider` in `useState` and issue a warning in case a developer tries to switch the mode after the fact.

packages/react/src/JsonFormsContext.tsx
```diff
@@ -144,6 +189,89 @@ export const JsonFormsStateProvider = ({
...
+  initState,
+  onChangeRef,
+}: JsonFormsStateProviderProps & { onChangeRef: any }) => {
+  const { data, schema, uischema, ajv, validationMode, additionalErrors } =
+    initState.core;
+
+  const onChangeHandler = useCallback(
+    (...args: any[]) => onChangeRef.current?.(...args),
+    []
+  );
```
sdirix, line 250: Is there a benefit of introducing this additional `useCallback`? Later where we use `onChangeHandler` we could just use `onChangeRef.current?` directly.

packages/react/src/JsonFormsContext.tsx
```diff
@@ -126,11 +129,53 @@ const useEffectAfterFirstRender = (
...
 export const JsonFormsStateProvider = ({
   children,
   initState,
   onChange,
-}: any) => {
+}: JsonFormsStateProviderProps & JsonFormsReactProps) => {
+  const onChangeRef = useRef(onChange);
+  useEffect(() => {
+    onChangeRef.current = onChange;
+  }, [onChange]);
```
sdirix, line 158: This was copied from the previous code and not introduced by you, but actually we don't need the `useEffect` here. We can simply set the `onChangeRef.current` directly here without the effect.

packages/react/src/JsonFormsContext.tsx
```diff
@@ -144,6 +189,89 @@ export const JsonFormsStateProvider = ({
...
+    )
+  );
+
+  const actualDispatch = (action: CoreActions) => {
+    if (action.type === 'jsonforms/UPDATE') {
+      const updatedState = coreReducer(core, action);
+      return onChangeHandler({
+        data: updatedState.data,
+        errors: core.errors,
+      });
```
sdirix, line 269: Ah interesting, this is not only emitting the data but actually is some kind of `middleware` support, only for the `jsonforms/UPDATE` case. However the user has no chance to actually return something sensible in the `onChangeHandler` here. To not update the internal state we need to return the `core` here, but the user does not have access to it.

Hmm thinking about it, we could actually just introduce a generic middleware support and then this use case of controlled vs uncontrolled mode would also be covered.

This could look like this:
```javascript
const middleware = props.middleware ?? (core, action, coreDispatch) => coreDispatch(core, action);

const actualDispatch = (action: CoreActions) => {
  return middleware(core, action, coreDispatch);
}
```

The middleware for this case could then look like this:

```javascript
const controlledModeMiddleware = (state, action, defaultDispatch) => {
  if(action.type === 'jsonforms/UPDATE'){
     const updatedState = defaultDispatch(state, action);
     myChangeHandler(data: updatedState.data, errors: updatedState.errors);
     return state;
   } else {
     return defaultDispatch(state, action);
   }
}
```

The new config could then be used to disable the default (debounced) emitter. Although not even that is needed as the `onChange` handler does not need to be handed over in that case.

sdirix (commented):
I had a look at the PR already. I'm very intrigued by the suggestion to return a user input in the customized dispatch. This basically introduces some kind of middleware support. Thinking about it, I think this use case can also be solved by actually introducing generic middleware support. I fleshed this out a bit in a comment. Let me know what you think about it.

packages/react/src/JsonFormsContext.tsx
```diff
@@ -126,11 +129,53 @@ const useEffectAfterFirstRender = (
...
+      : JsonFormsMixedStateProvider;
+
+  return (
+    <StateProvider initState={initState} onChangeRef={onChangeRef}>
+      {children}
+    </StateProvider>
+  );
+};
+
+const JsonFormsMixedStateProvider = ({
```
dchambers, line 174: Maybe something you'll come back to as the PR is refined, but might there be an opportunity to repeat the amount of repeated code between `JsonFormsMixedStateProvider`, `JsonFormsControlledStateProvider` and `JsonFormsUncontrolledStateProvider` (e.g. either by having some state providers be defined in terms of other state providers, by having the providers share re-usable blocks)?

packages/react/src/JsonFormsContext.tsx
```diff
@@ -208,36 +407,10 @@ export const JsonFormsStateProvider = ({
...
-   * of time we debounce them over a short amount of time. 10ms was chosen as this worked well
-   * even on low-end mobile device settings in the Chrome simulator.
-   */
-  const debouncedEmit = useCallback(
-    debounce((...args: any[]) => onChangeRef.current?.(...args), 10),
-    []
-  );
-  useEffect(() => {
-    debouncedEmit({ data: core.data, errors: core.errors });
+    if (initState.config.mode !== 'controlled') {
```
dchambers, line 411: Having mode checking logic still be needed in `JsonFormsDataProvider` (and not be in the state providers) may be indicative of an opportunity to improve things further.

dchambers (commented):
LGTM! Some comments for you however.

packages/react/src/JsonFormsContext.tsx
dchambers: Hi @sdirix 👋.

As I mentioned on a couple of my PR review comments, there's a lot of repetition between the different state providers, and one piece of mode related logic is not currently encapsulated within the relevant state provider, where it would ideally reside.

I just performed some diffs between the mixed, controlled and uncontrolled state providers to see what the lie of the land was, and this is what I determined:

1. Mixed debounces `onChange`.
2. Uncontrolled passes `core.data` to `coreDispatch(Actions.updateCore)`, whereas Mixed and Controlled pass `data`.
3. Controlled returns a wrapped dispatch function rather than the underlying dispatch function.
4. Controlled 'effectively' passes a dummy `onChange`:
   1. This is currently implemented within `JsonFormsDataProvider`, which only invokes `onChange` if `mode !== 'controlled'`, but it makes more sense that all mode logic be centralised within the data provider.

This means that there are three points of variance that any middleware (or data-provider) would have to account for:

1. Whether the internally or externally sourced data is passed to the reducer.
2. Whether `onChange` is wrapped.
3. Whether `dispatch` is wrapped.

As I understand it, the middleware idea would only help with the last of these three, whereas the data provider can handle all three.

Although the state providers currently have a lot of repetition between them, I've taken a look and this can be dramatically reduced by creating a couple of shared hooks (e.g. `useCoreDispatcher` and `useCoreUpdater`) to encapsulate the common blocks of code. If we also take the opportunity to move the state providers into their own files, then that would clean up things even further.

Let me know if you agree or not, and I (or @dholmes2409) can take a look at getting the PR updated as described.

sdirix: Hi @dchambers, thanks for the summary! I thought we could efficiently tackle the "mixed" vs "controlled" case by only implementing the suggested middleware approach without the need to introduce multiple state providers. However I might have also overlooked some edge case / use case. If only "mixed" and "controlled" is offered, would this be sufficient for your use cases, or would you specifically require all three modes?

I will take a more detailed look at your analysis and my suggestion early next week and then come back to you with the go ahead for the middleware or the different state providers. Overall planning wise I'm pretty confident that we will get this into a November release.

dchambers: > Hi @dchambers, thanks for the summary! I thought we could efficiently tackle the "mixed" vs "controlled" case by only implementing the suggested middleware approach without the need to introduce multiple state providers. However I might have also overlooked some edge case / use case. If only "mixed" and "controlled" is offered, would this be sufficient for your use cases, or would you specifically require all three modes?

Hi @sdirix 👋, I hope you had a good weekend. "controlled" is the only mode we'd really want to use, so this would definitely be sufficient for us if it can be made to work. That said, I'm not currently clear as to how a middleware could replace the `onChange` that ends up being passed to `JsonFormsDataProvider`, if you're able to provide some clarity here?

sdirix: Hi @dchambers!

I quickly implemented my suggestion. However I did not yet test nor compile it. You can find it in my fork, see [here](https://github.com/sdirix/jsonforms/tree/middleware-support).

Please have a look. Using a middleware you no longer need to hand over a "onChange" as you get the latest updates anyway. Therefore the `debounce` is also avoided.

dchambers: Hi @sdirix 👋. Thanks for this, and the code looks really clean 👌.

> Using a middleware you no longer need to hand over a "onChange" as you get the latest updates anyway. Therefore the debounce is also avoided.

Sorry, I'm struggling to understand this bit. Your branch has an implementation of `defaultMiddleware`, but no implementation of `controlledMiddleware`, yet this would probably help to clarify things. Here's a stab at what I think this might look like:

```ts
const controlledMiddleware: Middleware = (state, action, defaultDispatch) => {
  if (action.type === 'jsonforms/UPDATE') {
    const [updatedState, errors] = reducer(state, action);
    return onChange({
      data: updatedState,
      errors,
    });
  } else {
    return defaultDispatch(state, action);
  }
};
```

but this begs the question: "Where do `reducer` and `onChange` come from?".

Unless I'm misunderstanding something, I'd therefore expect the middleware interface to get access to more of the same constants that the data provider component had access to. Consider the following pseudcode, for example:

```ts
export type MiddlewareDispatcher = (
  state: JsonFormsCore,
  action: CoreActions
) => JsonFormsCore;

export type MiddlewareOptions = {
  reducer: (
    state: JsonFormsCore,
    action: CoreActions
  ) => [JsonFormsCore, JsonFormsErrors];
  onChange: (update: { data: JsonFormsCore; errors: JsonFormsErrors }) => void;
  defaultDispatcher: MiddlewareDispatcher;
};

export type Middleware = (
  state: JsonFormsCore,
  action: CoreActions,
  options: MiddlewareOptions
) => JsonFormsCore;

export const defaultMiddleware: Middleware = (state, action, options) =>
  options.defaultDispatcher(state, action);

export const controlledMiddleware: Middleware = (state, action, options) => {
  if (action.type === 'jsonforms/UPDATE') {
    const [updatedState, errors] = options.reducer(state, action);
    options.onChange({
      data: updatedState,
      errors,
    });

    return updatedState;
  } else {
    return options.defaultDispatcher(state, action);
  }
};
```

Thoughts?

sdirix: Hi @dchambers,

I didn't add the `controlledMiddleware` to the repository as I expect it to be implemented in JSON Forms consumers. If more people have this use case we could maybe offer a factory util to create such a middleware conveniently.

Based on your example I think this can look like this:

```typescript
/**
 * Controlled variant of JSON Forms.
 * Props are only read in first render pass. Updates to them are ignored.
 */
const ControlledJsonForms = ({ schema, initialData }) => {
  const [formState, setFormState] = useState<{
    data: any;
    errors: ErrorObject[];
  }>({ data: initialData, errors: [] });

  const validator = useMemo(() => createAjv().compile(schema), []);

  const controlledMiddleware: Middleware = (state, action, reducer) => {
    if (action.type === 'jsonforms/UPDATE') {
      const newState = reducer(state, action);
      validator(newState.data);
      setFormState({ data: newState.data, errors: validator.errors || [] });
      return state; // this will lead to JSON Forms ignoring the change within itself
    } else {
      return reducer(state, action);
    }
  };

  return (
    <JsonForms
      schema={schema}
      data={formState.data}
      additionalErrors={formState.errors}
      validationMode='NoValidation'
      middleware={controlledMiddleware}
      renderers={materialRenderers}
      cells={materialCells}
    />
  );
};
```
The `controlledMiddeware` could be placed in a `useCallback` but it does not have to be.

Overall this looks like I expected, except for the validation part. I realized that if the validation shall also be controlled then it either must be performed outside, like I did now, or it would require more code within the `controlledMiddleware` delving deeper into the internals. As there is no downside in doing the validation on our own I preferred that approach.

What do you think, does this cover your use case?

sdirix: @dchambers Does this work for you? If yes, then I would like to get it in ideally before the next release.

dholmes2409: Hey @sdirix, I had play around with this today and it seems to cover our use case. We would need to tweak our usage as we pass a custom onChange handler but we could easily move the logic into the controlledMiddleware.
Should I close this PR and we go with the approach you committed to your fork?

sdirix: To get this in faster it would be nice to have some tests. Would you like to contribute them? You could force push this PR and use my commit as basis. In the tests you could add the `controlledMiddleware` and verify that your use case works as expected. Then it's ensured that it will not be broken in the future ;)

If you don't have time then that's also fine, we are definitely eager to get this feature in, so we will eventually also work on this. However as always I can't give any promises regarding a due date as this depends on our work load.

dholmes2409: Moved this into PR #2220

Issue #2198, opened 2023-10-25
Orphaned labels on single select dropdown


sgober:
### Describe the bug

When testing for 508 compliance, I noticed that the single select dropdown renderers have orphaned labels.  The htmlFor on the label is for example "#/properties/92-input" but the id on the input is "#/properties/92", thus the label is orphaned and it shows up that the input has no label.

### Expected behavior

Label should be correctly associated with the single select dropdown.

### Steps to reproduce the issue

I use the wave chrome extension to see what 508 compliance errors there are, so all you need to do is have a single select in your schema and you can see that the label and input are not associated

### Screenshots

Wave output example (red is missing input label error and orange is orphaned label warning):
<img width="894" alt="Screenshot 2023-10-25 at 10 45 49 AM" src="https://github.com/eclipsesource/jsonforms/assets/35042815/3a051dcc-10dc-4c92-ab30-d68cd18bc8ea">

Output example where for and id don't match:
<img width="542" alt="Screenshot 2023-10-25 at 10 45 24 AM" src="https://github.com/eclipsesource/jsonforms/assets/35042815/1313a07d-a53d-49e0-8028-6ef0ddfc0d9c">

### In which browser are you experiencing the issue?

Google Chrome 118.0.5993.117

### Which Version of JSON Forms are you using?

v3.0.0

### Framework

React

### RendererSet

Material

sdirix:
Hi @sgober,

Thanks for the report! Would you like to contribute a fix?

sgober:
@sdirix - I have this fixed locally but can't push my fix branch up.  Can you give me creds to do this?

sgober:
PR here: https://github.com/eclipsesource/jsonforms/pull/2199

sdirix:
Was fixed with `3.2`

PR #2199, opened 2023-10-26
BUGFIX: Orphaned labels on single select dropdown

sgober:
Removed "-input" from id on input so that the label and input are associated correctly.  Will fix issue #2198


sdirix:
Thanks for the contribution :heart: We will take a look soon

sdirix (approved):
Works for me as intended. Thanks for the contribution :heart:

PR #2200, opened 2023-10-27
Fix/datepicker validation

### LINKED ISSUE #2183: Date Picker Validation
### Describe the bug

When you delete (using the backspace button) an invalid date (whether by the default date format or additional errors such as min or max date), even if the field is not required the invalid date error persists. 

Screenshot 2023-10-04 at 12 05 13 PM

### Expected behavior

When you delete (using the backspace button) an invalid date (whether by the default date format or additional errors such as min or max date), if the field is not required there is no error showing and the data should be undefined for that input.

### Screenshots

Screen Recording 2023-10-05 at 12.21.09 PM.mov

Google Chrome Version 117.0.5938.92 

v3.0.0

### Framework

React

### RendererSet

Material

### Additional context

_No response_

LukasBoll:
Previously, data wasn't cleared when the input field was emptied.
This commit will set the value to 'undefined' when the input is empty or
invalid upon blurring. During editing, the data is only updated when the
current input is valid.

Closes #2183

### DIFF
--- a/packages/material-renderers/src/controls/MaterialDateControl.tsx
+++ b/packages/material-renderers/src/controls/MaterialDateControl.tsx
@@ -23,9 +23,10 @@
   THE SOFTWARE.
 */
 import merge from 'lodash/merge';
-import React, { useMemo } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import {
   ControlProps,
+  defaultDateFormat,
   isDateControl,
   isDescriptionHidden,
   RankedTester,
@@ -35,7 +36,12 @@ import { withJsonFormsControlProps } from '@jsonforms/react';
 import { FormHelperText, Hidden } from '@mui/material';
 import { DatePicker, LocalizationProvider } from '@mui/x-date-pickers';
 import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
-import { createOnChangeHandler, getData, useFocus } from '../util';
+import {
+  createOnBlurHandler,
+  createOnChangeHandler,
+  getData,
+  useFocus,
+} from '../util';
 
 export const MaterialDateControl = (props: ControlProps) => {
   const [focused, onFocus, onBlur] = useFocus();
@@ -62,8 +68,10 @@ export const MaterialDateControl = (props: ControlProps) => {
     appliedUiSchemaOptions.showUnfocusedDescription
   );
 
+  const [key, setKey] = useState<number>(0);
+
   const format = appliedUiSchemaOptions.dateFormat ?? 'YYYY-MM-DD';
-  const saveFormat = appliedUiSchemaOptions.dateSaveFormat ?? 'YYYY-MM-DD';
+  const saveFormat = appliedUiSchemaOptions.dateSaveFormat ?? defaultDateFormat;
 
   const views = appliedUiSchemaOptions.views ?? ['year', 'day'];
 
@@ -73,20 +81,36 @@ export const MaterialDateControl = (props: ControlProps) => {
     ? errors
     : null;
   const secondFormHelperText = showDescription && !isValid ? errors : null;
+
+  const updateChild = useCallback(() => setKey((key) => key + 1), []);
+
   const onChange = useMemo(
     () => createOnChangeHandler(path, handleChange, saveFormat),
     [path, handleChange, saveFormat]
   );
 
+  const onBlurHandler = useMemo(
+    () =>
+      createOnBlurHandler(
+        path,
+        handleChange,
+        format,
+        saveFormat,
+        updateChild,
+        onBlur
+      ),
+    [path, handleChange, format, saveFormat, updateChild]
+  );
   const value = getData(data, saveFormat);
 
   return (
     <Hidden xsUp={!visible}>
       <LocalizationProvider dateAdapter={AdapterDayjs}>
         <DatePicker
+          key={key}
           label={label}
           value={value}
-          onChange={onChange}
+          onAccept={onChange}
           format={format}
           views={views}
           disabled={!enabled}
@@ -109,7 +133,7 @@ export const MaterialDateControl = (props: ControlProps) => {
               },
               InputLabelProps: data ? { shrink: true } : undefined,
               onFocus: onFocus,
-              onBlur: onBlur,
+              onBlur: onBlurHandler,
             },
           }}
         />
--- a/packages/material-renderers/src/controls/MaterialDateTimeControl.tsx
+++ b/packages/material-renderers/src/controls/MaterialDateTimeControl.tsx
@@ -22,10 +22,11 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 */
-import React, { useMemo } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import merge from 'lodash/merge';
 import {
   ControlProps,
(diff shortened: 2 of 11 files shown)


sdirix:
New behavior:
 - During editing, when value is a valid date/time, immediately write through
 - During editing, when value is not valid, don't trigger any change
 - On blur, if value is not valid, write `undefined` in data

packages/material-renderers/src/controls/MaterialDateControl.tsx
```diff
@@ -73,17 +80,28 @@ export const MaterialDateControl = (props: ControlProps) => {
     ? errors
     : null;
   const secondFormHelperText = showDescription && !isValid ? errors : null;
+
+  const updateChild = () => {
+    setKey(key + 1);
+  };
```
sdirix, line 86: Whenever the new value of a state depends on the old value, the functional updater should be used, i.e. `const updateChild = () => setKey((key) => key + 1)`

packages/material-renderers/src/controls/MaterialDateControl.tsx
```diff
@@ -73,17 +80,28 @@ export const MaterialDateControl = (props: ControlProps) => {
...
   const onChange = useMemo(
     () => createOnChangeHandler(path, handleChange, saveFormat),
     [path, handleChange, saveFormat]
   );
 
+  const onBlur = useMemo(
+    () =>
+      createOnBlurHandler(path, handleChange, format, saveFormat, updateChild),
+    [path, handleChange, format, saveFormat, updateChild]
+  );
```
sdirix, line 97: This `useMemo` will never work as `updateChild` is a new instance on every `render` call, therefore always breaking the memo.

packages/material-renderers/src/controls/MaterialDateControl.tsx
```diff
@@ -62,6 +67,8 @@ export const MaterialDateControl = (props: ControlProps) => {
     appliedUiSchemaOptions.showUnfocusedDescription
   );
 
+  const [key, setKey] = useState<any>(0);
```
sdirix, line 70: We should avoid `any` at all cost and only use it when there is no better way. Here we can just type as `number`.

packages/material-renderers/src/controls/MaterialDateTimeControl.tsx
```diff
@@ -66,6 +71,8 @@ export const MaterialDateTimeControl = (props: ControlProps) => {
   const format = appliedUiSchemaOptions.dateTimeFormat ?? 'YYYY-MM-DD HH:mm';
   const saveFormat = appliedUiSchemaOptions.dateTimeSaveFormat ?? undefined;
 
+  const [key, setKey] = useState<any>(0);
```
sdirix, line 74: The `key` here is never used

packages/material-renderers/src/controls/MaterialTimeControl.tsx
```diff
@@ -56,6 +61,8 @@ export const MaterialTimeControl = (props: ControlProps) => {
   const appliedUiSchemaOptions = merge({}, config, uischema.options);
   const isValid = errors.length === 0;
 
+  const [key, setKey] = useState<any>(0);
```
sdirix, line 64: The `key` here is never used

packages/material-renderers/src/controls/MaterialDateControl.tsx
```diff
@@ -35,10 +35,15 @@ import { withJsonFormsControlProps } from '@jsonforms/react';
...
+import {
+  createOnBlurHandler,
+  createOnChangeHandler,
+  getData,
+  useFocus,
+} from '../util';
 
 export const MaterialDateControl = (props: ControlProps) => {
-  const [focused, onFocus, onBlur] = useFocus();
+  const [focused, onFocus] = useFocus();
```
sdirix, line 46: We lose the focus handling of `useFocus` here, i.e. it expects that `onBlur` is called at some point to remove the focus. Shouldn't we call the `onBlur` here in our custom `onBlur`?

sdirix (changes requested):
In principal looks good, however I found some issues in the code. Please also test the exceptional cases for which we require all that custom blur update handling.

packages/material-renderers/src/util/datejs.tsx
```diff
@@ -13,10 +13,31 @@ export const createOnChangeHandler =
...
+  };
+
+export const createOnBlurHandler =
+  (
+    path: string,
+    handleChange: (path: string, value: any) => void,
+    format: string | undefined,
+    saveFormat: string | undefined,
+    rerenderChild: () => void,
+    resetFocus: () => void
```
sdirix, line 29: This method does not care why this callback is handed over. The parameter should be called `onBlur` or `callback`

packages/material-renderers/src/controls/MaterialTimeControl.tsx
```diff
@@ -75,16 +82,32 @@ export const MaterialTimeControl = (props: ControlProps) => {
     : null;
   const secondFormHelperText = showDescription && !isValid ? errors : null;
 
+  const updateChild = useCallback(() => setKey((key) => key + 1), [setKey]);
```
sdirix, line 85: technically `setKey` is not required to be in the dependency list as it's guaranteed to be stable by React, however adding it doesn't break anything.

sdirix (commented):
one minor change request regarding the naming. besides that, look's good!

sdirix (changes requested):
There is still some error lurking. To reproduce:

1. Open the React Material Ui example application ([link](https://deploy-preview-2200--jsonforms-examples.netlify.app/))
2. Into the `birthDate` field, enter the year 900. Observe that the `data` is not updated as intended because this year is invalid
3. Blur the field, e.g. by clicking into another one
4. Observe that the field is now empty, however the invalid date was saved into the data
5. Refocus the field and blur it again
6. Now the invalid date is actually removed from the data

Expected behavior:
After executing steps 1-3, the field and the actual data should be empty.

DateErrors

LukasBoll:
Hi @sdirix ,
this commit will change the default format of datetime to "YYYY-MM-DDTHH:mm:ss.sssZ," making it coherent with `convertDateToString` wich is used for `createDefaultValue` in the core.
The previous default format was the default format from dayjs (current date in ISO8601, without fraction seconds, e.g. '2020-04-02T08:02:17-05:00')
Is this fine or should we keep the old default format?

sdirix:
>Is this fine or should we keep the old default format?

This is fine as `sss` is also valid ISO8601

packages/material-renderers/src/controls/MaterialTimeControl.tsx
```diff
@@ -64,7 +72,7 @@ export const MaterialTimeControl = (props: ControlProps) => {
   );
 
   const format = appliedUiSchemaOptions.timeFormat ?? 'HH:mm';
-  const saveFormat = appliedUiSchemaOptions.timeSaveFormat ?? 'HH:mm:ss';
+  const saveFormat = appliedUiSchemaOptions.timeSaveFormat ?? defaultTimeFormat;
```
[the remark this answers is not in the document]
LukasBoll: The previous implementation didn’t support formats without seconds by default. So this wouldn't be an unexpected/new behavior. 
Currently I adjusted the examples in our application to use the `HH:mm:ss` format.
But it is also possible to check if the time-value uses the `HH:mm:ss` or `HH:mm` format and set the `saveFormat` accordingly. Would you prefer this solution?

sdirix (approved):
LGTM! Thanks

PR #2201, opened 2023-10-27
fix(core): Disable schema caching in AJV instance

### LINKED ISSUE #2071: schema with key or id "<url>" already exists
### Describe the bug

It seems like the schema passed to the ajv instance is not cleared when a new update/schema is passed to the jsonForms with updated values.

image

This seems to come from 
`thisAjv.compile(action.schema);` reusing existing instance and adding to it.

### Expected behavior

Can pass in the same schema with updated values.

It should likely call removeSchema on update?

Schema root has $id.
```
{
  "$id": "something",
  "type": "object",
  "properties": {}
}
```

### Screenshots

_No response_

Version 107.0.5304.89 (Official Build) (64-bit)

3.0.0

### Framework

Angular

### RendererSet

### Additional context

_No response_

LukasBoll:
Set the 'addUsedSchema' option to false during AJV instantiation to prevent schema caching by default. This resolves issues where updating a schema that contains an ID leads to duplicate ID errors in AJV.

Closes #2071

### DIFF
--- a/packages/core/src/reducers/core.ts
+++ b/packages/core/src/reducers/core.ts
@@ -90,16 +90,6 @@ const initState: JsonFormsCore = {
   additionalErrors: [],
 };
 
-const reuseAjvForSchema = (ajv: Ajv, schema: JsonSchema): Ajv => {
-  if (
-    Object.prototype.hasOwnProperty.call(schema, 'id') ||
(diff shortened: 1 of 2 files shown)


sdirix (approved):
LGTM

PR #2202, opened 2023-10-27
fix(material): Resolve refs in object renderer

### LINKED ISSUE #2187: Property with reference is not rendered
### Describe the bug

While the address renders fine, given the json schema below, the parent does not render at all.
```
type: "object",
definitions: {
  Client: {
      "address": {
          "type": "array",
          "items": {
              "$ref": "#/definitions/Address"
          }
      },
      "parent": {
          // https://jsonforms.io/docs/tutorial/multiple-forms#interlinked-inputs
          "$ref": "#/definitions/SearchableObject/properties/a_very_special_field"
          //"$ref": "#/definitions/SearchableObject"
          //"type": "object"
          //items: {"$ref": "#/definitions/SearchableObject"}
          //item: {"$ref": "#/definitions/SearchableObject"}
      }
  },
  SearchableObject: {
      "type": "object",
      "properties": {
          "a_very_special_field": {
              "type": "integer"
          }
      },
      "required": []
  },
 properties: {
    "Client": {
        "$ref": "#/definitions/Client"
    }
}
```

As soon as I turn the `parent` into an array field just like the `address` it renders fine again (but of course is an array which is not what I need).

### Expected behavior

Render the Sub-Form for the `parent` object like for array fields

Render form using the jsonschema above und leave the uischema undiefined.

### Screenshots

_No response_

Google Chrome 

3.1.0

### Framework

React

### RendererSet

Material

### Additional context

_No response_

LukasBoll:
Previously, UI schemas generated in the Material object renderer were incomplete because not all refs could be resolved.

Closes #2187

### DIFF
--- a/packages/angular-material/src/other/object.renderer.ts
+++ b/packages/angular-material/src/other/object.renderer.ts
@@ -73,7 +73,12 @@ export class ObjectControlRenderer extends JsonFormsControlWithDetail {
         delete newSchema.oneOf;
         delete newSchema.anyOf;
         delete newSchema.allOf;
-        return Generate.uiSchema(newSchema, 'Group');
+        return Generate.uiSchema(
+          newSchema,
+          'Group',
+          undefined,
+          this.rootSchema
+        );
       },
       props.uischema,
       props.rootSchema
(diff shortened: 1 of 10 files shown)


sdirix (changes requested):
Looks good! Can you also check all other invocations of `Generate.uiSchema` in the code base? We are missing the extra parameters in multiple places.

sdirix (approved):
LGTM

PR #2182, opened 2023-09-29
Make material renderers aware of input variant in theme

sebastianfrey:
Resolves #1797.

This PR makes material renderers aware of the input variant defined in the global MUI theme.

The example app was enhanced with a select, which allows to change the input variant.

PR #2205, opened 2023-10-31
Use `options.value` instead of `ev.currentTarget.value` for value handling within vanilla-renderers

### LINKED ISSUE #2045: Vanilla react renderer: integer enum radio group values are converted to string
### Describe the bug

I want to render an integer enum control as a radio via the following schemas:

schema.json
```
"amount": {
  "type": "integer",
  "enum": [50, 100, 250, 500]
}
```

uischema.json
```
{
  "type": "Control",
  "scope": "#/properties/amount",
  "options": {
    "format": "radio"
  }
}
```

The issue is now that apparently the values of the radio inputs are converted/considered to be string instead of integers. Because of that i cannot select any of the values because the validator is righfully complaining that the selected values are not valid.

### Expected behavior

The radio group values should be selectable

1. Provide schema as outlined above
2. Try to select one of the options

### Screenshots

Peek 2022-11-09 16-18

Google Chrome Version 107.0.5304.110 (Official Build) (64-bit)

v3.0.0

### Framework

### RendererSet

Vanilla

### Additional context

I am using Preact

TheZoker:
This PR fixes a couple of issues:
- Fix the folder reference for the example apps in the README file
- Fix the `index` file naming within the example folder of `vanilla-renderer` (bug introduced in #2182)
- Use `options.value` instead of `ev.currentTarget.value` for value handling to preserve the type of the value

Fixes #2045


sdirix (approved):
LGTM!

PR #2206, opened 2023-11-09
feat: respect "default" values when creating new elements in an array

### LINKED ISSUE #2195: Respect "default" values when creating new elements in an array

JSON Schema specifies `default` property. As there is no specific behavior defined in the specification regarding this attribute, we usually ignore it.

Usually we ignore this property in JSON Forms and [let for example AJV handle it](https://jsonforms.io/faq#how-can-i-use-default-values-within-json-forms). However in the case of adding new elements in arrays, it would make sense for JSON Forms to respect the `default` attribute and initialize the new element in this way. This feature could be made optional either via UI Schema options or the [config](https://github.com/eclipsesource/jsonforms/issues/2174#issuecomment-1771202751).

Support the `default` attribute in JSON Schema when creating new array elements.

Not supporting this feature or turn it off by default.

### Framework

Core, React, Angular, Vue 3

### RendererSet

_No response_

### Additional context

_No response_

AlexandraBuzila:
Initialize new array elements with their default values.

Can be tested in the `Array with defaults` example.

Fixes #2195


packages/core/src/util/renderer.ts
```diff
@@ -161,11 +165,32 @@ export const createDefaultValue = (schema: JsonSchema) => {
...
   }
 };
 
+/**
+ * Returns the default value defined in the given schema.
+ * @param schema the schema for which to create a default value.
+ * @returns {any}
+ */
+export const extractDefaults = (schema: JsonSchema) => {
+  if (schema.type === 'object' && schema.default === undefined) {
```
sdirix, line 181: The `type === 'object'` is optional in JSON Schema, so we should not rely on it. We have a util method somewhere determining the type of JSON (sub) schemas.

packages/core/src/util/renderer.ts
```diff
@@ -161,11 +165,32 @@ export const createDefaultValue = (schema: JsonSchema) => {
...
+    const result: { [key: string]: any } = {};
+    for (const key in schema.properties) {
+      const property = schema.properties[key];
```
sdirix, line 184: For each property we need to check whether it is a reference, i.e. `'$ref' in schema` and resolve accordingly

AlexandraBuzila:
Thank you for the review, Stefan! I integrated your comments.

packages/core/src/util/renderer.ts
```diff
@@ -137,33 +138,65 @@ export const showAsRequired = (
...
+    return 0;
+  } else if (hasType(resolvedSchema, 'boolean')) {
+    return false;
+  } else if (hasType(resolvedSchema, 'array')) {
+    return [];
+  } else if (hasType(resolvedSchema, 'null')) {
+    return null;
+  } else if (hasType(resolvedSchema, 'object')) {
+    return extractDefaults(resolvedSchema, rootSchema);
+  } else {
```
sdirix, line 175: I would prefer checking first for object and then later for null. So in case people modeled nullable objects we are more constructive.

packages/core/test/util/renderer.test.ts
```diff
@@ -493,39 +494,171 @@ test('mapDispatchToControlProps', (t) => {
   t.is(updateAction.updater(), 42);
 });
 
-test('createDefaultValue', (t) => {
-  t.true(
-    _.isDate(
-      createDefaultValue({
+test.only('createDefaultValue', (t) => {
```
sdirix, line 497:
```suggestion
test('createDefaultValue', (t) => {
```

sdirix (approved):
LGTM!

PR #2218, opened 2023-11-20
update  @mui/icons-material depdency in material-renderers

devbydaniel:
change dependency of @mui/icons-material in material-renderers from `~5.11.16` to `^5.11.16`


devbydaniel:
Sorry if I'm missing something super obvious - I basically just changed the `~` to `^`, updated the lock file and updated dayjs from `1.10.6` to `1.10.7`. The error during deployments happen to me even before applying the changes.

lucas-koehler:
Hi @devbydaniel ,
thanks for the contribution :)
I can see that even the normal build (not just the deployment) fails due to some react error. I don't know what exactly causes this. However, I think it makes sense to only update directly material related dependencies in this PR:

1. It's easier to see what could cause errors
2. Many of the other dependencies (e.g. webpack, rollup, react and its types, etc.) are also used in other packages and I'd prefer keeping the versions consistent.

Thus, my suggestion is to revert all changes that don't update mui, emotion or dayjs. What do you think?

lucas-koehler:
Hi @devbydaniel , I just tried to build on master and the normal build (`pnpm run build`) as well as the examples app build (`pnpm run build:examples-app` ) both worked fine. Maybe you reverted back to master but still had the new dependencies in the node_modules folder.

My suggestion would be to revert back to master, then clean the repository with `git clean -dxf`. This deletes all untracked files in the repository! Afterwards, adapt the dependencies of the material and the dayjs package. Then install dependencies with `pnpm install`. You should not need to call `pnpm up` because we do not want to generally update all dependencies in this PR.

devbydaniel:
@lucas-koehler Thanks again for your help, I think / hope I got it right this time!

lucas-koehler:
Hi @devbydaniel , thanks for the update. It seems correct but I'd like that we also make the `@mui/material` use `^` instead of `~` while we are at it to keep it consistent. Could you please update the `@mui/material` dependency in the peer and dev dependencies accordingly?
I think, you'll need to do the `git clean -dxf` and `pnpm install` again afterwards to update the lock file

devbydaniel:
@lucas-koehler done!

lucas-koehler (approved):
LGTM, thanks for your contribution :)

PR #2220, opened 2023-11-24
introduce middleware support 

dholmes2409:
Resolves https://github.com/eclipsesource/jsonforms/issues/2174

Introduces middleware support which can override state actions 'such as UPDATE' allowing for handling of controlled states effectively bypassing the debounce

### DIFF
--- a/packages/react/test/renderers/JsonForms.test.tsx
+++ b/packages/react/test/renderers/JsonForms.test.tsx
@@ -58,6 +58,7 @@ import {
 } from '../../src/JsonForms';
 import {
   JsonFormsStateProvider,
+  Middleware,
   useJsonForms,
   withJsonFormsControlProps,
 } from '../../src/JsonFormsContext';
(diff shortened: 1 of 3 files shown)


packages/react/test/renderers/JsonForms.test.tsx
```diff
@@ -1100,3 +1101,49 @@ test('JsonForms should use react to additionalErrors update', () => {
   expect(wrapper.find('h5').text()).toBe('Foobar');
   wrapper.unmount();
 });
+
+test('JsonForms should used provided middleware if available', () => {
```
dholmes2409, line 1105: @sdirix if there are any other scenarios you think would be good to add let me know, this was the most immediate one I could think of

sdirix: Hi! That's already great!

What's tested now is that if the middleware rejects a modification, that then `onChangeHandler` was not called. We should also check that the rendered value did not change.

It would be good to also test the modification case. For example if the middleware modifies an additional value in addition to the regular update and returns that as the new state. Then the `onChange` handler should also be given that modified value and the new value should be rendered.

I think then we have all basic cases covered!

dholmes2409: I've added a test case for modifying the state value within the middleware, hopefully this covers the modification case. I wasn't fully sure what you meant by `Then the onChange handler should also be given that modified value and the new value should be rendered.`. Wouldn't the onChange handler only be called if we explicitly call it within the middleware?

dchambers: Hi @sdirix 👋, how goes it? Just a reminder that the extra unit test you asked about has been added, and a query as to whether anything else will be needed before this PR can be further progressed?

sdirix (commented):
Looks very good. There is a small detail which needs to be adapted in the tests. This week is very busy for us, but I think we can merge/release this within 2 weeks.

sdirix (approved):

sdirix:
@dchambers @dholmes2409 Thank you so much for all of the contribution work :heart: and architecture discussions. Without your help this feature would not arrived so soon.

dchambers:
> @dchambers @dholmes2409 Thank you so much for all of the contribution work ❤️ and architecture discussions. Without your help this feature would not arrived so soon.

No, thank you helping us land this @sdirix 🙇

PR #2223, opened 2023-12-04
feat: Implement GitHub Action for Publishing

### LINKED ISSUE #2047: GH action for publish
We need a manual GH action for publishing to ease publishing and document the process

LukasBoll:
Introduce a GitHub Action to automate the publishing process for JSON Forms.
adjust dependencies with lerna
adjust peerdependencies with custom script

closes #2047


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - uses: pnpm/action-setup@v2.2.4
```
sdirix, line 25: We can't use any third party actions. I think you need to install `pnpm` via `npm`

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Configure Git Credentials
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
sdirix, line 23: Let's use Node 18 as 16 is no longer maintained

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      - uses: pnpm/action-setup@v2.2.4
+        name: Install pnpm
+        id: pnpm-install
+        with:
+          version: 8
+          run_install: false
+
+      - name: Install Packages
+        run: pnpm install --no-save
```
sdirix, line 33: install with frozen lockfile instead of `--no-save` so we exactly replicate the stored lock

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,55 @@
...
+          cd ../vue && pnpm pkg set peerDependencies.@jsonforms/core="${{ github.event.inputs.next_version }}"
+          cd ../vue-vanilla && pnpm pkg set peerDependencies.@jsonforms/core="${{ github.event.inputs.next_version }}" && pnpm pkg set peerDependencies.@jsonforms/vue="${{ github.event.inputs.next_version }}"
+
+      - name: Tag and Commit
+        run: |
+          git add -A && git commit --amend --no-edit
+          git tag v${{ github.event.inputs.next_version }} -f
+
+      - name: push
+        run: git push
```
sdirix, line 55: Is the tag pushed too with this commit? I think by default only branches are pushed

sdirix (commented):
Thanks! I left some comments

.github/workflows/publish.yaml
sdirix: Correction: We whitelisted this pnpm action, so it's fine to keep it

.github/workflows/publish.yaml
LukasBoll: I updated the version to 18, but I believe we are still recommending to use version 16 in the readme

.github/workflows/publish.yaml
LukasBoll: No, it doesn't. I added another command to push the tag

sdirix (commented):
Please add build and test steps, please check `preparePublish` script

sdirix (approved):

Issue #2122, opened 2023-03-31
Description is not rendered for array type


hungbie:
### Describe the bug

Normally, a description will be render for the field if they have `description` in their schema.
However, for field of type `array`, the `description` field is not being rendered.

### Expected behavior

`description` being rendered.

### Steps to reproduce the issue

Example code:
```typescript
import React from 'react';
import { JsonForms } from '@jsonforms/react';
import {
  materialRenderers,
  materialCells,
} from '@jsonforms/material-renderers';

export default function ExampleJsonForm() {
  const schema = {
    type: 'object',
    description: 'This is a description of the object.',
    properties: {
      comments: {
        type: 'array',
        description: 'This is a description of the comments array.',
        items: {
          type: 'object',
          properties: {
            unfocusedDescription: {
              type: 'string',
              description: 'This description is shown even when the control is not focused',
            },
          },
        },
      },
      unfocusedDescription: {
        type: 'string',
        description: 'This description is shown even when the control is not focused',
      },
    },
  };
  const uischema = {
    type: 'VerticalLayout',
    elements: [
      {
        type: 'Control',
        scope: '#/properties/comments',
        options: {
          showUnfocusedDescription: true,
        },
      },
      {
        type: 'Control',
        scope: '#/properties/unfocusedDescription',
        options: {
          showUnfocusedDescription: true,
        },
      },
    ],
  };
  const data = {
  };
  return (
      <JsonForms
        schema={schema}
        uischema={uischema}
        data={data}
        renderers={materialRenderers}
        cells={materialCells}
      />
  );
}
```

### Screenshots

![Screenshot 2023-03-31 at 4 31 36 PM](https://user-images.githubusercontent.com/25290547/229068582-3faa94a1-4a23-4252-8105-037c6c2804c3.png)

### In which browser are you experiencing the issue?

Google Chrome Version 111.0.5563.110 (Official Build) (x86_64)

### Which Version of JSON Forms are you using?

v3.0.0

### Framework

React

### RendererSet

Material

hungbie:
hope someone can take a look at this

sdirix:
Hi @hungbie, our array renderer does not print the description anywhere. I could imagine to support it as an overlay to the array label, similar to the boolean controls. Would you like to contribute this feature?

hungbie:
@sdirix sure. Happy to contribute, can you give me some pointers to where should I put this enhancement in?

SaSteffen:
[...]

PR #2227, opened 2023-12-12
Enhancement/2122 add description to array renderer

SaSteffen:
As discussed in #2122: This is a minimal change that renders a description for primitive and object array renderers in React MUI.

The Primitive example looks like this:
image

The other one looks like this:
image

As discussed, I just used a FormHelperText. We could achieve similar results using custom typography elements probably.


sdirix:
Hi @SaSteffen, thanks for the contribution :heart: We will take a look very soon

sdirix (changes requested):
Thanks for the contribution! The PR looks good overall.

I have two change requests:
 - Make sure to not render the `FormHelperText` at all if there is no description
 - Add a test case for the existence of the description when given

SaSteffen:
@sdirix Sure. Added the conditional rendering and the two tests.

packages/material-renderers/src/complex/TableToolbar.tsx
```diff
@@ -66,26 +76,29 @@ const TableToolbar = React.memo(function TableToolbar({
...
+                  <ValidationIcon
+                    id='tooltip-validation'
+                    errorMessages={errors}
+                  />
+                </Grid>
+              )}
+            </Grid>
           </Grid>
-        </Grid>
+          {description ?? <FormHelperText>{description}</FormHelperText>}
```
sdirix, line 100:
```suggestion
          {description && <FormHelperText>{description}</FormHelperText>}
```

SaSteffen:
> We need to use `&&` instead of `??` as `??` short circuits and will use the `description` as is and not with the `FormHelperText`.
> 
Sorry, of course.

> Is it possible to search for the `FormHelperText` element in the tests too?

I checked the DOM and it does not generate any element around it, even if the FormHelperText has an id attribute. If you prefer, I can put it in a div and use that one in the tests.

sdirix:
> I checked the DOM and it does not generate any element around it, even if the FormHelperText has an id attribute. If you prefer, I can put it in a div and use that one in the tests.

I used nested selectors to check for the respective elements. I also added tests checking for the non-existence of them in case there is no description.

SaSteffen:
> > I checked the DOM and it does not generate any element around it, even if the FormHelperText has an id attribute. If you prefer, I can put it in a div and use that one in the tests.
> 
> I used nested selectors to check for the respective elements. I also added tests checking for the non-existence of them in case there is no description.

That leaves nothing else to be done, right? Thank you for the support.

sdirix (approved):

sdirix:

Issue #1390, opened 2019-05-31
Required is false when making specific field required based on condition


sasmangulyan-vineti:
**Describe the bug**
When making the specific field required based on some condition the required property is still false.

**To Reproduce**
1. Schema contains something conditional
`{
  type: "object",
  properties: {
    b: {
      type: "boolean"
    },
    c: {
      type: "string",
      minLength: 1
    }
  },
  if: {
    properties: {
      b: { enum: [false] }
    }
  },
  then: {
    required: ["c"]
  }
}`
2. When making the "b" from true to false (for example checking and unchecking the box), required property of ''c'' which has the renderer of MaterialInputControl will not be changed and will remain false.

**Expected behavior**
After a conditional action the required property of MaterialInputControl to become true.

**Browser (please complete the following information):**
 - Chrome 74.0.3729.157 (Official Build) (64-bit)
 - Firefox 66.0.5 (64-bit)

**Used Setup (please complete the following information):**
 - react
 - material

eneufeld:
Thank you for your report. 
I just double checked, the if/then/else is working fine.
We don't set the required property at all in the renderers.
I adapted the title because of this.

We happily would accept a contribution for this.

stevenmckinnon:
Hey @eneufeld, I see this has been moved to the backlog, are there no plans to fix this soon? Having it in would be a big win for me 👍🏻

sdirix:
@stevenmckinnon In general I think setting the `required` prop as suggested by you in #1695 is a good idea and we should implement it in one of the next versions (i.e. `next`), however it's not on our priority shortlist. You're welcome to open a PR to fix the issue, and we'll definitely have a look at that.

The requested feature here relates to `if/then/else` and will be much harder to support. As the use case is rather niche we have no plans to support this in the near future. If you'd like to contribute this feature we should discuss how the optional `required` can be detected and evaluated efficiently.

In case you need a solution for right now you can always just write your own custom renderers which set the required prop appropriately.

You might also be interested in our support packages which includes prioritization and development of requested features.

stevenmckinnon:
Good to know, thanks @sdirix!

nmaier95:
Hey there, we stumbled upon the exact same issue of properties becoming required via some conditions. Whether that is directly in the "root" schema or in some nested anyOf|oneOf|allOf section. And unfortunately the renderers property `control.required` do not get updated if the conditions are met.
[...]

PR #2228, opened 2023-12-13
Fix conditionally required property not applying #1390

gatanasov-asteasolutions:
This PR fixes the long lasting problem of having conditionally required properties in the schema that do not show as required in the UI. It addresses issue #1390.


packages/core/src/util/renderer.ts
```diff
@@ -80,11 +82,284 @@ import {
...
+) => {
+  if (has(propertyCondition, 'const')) {
+    return (
+      has(data, property) && data[property] === get(propertyCondition, 'const')
+    );
+  } else if (has(propertyCondition, 'enum')) {
+    return (
+      has(data, property) &&
+      (get(propertyCondition, 'enum') as unknown[]).includes(data[property])
+    );
```
sdirix, line 101: These will only work when values are primitives. If a value is an object these comparisons will fail. We should probably use lodash's `isEqual` here to cover the more complex use cases too.

packages/core/src/util/renderer.ts
```diff
@@ -80,11 +82,284 @@ import {
...
+  data: Record<string, unknown>
+): boolean => {
+  const propertiesConditionSchema = get(schema, 'if');
+
+  const condition = evaluateCondition(propertiesConditionSchema, data);
+
+  const ifInThen = has(get(schema, 'then'), 'if');
+  const ifInElse = has(get(schema, 'else'), 'if');
+  const allOfInThen = has(get(schema, 'then'), 'allOf');
+  const allOfInElse = has(get(schema, 'else'), 'allOf');
```
sdirix, line 251: This seems a bit too hardcoded. I would prefer if the logic could be generalized. For example `anyOf` and `oneOf` are not checked here.

packages/core/src/util/renderer.ts
```diff
@@ -80,11 +82,284 @@ import {
 } from '../i18n/arrayTranslations';
 import { resolveSchema } from './resolvers';
 import cloneDeep from 'lodash/cloneDeep';
+import { has } from 'lodash';
+import { all, any } from 'lodash/fp';
```
sdirix, line 86: We use the modular imports of lodash, see line 83,84.

sdirix (commented):
Thanks for this interesting contribution!!

Technically speaking, what this PR is missing is a set of unit tests, testing the new functionality in detail.

Conceptionally this is breaking new grounds for JSON Forms as this introduces dynamic evaluation mechanism which we do not do yet. This is potentially expensive and will be executed with every render call. I'm wondering whether we should
 - make this behavior optional (default: false), and/or
 - introduce a generic prop-adaption mechanism. Then this "advanced" `required` analysis could be manually configured by the user if they need it.
    - Of course then we could already provide this adapter for them.

Once we introduce this, there will be a lot of follow up requests. For example it's a bit arbitrary that we only do this for `required` and not for other attributes too, e.g. `title`. So I'm leaning a bit to the generic prop-adaption mechanism which would allow for arbitrary behavior while keeping JSON Forms itself slim.

What do you think?

gatanasov-asteasolutions:
> I'm wondering whether we should
> 
> * make this behavior optional (default: false), and/or
> * introduce a generic prop-adaption mechanism. Then this "advanced" `required` analysis could be manually configured by the user if they need it.
>   
>   * Of course then we could already provide this adapter for them.
> 
> Once we introduce this, there will be a lot of follow up requests. For example it's a bit arbitrary that we only do this for `required` and not for other attributes too, e.g. `title`. So I'm leaning a bit to the generic prop-adaption mechanism which would allow for arbitrary behavior while keeping JSON Forms itself slim.
> 
> What do you think?

I agree that these checks are potentially expensive, but I think this is the expected behavior, at least for `required`, since `AJV` marks the fields as required while JSON Forms did not. The option to allow the user to choose if they want to have this checks performed sounds like a good way of improving the performance when they are not needed. The problem I am seeing right now is that I am not sure where should we put the "toggle" for this behavior. Here are some of my thoughts:

- adding new keyword `conditionallyRequired` that marks the properties to be checked dynamically;
- adding new schema property `allowRecursive` that marks the properties inside the subschema to be checked dynamically (potentially not only `required` will be affected);
- some other form of top-level declaration for dynamic checks.

I am open to other suggestions, since I can not think of anything better.

packages/core/src/util/renderer.ts
gatanasov-asteasolutions: I see, I have changed it in my latest commit.

packages/core/src/util/renderer.ts
gatanasov-asteasolutions: You are right, there were other places as well where it was needed to compare the values with `isEqual`. It is fixed in my latest commit.

sdirix:
Hi @gatanasov-asteasolutions, we were currently busy with the 3.2 release. We will discuss this proposal and what we think what the best way forward is within the team in the upcoming weeks and let you know.

In the mean time:
- If you would like to contribute the "prop-adaption" mechanism to accomplish this feature for your renderers in a convenient way, then feel free to do so
- If you would like to already consume your contributed mechanism, then you can do so via custom renderers. In there you can call the current bindings of JSON Forms and then just apply the "conditionally required property" evaluation on top.

gatanasov-asteasolutions:
Hi @sdirix can you please review these changes again? The fork is updated to include the latest changes of the master branch and new changes were introduced in order to support conditionally required validations for array controls. I am also looking forward to your team's decision on the future of this proposal and if I can assist you in any way.

lucas-koehler:
Thanks again for the contribution :heart: 
Closing this now as it did not have any activity in over a year and the original concept of the PR unfortunately does not fit JSON Forms as described in https://github.com/eclipsesource/jsonforms/pull/2228#issuecomment-2337711180

Issue #2138, opened 2023-05-16
Support Angular w/ Material  v16


Scarylaggy:
### Is your feature request related to a problem? Please describe.

We would like to use Angular for an Application in the version 16 as it is the latest version of angular and stumbled upon this repo for a great (in version 14) working form system. As it looks like the Angular package is still worked on, in version 12 which is quite old by now and should be updated due to some vulneribiltys.

### Describe the solution you'd like

Update the dependencys used for the angular and angular material packages to latest standards.

### Describe alternatives you've considered

Not using jsonforms if it is not possible.

### Framework

Angular

JonasDev17:
Another alternative would be Formly: https://github.com/ngx-formly/ngx-formly
Made specifically for Angular whcich means it has a littel more flexibility since they can upgrade their typescript without having to worry about breaking any other packages.

Scarylaggy:
Yes, their product does look very good and i tried it but unfortunatly they got some issues with Angular v16 too.
Which is quite understandable beeing how young it is at this point.

chladog:
I also wanted to implement it in my project as the project looks quite practical.
To clarify on this topic, the cause is **lack of compability with Ivy compiler**. In Angular version **16 the Angular Compatibility Compiler (ngcc) has been removed** (which allowed older non-ivy libraries to be compiled with Ivy). So that's why it was working until now.

sdirix:
In general we would like to support the latest Angular version when possible. This is complicated by the fact that our Angular builds [are very old school and need to be modernized](https://github.com/eclipsesource/jsonforms/issues/1546).

The community driven PR to [support Angular 15 sadly got stuck](https://github.com/eclipsesource/jsonforms/pull/2130). If someone of the community wants to contribute Angular 15/16 support then we will gladly accept that.

We are currently working on modernizing the repo with `pnpm` etc. so it probably makes sense to wait for this.

Altair-Bueno:
Wondering if it would made more sense to just use [Web components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) instead of maintaining integrations for each framework. There are projects following this direction, such as [material web](https://github.com/material-components/material-web), which aims to be compatible with multiple frameworks with the same codebase. 

- Benefits: Less packages and code to maintain. The ability to easily use this package with plain HTML without much hassle.
- Disadvantages: Somewhat limited props, based on HTML attributes and events. Custom renderers are possible but only if they are based on Web components
[...]

Issue #2224, opened 2023-12-06
Move to Node 18


sdirix:
As Node 16 is EOL we should update the developer tooling to at least Node 18.

### ToDo

Update documentation and CI jobs to Node 18 and fix all errors currently encountered

PR #2212, opened 2023-11-14
Angular v15

JBBianchi:
(Reopening #2210 after messing up with git.)

Migrated Angular/Angular Material to v15 (dropped compatibility with previous versions. Could work, I didn't check.)

I didn't manage to reconfigure the test environment yet for angular-material.

Any help is more than welcome! <3

- [x] pnpm run build
- [x] pnpm run doc
- [x] pnpm run test (fails: trying to "rewire" tests for angular-material)
- [x] pnpm run test-cov (same as above)
- [x] pnpm run lint (will do when the rest is fixed)
- [x] run example
- [x] Tested in angular seed project v15

Next steps:
- Remove support for deprecated `@angular/flex-layout`
- Add support for Angular v16 and v17

(Dev envs: Win 10/11 - NodeJS 16.14.2 & 18.18.2)

Related to

- https://github.com/eclipsesource/jsonforms/pull/2130
- https://github.com/eclipsesource/jsonforms/pull/2125 ?
- https://github.com/eclipsesource/jsonforms/pull/2117
- https://github.com/eclipsesource/jsonforms/pull/1994
- https://github.com/eclipsesource/jsonforms/issues/2116
- https://github.com/eclipsesource/jsonforms/issues/1546
- https://github.com/eclipsesource/jsonforms/issues/1993
- https://github.com/eclipsesource/jsonforms/issues/2138

@JonasDev17 @sdirix if you have some time to have a look, it would be great. Angular is a great framework often used in companies, it would be sad to drop support for it.

PR #2229, opened 2023-12-14
Improved Angular compatibility

JBBianchi:
In the current stage, `@jsonforms/angular-material` still relies on `@angular/flex-layout`.

This package has been deprecated and isn't compatible with Angular v16+, therefore `@jsonforms/angular-material` would be stuck to Angular v15.

This PR removes the dependency to `@angular/flex-layout` by replacing the attributes with plain CSS instead.

Doing so, `@jsonforms/angular-material` becomes compatible with Angular v16 and v17 (at first sight).

For testing purposes, I forced upgrade the `jsonforms-angular-seed` projects to both v16 and v17:
- https://github.com/JBBianchi/jsonforms-angular-seed/tree/v15-to-v16
- https://github.com/JBBianchi/jsonforms-angular-seed/tree/v16-to-v17

I tried building the packages as submitted in this PR and running them in the above projects and it **seems** to work fine (it displays, I can remove one of the two items and I can update a text field. I didn't test more than that).

Related to
- https://github.com/eclipsesource/jsonforms/issues/2138


JBBianchi:
(Follow up of #2212)

JBBianchi:
I might have messed up the [pnpm-lock.yaml](https://github.com/eclipsesource/jsonforms/pull/2229/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb) ...

Is there a way for pnpm to use Angular v15 even if the package is compatible with higher versions ? (compile vs usage)

gexclaude:
@JBBianchi really appreciate your efforts!

JBBianchi:
We could actually use Angular v16 for the build process, which will allow to be compliant with #2224 .

Should I add this to this PR as well ? It's basically just changing the pnpm.overrides and removing the `entryComponents` config from the tests so the impact is pretty low.

@sdirix @lucas-koehler WDYT ?

lucas-koehler:
Hi @JBBianchi,
first of all thanks for the PR and your continued effort on improving the angular renderer set :heart: 
I see no harm in using Angular 16 for the build process already. So gladly go ahead with that :)

JBBianchi:
> Hi @JBBianchi, first of all thanks for the PR and your continued effort on improving the angular renderer set ❤️ I see no harm in using Angular 16 for the build process already. So gladly go ahead with that :)

For ng16, I'll hold my horses because the example doesn't seem to run. I'll need to investigate a bit more even though all lights seemed green at first.

JBBianchi:
@lucas-koehler @sdirix the PR should be ready for review.

packages/angular-material/src/controls/range.renderer.ts
```diff
@@ -33,10 +33,11 @@ import { isRangeControl, RankedTester, rankWith } from '@jsonforms/core';
 @Component({
   selector: 'RangeControlRenderer',
   template: `
-    <div fxFlex fxLayout="column" [fxHide]="hidden">
+    <div [ngStyle]="{ display: hidden ? 'none' : '' }" class="range-control">
       <label class="mat-caption" style="color:rgba(0,0,0,.54)">{{
         label
       }}</label>
+      <!-- TODO: The 'tickInterval' property no longer exists -->
```
lucas-koehler, line 40: Is there a problem with the current solution (using `showTickMarks`). If it currently works fine, I don't think we need this TODO

packages/angular-material/src/layouts/group-layout.renderer.ts
```diff
@@ -34,19 +34,34 @@ import { JsonFormsAngularService } from '@jsonforms/angular';
...
     </mat-card>
   `,
+  styles: [
+    `
+      .group-layout {
+        display: flex;
+        flex-direction: column;
+        gap: 16px;
+      }
+      .group-layout > * {
```
lucas-koehler, line 60: Should this really apply to all elements  or only `div`s? I ask because `mat-card-title` did not have `fxFlex` before

packages/angular-material/test/group-layout.spec.ts
```diff
@@ -46,7 +45,7 @@ describe('Group layout', () => {
   beforeEach(() => {
     fixture = beforeEachLayoutTest(GroupLayoutRenderer, {
       declarations: [LayoutChildrenRenderPropsPipe, MatCard, MatCardTitle],
-      imports: [FlexLayoutModule],
+      imports: [],
```
lucas-koehler, line 48: can this be removed now that it's empty?

packages/angular-material/test/horizontal-layout.spec.ts
```diff
@@ -45,7 +44,7 @@ describe('Horizontal layout', () => {
   beforeEach(() => {
     fixture = beforeEachLayoutTest(HorizontalLayoutRenderer, {
       declarations: [LayoutChildrenRenderPropsPipe],
-      imports: [FlexLayoutModule],
+      imports: [],
```
lucas-koehler, line 47: can this be removed now that it's empty?

packages/angular-material/tsconfig.spec.json
```diff
@@ -11,11 +11,5 @@
   "include": [
     "**/*.spec.ts",
     "**/*.d.ts"
-  ],
-  "angularCompilerOptions": {
-    "compilationMode": "partial",
-    "strictMetadataEmit": true,
-    "generateCodeForLibraries": false,
-    "skipTemplateCodegen": true
```
lucas-koehler, line 19: Why did you remove this? Are these all defaults?

lucas-koehler (commented):
Hi thanks again for the PR :heart: 
The changes already look pretty good to me :)
I have some minor comments/questions inline.

packages/angular-material/tsconfig.spec.json
JBBianchi: I must admit I don't think they are necessary anymore. If I'm not mistaken, the only one still present when generating a lib with the angular cli now is `generateCodeForLibraries` for the production build.

In doubt, I'll add them again.

packages/angular-material/test/horizontal-layout.spec.ts
JBBianchi: sure

packages/angular-material/test/group-layout.spec.ts
JBBianchi: sure

packages/angular-material/src/layouts/group-layout.renderer.ts
JBBianchi: I don't think it have an impact but to be on the safe side I'll specify div indeed.

packages/angular-material/src/controls/range.renderer.ts
JBBianchi: Indeed, makes sense.

packages/angular-material/src/layouts/group-layout.renderer.ts
```diff
@@ -34,19 +34,34 @@ import { JsonFormsAngularService } from '@jsonforms/angular';
...
         "
-        fxFlex
       >
         <jsonforms-outlet [renderProps]="props"></jsonforms-outlet>
       </div>
     </mat-card>
   `,
+  styles: [
+    `
+      .group-layout {
```
lucas-koehler, line 55: I think the .group-layout class misses a `padding: 16px;`. Before there was some distance to the borders of the group. This is gone in this PR. The padding in before this was 16px.

lucas-koehler (changes requested):
Hi the code changes LGTM. I had a look at the example. It looked pretty good to me. I only noticed that the group layout is now a bit cramped due to missing padding.

packages/angular-material/src/layouts/group-layout.renderer.ts
JBBianchi: It's an artifact from the previous PR but indeed, `.mat-card` had a padding of 16px before. I'll fix it.

lucas-koehler:
@JBBianchi Thanks for the update. Do you know whether the code will still work for Angular 15 when it's built with Angular 16 ?
If yes that's fine for me :) Otherwise, we can also remove compatibility to Angular 15.

lucas-koehler (changes requested):
In the seed, I tested the changes with 15, 16 and 17. The seed did not compile with 15 but worked with 16 and 17. Thus, I suggest removing compatibility to 15. It is end of life in less than half a year anyway.

Also I noticed another bug, the `packages/angular-material/src/controls/range.renderer.ts` does no longer have functioning data binding. I think this might already have been introduced with the Angular 15 support PR when the value and onChange properties were removed in `packages/angular-material/src/controls/range.renderer.ts`

JBBianchi:
> In the seed, I tested the changes with 15, 16 and 17. The seed did not compile with 15 but worked with 16 and 17. Thus, I suggest removing compatibility to 15. It is end of life in less than half a year anyway.
> 
> Also I noticed another bug, the `packages/angular-material/src/controls/range.renderer.ts` does no longer have functioning data binding. I think this might already have been introduced with the Angular 15 support PR when the value and onChange properties were removed in `packages/angular-material/src/controls/range.renderer.ts`

Thanks for testing in the seed v15, that's unfortunate but indeed, support could be dropped.

I'll try to have look during next week if I have some time.

JBBianchi:
It seems indeed that a library built with ng 16 is not compatible with ng 15 because the signature of the `@Input()` directive changed.

For the range control, the binding of the change event was "cleaned out" by mistake in this [commit](https://github.com/JBBianchi/jsonforms/commit/e393bfb46ada53685a46a2a2d01cd9408ad77167).

Next commit will remove compatibility with ng 15 and rebind a the event to the range control.

package.json
```diff
@@ -48,5 +48,28 @@
...
+      "@angular-eslint/eslint-plugin-template": "^16.0.0",
+      "@angular-eslint/schematics": "^16.0.0",
+      "@angular-eslint/template-parser": "^16.0.0",
+      "@angular/compiler": "^16.0.0",
+      "@angular/compiler-cli": "^16.0.0",
+      "@angular/platform-browser-dynamic": "^16.0.0",
+      "@ngtools/webpack": "^16.0.0",
+      "ng-packagr": "^16.0.0",
+      "zone.js": "^0.13.0"
+    }
```
lucas-koehler, line 73: Do we still need these overrides now that we removed Angular 15 compatibility?

lucas-koehler (commented):
Hi @JBBianchi thanks for the updates :) Changes look pretty good to me now. I have just one question regarding the pnpm overrides inline.
I also updated the lock file.

package.json
JBBianchi: I'm not 100% sure. If I'm not mistaken, as the packages declare a compatibility with Angular 16 and 17 (and maybe 18 in the future), pnpm will install the highest version compatible by default, so v17. So we need the overrides to say "build with v16 even if the lib is compatible with v17".

sdirix: Can we use Angular `16` in the `devDependencies` and not `16 || 17`? As we don't bundle Angular the `peerDependencies` version is the important one for the consumer and could still be `16 || 17` even of the `devDependencies` only declare `16`.

JBBianchi: @sdirix thanks for your input, that's better indeed.

lucas-koehler (approved):
Hi @JBBianchi , the PR looks good to me now. I updated all dependencies and added overrides for vue because its latest version doesn't work. I added issue https://github.com/eclipsesource/jsonforms/issues/2234 to fix this later.

Thanks again for your effort on the Angular packages :heart:

packages/angular-material/src/other/master-detail/master.ts
```diff
@@ -140,9 +143,6 @@ export const removeSchemaKeywords = (path: string) => {
         top: 0;
         right: 0;
       }
-      .hide {
-        display: none;
```
stanisev, line 144: Why is this deleted?

JBBianchi: It might be an error when factoring out `@angular/flex-layout` with `[ngStyle]="{ display: hidden ? 'none' : '' }"` instead of `[fxHide]="hidden"`. Do you encounter any problem ?

stanisev: Yeah right, that's why I am writing. The delete button is always shown. We created a custom listWithDetail renderer to counter this problem. Is there a better solution? Or can you restore the deleted CSS class?

JBBianchi: I'm not a maintainer of this repo but I'd assume you can open an issue explaining the problem and them submit a PR to solve it. Then maintainers can maybe accept the PR and push it in the next release.

sdirix: Happy to review and merge a PR

stanisev: I've created the issue: https://github.com/eclipsesource/jsonforms/issues/2327
Should you give me permission in order to open a MR?

lucas-koehler: @stanisev Thanks for creating the issue :) You can gladly open a MR anytime you want.

lucas-koehler: Hi @stanisev , did you try to push the branch directly to this repository? This is not possible because you do not have write permissions. You need to fork the repository and push the branch to your fork. Then, you should be able to open a MR against this repository.

PR #2230, opened 2023-12-15
fix: remove @jsonforms/angular-test peerDependency

sdirix:
The @jsonforms/angular-material package specified
@jsonforms/angular-test as peerDependency. This is now removed.


lucas-koehler (approved):
LGTM, thanks :)

PR #2240, opened 2024-01-05
Adapt issue templates to vue 2 removal

lucas-koehler:


sdirix:
Good fix. I would like to suggest to also remove the Browser entry. This information very rarely provides a benefit but is annoying for the reporter to enter.

lucas-koehler:
@sdirix good point, i removed it .

sdirix (approved):

PR #2241, opened 2024-01-05
dev: typescript 4.9.5, TS target ES2021, TS module ES2020, typedoc ~0.25.3

lucas-koehler:
- Upgrade all packages to use  typescript `~4.9.5`
- Update typedoc version to `~0.25.3` and remove custom typedoc theme
- Adapt @mui/icons-material imports in material-renderers to work with later typescript versions. The used imports broke starting with 4.2.4
- Fix typescript target from ESNext to ES2021 because
- Fix typescript module from ESNext to ES2020

Remove ESNext from typescript configurations because it leads to changes in output when updating the typescript version.
For `target` ESNext is explicitly discouraged in the typescript docs. E.g. it broke vue linting when only upgrading the typescript version.


packages/material-renderers/src/complex/MaterialTableControl.tsx
```diff
@@ -57,9 +57,11 @@ import {
...
   ArrayTranslations,
 } from '@jsonforms/core';
-import DeleteIcon from '@mui/icons-material/Delete';
-import ArrowDownward from '@mui/icons-material/ArrowDownward';
-import ArrowUpward from '@mui/icons-material/ArrowUpward';
+import {
+  Delete as DeleteIcon,
+  ArrowDownward,
+  ArrowUpward,
+} from '@mui/icons-material';
```
sdirix, line 64: Are these icon changes necessary?

lucas-koehler: For me this was necessary. With updating the typescript version, the old imports no longer worked. This was the case during the build as well in the IDE. Curiously that already happens when going from TS 4.2.3 to 4.2.4. I looked into the changelog and there as a change about resolving something. So that could be the difference.

sdirix (approved):

Issue #2242, opened 2024-01-06
Performance Issue for Custom Table renderer component


howdyAnkit:
### Describe the bug

The given scenario highlights performance issues caused by excessive function calls, leading to latency problems, particularly in a nested object scenario within a custom Angular table renderer. To improve performance, consider optimizing the function calls and reducing unnecessary re-render.

Here's a a suggestion how I approached the problem using Angular's PurePipe to optimize function calls:

PurePipe: Utilize an Angular PurePipe to reduce unnecessary re-render of complex logic in the template.

while i was debugging I noticed that even while we are hovering certain elements this function was called unnecessarily multiple times even though when we are not making interacting with UI or performing any action.

By incorporating PurePipe in your Angular component, you can significantly optimize function calls and improve performance, especially in scenarios involving complex nested objects or custom renderers.

https://github.com/eclipsesource/jsonforms/blob/master/packages/angular-material/src/library/other/table.renderer.ts

### Expected behavior

The getProps function should only be getting called when a user is performing any action releated to the Table to prevent memory leaks.

### Steps to reproduce the issue

1) Add a custom renderer(https://github.com/eclipsesource/jsonforms/blob/master/packages/angular-material/src/library/other/table.renderer.ts)
2) Make sure you have lot of nested data and dataset is huge. 
3) place a console statement inside the getProp function and Observe the statement getting triggered on one click of add multiple times same with delete and also while you are hovering any elements on the table.

### Which Version of JSON Forms are you using?

v3.1.0

### Framework

Angular

### RendererSet

Material

howdyAnkit:
Please reproduce it once and let me know if I can attach a PR to fix it.

lucas-koehler:
Hi @howdyAnkit , thank you for raising this issue.
Would you like to contribute a fix for this? We would review and accept a contribution to improve this.

howdyAnkit:
Hi @lucas-koehler yes please assign it to me. will add a PR

PR #2243, opened 2024-01-08
Fix memory leak issue

howdyAnkit:
Fixes issue for #2242


howdyAnkit:
Hi @lucas-koehler please review and let me know if you have any suggestions or any changes required thanks

sdirix:
I'm not familiar with pipes. Can you quickly explain for the change when and how now a rerender is triggered?

Will the table rerender when seemingly unrelated changes happen? For example a property outside the table is changed and a `SHOW rule` associated with the `table` now evaluates differently. Will the table be updated, i.e. shown/hidden then?

howdyAnkit:
Hi @sdirix thanks for reviewing.

Can you quickly explain for the change when and how now a re-render is triggered?

The changes should be triggered normally as compared to previously the function was getting triggered aggressively because the function was being called on the html itself. but now since I have changed with Pipes. Pipes by themselves do not trigger re-renders or DOM updates. They perform transformations based on input data and are primarily used for display purposes and perform transformations on data without triggering re-renders. 

Considering previous example where the function was called directly and it was detecting the changes on hover as-well instead now if there's any new change it will definitely trigger a re-render it's just that it will be triggered once and the no of time's you interact or the no of time it's been called. I'd still like to see if you have any examples that could possibly be a better way to validate the code instead of assumptions.

howdyAnkit:
Hi @lucas-koehler and @sdirix any updates on the above. any inputs required from my end?

sdirix:
Hi @howdyAnkit, we were busy with the 3.2. in the last two weeks, therefore the late reply.

From what you explained, I think the approach is sound.

This is the scenario to be tested:

JSON Schema
```
{
  "type": "object",
  "properties": {
    "enabled": {
      "type": "boolean"
    },
    "people": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
          },
          "age": {
            "type": "number"
          }
        }
      }
    }
  }
}
```

Ui Schema
```
{
  "type": "VerticalLayout",
  "elements": [
    {
      "type": "Control",
      "scope": "#/properties/enabled",
      "label": "Is Enabled"
    },
    {
      "type": "Control",
      "scope": "#/properties/people",
      "label": "People",
      "options": {
        "detail": {
          "type": "VerticalLayout",
          "elements": [
            {
              "type": "Control",
              "scope": "#/items/properties/name",
              "label": "Name"
            },
            {
              "type": "Control",
              "scope": "#/items/properties/age",
              "label": "Age"
            }
          ]
        }
      }
    }
  ]
}
```
The table should still be enabled or disabled based on the enabled property.

I would also recommend to rebase the PR on the latest state of master to potentially fix the seemingly unrelated test errors.

howdyAnkit:
Hi @sdirix thanks for replying and getting back PFA working repo to validate the schema's given above (https://github.com/howdyAnkit/fix-2242). Everything looks working fine to me. i'll rebase the branch and push again and if any other changes required please let me know. 

Thanks Ankit

howdyAnkit:
Also I have added a last commit to include the pipe in module folder to make it work or else it will throw errors please do let me know if it needs to be added few other places aswell. Thanks for checking : )

howdyAnkit:
Hi @sdirix thanks for reviewing. did it passed this time?

howdyAnkit:
I can still see it failing 💔. the only problem in my view was the missing import in module i'd like to hear you're suggestion if you have.

sdirix (approved):

howdyAnkit:
Thanks @sdirix for adding the imports and fixing this issue ❤️.

PR #2248, opened 2024-01-09
chore: use npm provenance

sdirix:
Configures the npm provenance mechanism for the Github release workflow.

The npm provenance assures consumers of JSON Forms that the libraries available on npmjs were actually produced by the JSON Forms project.

See https://docs.npmjs.com/generating-provenance-statements


lucas-koehler (approved):
LGTM, thanks :)

PR #2249, opened 2024-01-11
Vanilla TextCell password

adammation:
The input type for a TextCell in the vanilla-renderers was 'text' and could not be changed.

This change keeps the default 'text' type but allows the type to be changed via an option "type"

`"options": {
              "format": "password"
            }`

See also:
https://jsonforms.discourse.group/t/input-type-with-vanilla-renderers/272

### DIFF
--- a/packages/vanilla-renderers/src/cells/TextCell.tsx
+++ b/packages/vanilla-renderers/src/cells/TextCell.tsx
@@ -50,7 +50,7 @@ export const TextCell = (props: CellProps & VanillaRendererProps) => {
   const appliedUiSchemaOptions = merge({}, config, uischema.options);
   return (
     <input
-      type='text'
+      type={appliedUiSchemaOptions.type ? appliedUiSchemaOptions.type : 'text'}
       value={data || ''}
       onChange={(ev) =>
         handleChange(path, ev.target.value === '' ? undefined : ev.target.value)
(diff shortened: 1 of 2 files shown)


sdirix (changes requested):
Hi! Thanks for the contribution :heart: 

In the React Material UI renderers we already support `format: 'password'` as the indicator to render a password field. Would this work for you too? I would prefer to be consistent between renderer sets.

adammation:
> Hi! Thanks for the contribution ❤️
> 
> In the React Material UI renderers we already support `format: 'password'` as the indicator to render a password field. Would this work for you too? I would prefer to be consistent between renderer sets.

Sounds great, changed to `format: "password"`

packages/vanilla-renderers/src/cells/TextCell.tsx
```diff
@@ -50,7 +50,9 @@ export const TextCell = (props: CellProps & VanillaRendererProps) => {
   const appliedUiSchemaOptions = merge({}, config, uischema.options);
   return (
     <input
-      type='text'
+      type={
+        appliedUiSchemaOptions.format ? appliedUiSchemaOptions.format : 'text'
+      }
```
sdirix, line 55:
```suggestion
      type={appliedUiSchemaOptions.format === 'password' ? 'password' : 'text'}
```

sdirix (changes requested):

packages/vanilla-renderers/src/cells/TextCell.tsx
sdirix: I would prefer only passing through the password as in React Material. Other types might be incompatible. I would have applied the change myself but maintainer editing was turned off for this PR

sdirix:
Closed in favor of https://github.com/eclipsesource/jsonforms/pull/2254. Thanks for the contribution :heart:

adammation:
Thanks @sdirix !

PR #2254, opened 2024-01-22
Vanilla textcell password

sdirix:
Follow up of https://github.com/eclipsesource/jsonforms/pull/2249


sdirix:
squash merged commits

PR #2256, opened 2024-01-24
fix(material): consistently use outlined input variant

LukasBoll:
Use outlined input variant as the default to match Material UI

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,43 @@
 # Migration guide
 
+## Migrating to JSON Forms 3.2
+
+### Material Renderers using Outlined Inputs
+
+JSON Forms now uses the `outlined` input variant as the default, aligning with the default style of Material UI since version 5.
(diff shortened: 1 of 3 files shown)


MIGRATION.md
```diff
@@ -1,5 +1,43 @@
 # Migration guide
 
+## Migrating to JSON Forms 3.2
+
+### Material Renderers using Outlined Inputs
+
+JSON Forms has been updated to use the `outlined` input variant as the default, aligning more closely with the default style of Material UI.
```
sdirix, line 7:
```suggestion
JSON Forms now uses the `outlined` input variant as the default, aligning with the default style of Material UI since version 5.
```

MIGRATION.md
```diff
@@ -1,5 +1,43 @@
...
+To use the `standard` input variant, as in previous versions of JSON Forms, it is possible to provide JSON Forms with a theme provider:
```
sdirix, line 8:
```suggestion
If you would like to use the `standard` input variant, as was default in previous versions of JSON Forms, then this can be accomplished using the Material UI `ThemeProvider`:
```

sdirix (approved):
Thanks!

PR #2257, opened 2024-01-24
dev: fix the React Material and Vanilla example dev setup

sdirix:
We recently updated Typescript. The new version is incompatible to our (pretty old) webpack dev-server setup.

We now update Webpack throughout the whole repository, fixing the webpack dev-server setup in the process.


LukasBoll (approved):
Looks good to me, all dev applications are working again!

Issue #2238, opened 2024-01-03
Add a required class to labels


vanillajonathan:
### Is your feature request related to a problem? Please describe.

Required labels have a asterisk `*` which cannot be styled. I would like like to change the asterisk to be red, or use a icon to indicate that the input is required, or have some text such as "(necessario)".

### Describe the solution you'd like

To inputs that are required add a `required` or `form-required` class to the label.

This way it can be styled using CSS:
```css
label.required::after {
  color: red;
  padding-left: 5px;
  content: "*"
}
```

### Describe alternatives you've considered

Fork the Vanilla renderer.

### Framework

Vue 3

### RendererSet

Vanilla

sdirix:
Hi @vanillajonathan, Thanks for the suggestion! Definitely makes sense to me. The required asterisk should reside in an own customizable element so that it becomes easily styleable.

PR #2274, opened 2024-02-09
feat: create own element for asterisk and add class in vue vanilla controlWrapper 

davewwww:
regarding issue #2238

### DIFF
--- a/packages/vue-vanilla/src/controls/ControlWrapper.vue
+++ b/packages/vue-vanilla/src/controls/ControlWrapper.vue
@@ -1,7 +1,11 @@
 <template>
   <div v-if="visible" :id="id" :class="styles.control.root">
-    <label :for="id + '-input'" :class="styles.control.label">
-      {{ computedLabel }}<span v-if="showAsterisk" :class="styles.control.asterisk">*</span>
+    <label
+      :for="id + '-input'"
+      :class="[styles.control.label, required ? styles.control.required : '']"
+    >
+      {{ label }}
+      <span v-if="showAsterisk" :class="styles.control.asterisk">*</span>
     </label>
     <div :class="styles.control.wrapper">
       <slot></slot>
@@ -13,7 +17,7 @@
 </template>
 
 <script lang="ts">
-import { isDescriptionHidden, computeLabel } from '@jsonforms/core';
+import { isDescriptionHidden } from '@jsonforms/core';
 import { defineComponent, PropType } from 'vue';
 import { Styles } from '../styles';
 import { Options } from '../util';
@@ -74,16 +78,9 @@ export default defineComponent({
         !!this.appliedOptions?.showUnfocusedDescription
       );
     },
-    computedLabel(): string {
-      return computeLabel(
-        this.label,
-        this.required,
-        true
-      );
-    },
-    showAsterisk():  boolean {
+    showAsterisk(): boolean {
       return this.required && !this.appliedOptions?.hideRequiredAsterisk;
-    }
+    },
   },
 });
 </script>
--- a/packages/vue-vanilla/src/controls/ControlWrapper.vue
+++ b/packages/vue-vanilla/src/controls/ControlWrapper.vue
@@ -1,7 +1,7 @@
 <template>
   <div v-if="visible" :id="id" :class="styles.control.root">
     <label :for="id + '-input'" :class="styles.control.label">
-      {{ computedLabel }}
+      {{ computedLabel }}<span v-if="showAsterisk" :class="styles.control.asterisk">*</span>
     </label>
     <div :class="styles.control.wrapper">
       <slot></slot>
@@ -78,9 +78,12 @@ export default defineComponent({
       return computeLabel(
         this.label,
         this.required,
-        !!this.appliedOptions?.hideRequiredAsterisk
+        true
       );
     },
+    showAsterisk():  boolean {
+      return this.required && !this.appliedOptions?.hideRequiredAsterisk;
+    }
   },
 });
 </script>
--- a/packages/vue-vanilla/tests/unit/complex/ObjectRenderer.spec.ts
+++ b/packages/vue-vanilla/tests/unit/complex/ObjectRenderer.spec.ts
@@ -21,6 +21,23 @@ const uischema = {
   scope: '#',
 };
 
+const schemaRequired = {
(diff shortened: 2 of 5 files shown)


packages/vue-vanilla/tests/unit/complex/ObjectRenderer.spec.ts
```diff
@@ -49,4 +66,14 @@ describe('ObjectRenderer.vue', () => {
...
+
+  it('renders a string input with an asterisk', () => {
+    const wrapper = mountJsonForms('a', schemaRequired, uischema);
+    expect(wrapper.find('label span.asterisk').exists()).to.be.true;
+  });
+
+  it('renders a string input but hide asterisk', () => {
+    const wrapper = mountJsonForms('a', schemaRequired, uischemaRequired);
+    expect(wrapper.find('label span.asterisk').exists()).to.be.false;
+  });
```
lucas-koehler, line 78: I think these tests fit better to `StringControlRenderer.spec.ts` because they are not related to the object renderer itself.

lucas-koehler (changes requested):
Hi @davewwww , thanks for your contribution!
The linter complains about the ocde not being formatted as expected. I left a comment about the location of the new tests. please have a look.

In https://github.com/eclipsesource/jsonforms/issues/2238, there is a suggestion to add the asterisk via CSS to be able to customize the `*` itself. Did you consider that?

packages/vue-vanilla/tests/unit/complex/ObjectRenderer.spec.ts
davewwww: but isn't the "required" feature a part of the object?

https://json-schema.org/understanding-json-schema/reference/object#required

davewwww:
> Hi @davewwww , thanks for your contribution! The linter complains about the ocde not being formatted as expected. I left a comment about the location of the new tests. please have a look.
> 
> In #2238, there is a suggestion to add the asterisk via CSS to be able to customize the `*` itself. Did you consider that?

yes, i read that suggestion too. but i thought it would make sense to solve this "add class" issue here at first and then solve the other one in a separate pr. 

Or do you have another suggestion?

packages/vue-vanilla/src/controls/ControlWrapper.vue
```diff
@@ -78,9 +78,12 @@ export default defineComponent({
       return computeLabel(
         this.label,
         this.required,
-        !!this.appliedOptions?.hideRequiredAsterisk
+        true
       );
```
sdirix, line 82: handing over `true` here is the same as just using the `this.label` directly, so in fact we can remove the `computedLabel` completely once we manually manage the asterisk.

packages/vue-vanilla/tests/unit/complex/ObjectRenderer.spec.ts
sdirix: It is conceptually, however the rendering of the "required" is not part of the object renderer. Therefore it should be tested in the respective renderers instead.

davewwww:
> > > Hi @davewwww , thanks for your contribution! The linter complains about the ocde not being formatted as expected. I left a comment about the location of the new tests. please have a look.
> > > In #2238, there is a suggestion to add the asterisk via CSS to be able to customize the `*` itself. Did you consider that?
> > 
> > 
> > yes, i read that suggestion too. but i thought it would make sense to solve this "add class" issue here at first and then solve the other one in a separate pr.
> > Or do you have another suggestion?
> 
> We can do a mix of these:
> 
> * I think it's fine to textually add the asterisk as we don't want to enforce CSS specifics
> * However the whole label should get a required CSS class set if we determine that it's required. This way consumers can style the label with CSS if they want to, including hiding the textual asterisk if they don't like it.

you mean something like that?:

```ts
    <label :for="id + '-input'" :class="[styles.control.label, required ? styles.control.required : '']:">
      {{ label }}
      <span v-if="showAsterisk" :class="styles.control.asterisk">*</span>
    </label>
```

sdirix:
> > We can do a mix of these:
> > 
> > * I think it's fine to textually add the asterisk as we don't want to enforce CSS specifics
> > * However the whole label should get a required CSS class set if we determine that it's required. This way consumers can style the label with CSS if they want to, including hiding the textual asterisk if they don't like it.
> 
> you mean something like that?:
> 
> ```ts
>     <label :for="id + '-input'" :class="[styles.control.label, required ? styles.control.required : '']">
>       {{ label }}
>       <span v-if="showAsterisk" :class="styles.control.asterisk">*</span>
>     </label>
> ```

Looks good to me. What do you think @lucas-koehler?

lucas-koehler (approved):
LGTM now! Thanks @davewwww for the updates and @sdirix  for the insights :)

vanillajonathan:
It doesn't need both an `asterisk` and `required` class, just `required` is fine, then you can add an asterisk with CSS.

```css
label.required::after {
  color: red;
  padding-left: 5px;
  content: "*"
}
```

Here the `content` attribute is set to `"*"` but it could also be changed to `"(required)"` or translated to some other language or changed to use an asterisk icon instead such as:
https://icons.getbootstrap.com/icons/asterisk/
https://fontawesome.com/icons/asterisk

lucas-koehler:
Hi @vanillajonathan , thanks for the insight. We opted for having the required class and rendering an asterisk for the following reasons outlined by @sdirix  above:

>- I think it's fine to textually add the asterisk as we don't want to enforce CSS specifics
>- However the whole label should get a required CSS class set if we determine that it's required. This way consumers can style the label with CSS if they want to, including hiding the textual asterisk if they don't like it.

vanillajonathan:
But now the textual asterisk cannot be changed to a asterisk icon, nor can it be changed to a text string indicating it is required, or use the `:not` CSS pseudo-class to have non-required fields marked as "optional".

https://getbootstrap.com/docs/5.3/examples/checkout/

Having a textual asterisk within an element is not so flexible as opposed to add it with the `content` attribute using CSS.

davewwww:
you can hide the textual asterisk and add you changes to the pseudo after element at the label
```css
label .asterisk {
  display: none;
}
label:not(.required):after {
  content: ' (optional)';
  color: gray;
}
label.required:after {
  content: '*';
  color: red;
}
```

Issue #2156, opened 2023-07-04
Vue3 Vanilla Time & DateTime Renderer do not work as expected


davewwww:
### Describe the bug

Expectation:
I have a Date field (format=date), then I click on a tag in the overlay and the day appears in the Input field and the overlay closes.

Actual situation:
If I have a Time/Datetime field (format=time/datetime) and I click on a time in the overlay, then the time appears in the Input field. When I want to close the window with a "click-out-side", the input field becomes empty again.

See here:
https://codesandbox.io/s/dazzling-water-tds7xh

is this behavior intentional?

the solution to the problem would be to use a v-model instead of a computed value in the datetime renderer.

### Expected behavior

I have a Date field (format=date), then I click on a tag in the overlay and the day appears in the Input field and the overlay closes.

### Steps to reproduce the issue

1. Go to: https://tds7xh.csb.app/
2. Click on clock icon
3. Select a time
4. close the window with a click-out-side 

### In which browser are you experiencing the issue?

Google Chrome 114.0.5735.124 

### Which Version of JSON Forms are you using?

v3.*

### Framework

Vue 3

### RendererSet

Vanilla

### Additional context

I have fixed that behavior with a custom renderer, but it would be good to know if this is intentional or if there is a reason why it was solved this way.

sdirix:
Hi @davewwww, thanks for the report.

Sadly I can't reproduce the behavior, even when using the `codesandbox` you provided.

![VueVanillaDateTime](https://github.com/eclipsesource/jsonforms/assets/8998368/d45de49b-b2be-4dcf-988f-778529b619dd)

There is an issue with the time renderer which stores the time in an incompatible format to JSON Schema. But besides that everything seems working. I clicked once on each overlay and then once outside of the overlay.

Can you still reproduce the issue? If yes, can you also attach a video or GIF?

davewwww:
https://github.com/eclipsesource/jsonforms/assets/921303/d8f0f628-697e-4373-9cd5-3bbb23431a56

davewwww:
if i change my browser language from german to english it works. very strange. 

https://github.com/eclipsesource/jsonforms/assets/921303/161bfa79-6f2b-415d-9c17-d5005ec1695f

sdirix:
I see. We're using the browser native controls, so it seems that they behave differently depending on the locale. That is certainly unintended.

LukasBoll:
For me this problem also occurred with english browser setting.
[...]

Issue #1720, opened 2021-03-15
Table Renderer for Vue Vanilla


sdirix:
Implement a table renderer for Vue Vanilla. Includes implementing a `cell` renderer set.

PR #2279, opened 2024-02-12
Cell renderer set for vue vanilla (work in progress)

davewwww:
regarding: #1720

Hi,

I have tried to implement a proposal for the missing cell feature in vue vanilla.

What do you think, does it work so far?


packages/vue-vanilla/src/cells/TextCell.vue
```diff
@@ -0,0 +1,36 @@
...
+const props = defineProps(rendererProps<ControlElement>());
+
+const input = useVanillaControl(
+  useJsonFormsControl(props),
+  (target) => target.value || undefined
+);
+const { styles, control, appliedOptions, onChange, isFocused } = input;
+
+defineOptions({
+  tester: rankWith(1, isStringControl) as RankedTester,
```
sdirix, line 34: I think we should call this `defaultTester`.

packages/vue-vanilla/src/cells/TextCell.vue
```diff
@@ -0,0 +1,36 @@
...
+} from '@jsonforms/core';
+import { rendererProps, useJsonFormsControl } from '@jsonforms/vue';
+import { useVanillaControl } from '../util';
+
+const props = defineProps(rendererProps<ControlElement>());
+
+const input = useVanillaControl(
+  useJsonFormsControl(props),
+  (target) => target.value || undefined
+);
```
sdirix, line 30: We should implement and use cell specific variants of the `useJsonFormsControl` and maybe `useVanillaControl` bindings, like we do in React.

By reusing the one for the `renderers`, we will calculate a lot of unnecessary props here (e.g. label) which are then never used.

packages/vue-vanilla/src/controls/index.ts
```diff
@@ -1,35 +1,38 @@
...
-  numberControlRendererEntry,
-  integerControlRendererEntry,
+  //numberControlRendererEntry,
+  //integerControlRendererEntry,
   enumControlRendererEntry,
   oneOfEnumControlRendererEntry,
   dateControlRendererEntry,
   dateTimeControlRendererEntry,
   timeControlRendererEntry,
   booleanControlRendererEntry,
```
sdirix, line 36: I think we could get rid of all the renderers in favor of cells, except maybe the boolean one.

sdirix (commented):
Hi @davewwww,

I discussed with the team. We very much like leveraging the cells mechanism and using it similar to the React Vanilla approach. If this PR is properly finished, then we don't even need the `ControlWrapper` customization mechanism anymore. Instead one can simply override the `InputControlRenderer` with the same end result. As this is the cleaner approach, we would prefer that.

Would you be willing to finish this contribution, i.e. removing most of the current control renderers which we have, replacing them with a cell variant?

davewwww:
> Hi @davewwww,
> 
> I discussed with the team. We very much like leveraging the cells mechanism and using it similar to the React Vanilla approach. If this PR is properly finished, then we don't even need the `ControlWrapper` customization mechanism anymore. Instead one can simply override the `InputControlRenderer` with the same end result. As this is the cleaner approach, we would prefer that.
> 
> Would you be willing to finish this contribution, i.e. removing most of the current control renderers which we have, replacing them with a cell variant?

That is of course an interesting thought. So far I use the ControlWrapper in each of my custom renderers.
But perhaps it would still be a good idea to keep the ControlWrapper in place if you don't want to go the cell route but want to stick with your "old" custom renderer. Topic backwards compatibility

And yes, if you give the GO, then I would continue to work on it.

sdirix:
> > Hi @davewwww,
> > I discussed with the team. We very much like leveraging the cells mechanism and using it similar to the React Vanilla approach. If this PR is properly finished, then we don't even need the `ControlWrapper` customization mechanism anymore. Instead one can simply override the `InputControlRenderer` with the same end result. As this is the cleaner approach, we would prefer that.
> > Would you be willing to finish this contribution, i.e. removing most of the current control renderers which we have, replacing them with a cell variant?
> 
> That is of course an interesting thought. So far I use the `ControlWrapper` in each of my custom renderers. But perhaps it would still be a good idea to keep the `ControlWrapper` in place if a customer don't want to go with this cells but want to stick with your "old" custom renderer. Topic backwards compatibility

- Refactoring the existing renderers is backward compatible without any effect for existing users (besides having to hand over the `cells` too).
- We can keep the `ControlWrapper` for the `InputControlRenderer` so that anyone who imported the `ControlWrapper` can still use it
- If users have existing custom renderers, they will continue to work without issues, even if not using the `cell` approach.

I would just avoid introducing the new control-wrapper injection mechanism as a new additional way of customizing as it's no longer needed. If you have existing custom renderers using the control-wrapper and you would like to use your own customized one, then you can simply use your own implementation of control-wrapper for them. Then you register an additional custom `InputControlWrapper` using your customized control-wrapper for all the off-the-shelf use cases you didn't customize before.

> And yes, if you give the GO, then I would continue to work on it.

Please go ahead, we're looking forward to this feature :+1:  Let us know if you need any support or have questions.

packages/vue-vanilla/src/cells/TextCell.vue
sdirix: In the current renderers it's called `tester` as we export the whole registry entry and there it must be named `tester`. However we already had use cases where people reused renderers with slightly different testers and/or ranks for the testers. So calling it the `defaultTester` made sense to me.

However I don't have any hard feelings here, so we could just keep `tester`

packages/vue-vanilla/src/cells/TextCell.vue
davewwww: @sdirix is the implementation in the current commit as you imagine it? I have used the useVanillaCell() and useJsonFormsCell() functions at the cells, whereby the useVanillaCell() is actually only a copy of the useVanillaControl(), so far correct?

packages/vue-vanilla/src/controls/index.ts
davewwww: why not the boolean one? is there any special handling there? looks pretty much the same as the others.

sdirix:
@davewwww Let me know once I shall take a look again :+1:

davewwww:
there is currently a error at the building that i wanted to investigate. but i don't really understand it, maybe you have an idea.
```bash
Error: [@vue/compiler-sfc] No fs option provided to `compileScript` in non-Node environment. File system access is required for resolving imported types.
```

Besides that, all cells are converted and you can have a look at it. especially the date & datetime cells.

davewwww:
@sdirix What do you think about my proposal? I would like to contribute more cells & renderers for vue vanilla .

sdirix:
Hi @davewwww, we had less resources than usual in the last weeks which is why I did not have time yet to look at the PR again. I will do so as soon as possible, probably within a week.

If possible it would be great if you could make sure that the build is green. Then if the code is fine, we can immediately merge.

packages/vue-vanilla/dev/components/App.vue
```diff
@@ -140,7 +146,8 @@ export default defineComponent({
           :schema="example.schema"
           :uischema="example.uischema"
           :renderers="renderers"
-          :config="config"
+          :cells="cells"
+        :config="config"
```
sdirix, line 150: Nitpick: Formatting

packages/vue-vanilla/src/cells/BooleanCell.vue
```diff
@@ -0,0 +1,33 @@
...
+
+const input = useVanillaCell(
+  useJsonFormsCell(props),
+  (target) => target.checked
+);
+const { styles, cell, appliedOptions, onChange, isFocused } = input;
+
+defineOptions({
+  tester: rankWith(1, isBooleanControl) as RankedTester,
+});
```
sdirix, line 32: Minor: Instead of casting we should make sure that the type system checks for us, e.g.
```suggestion
const tester: RankedTester = rankWith(1, isBooleanControl);
defineOptions({
  tester,
});
```

packages/vue-vanilla/src/cells/DateTimeCell.vue
```diff
@@ -0,0 +1,44 @@
...
+const fromISO = (str: string | undefined) => str?.slice(0, 16);
+const toISO = (str: string) => str + ':00.000Z';
+
+const dataTime = ref();
+const setDateTime = (str: string | undefined) => {
+  dataTime.value = fromISO(str);
+};
+
+setDateTime(props.data.value);
+watch(() => props.data, setDateTime);
```
sdirix, line 39: This is basically just a mapper, right? So I don't think we need to go through an additional binding, couldn't we just do the following:
```
":value=fromISO(cell.data)"
```

Also `props.data` does not exist, right?

packages/vue-vanilla/src/controls/InputControlRenderer.vue
```diff
@@ -5,40 +5,35 @@
...
-      :disabled="!control.enabled"
-      :autofocus="appliedOptions.focus"
-      :placeholder="appliedOptions.placeholder"
-      @change="onChange"
-      @focus="isFocused = true"
-      @blur="isFocused = false"
+    <DispatchCell
+      v-bind="control"
+      :id="controlWrapper.id"
+      :handle-change="handleChange"
```
sdirix, line 11: Similar to React Vanilla ([see here](https://github.com/eclipsesource/jsonforms/blob/24f4e5f6280263a4b8e1263aa2a302613b546272/packages/vanilla-renderers/src/controls/InputControl.tsx#L108-L114)) we should only hand over the minimum and not bind all of `control` here. The cell bindings will take care of the rest

sdirix (commented):
I had a look through the sources and commented what stuck out to me. Generally speaking this is good work and I love to see this merged in the near future ;)

sdirix:
> there is currently a error at the building that i wanted to investigate. but i don't really understand it, maybe you have an idea.
> 
> ```shell
> Error: [@vue/compiler-sfc] No fs option provided to `compileScript` in non-Node environment. File system access is required for resolving imported types.
> ```
> 
> Besides that, all cells are converted and you can have a look at it. especially the time & datetime cells.

This error likely comes from the type-based `defineProps` generation you are trying to use. It seems it's not so stable of a feature yet as I found multiple unresolved error reports of people running into problems with that. My suggestion would be to go back to the regular prop definitions, as we use them the other Vue components, like [here](https://github.com/eclipsesource/jsonforms/blob/24f4e5f6280263a4b8e1263aa2a302613b546272/packages/vue/src/jsonFormsCompositions.ts#L59-L92).

packages/vue-vanilla/src/cells/BooleanCell.vue
davewwww: that is not possible with `defineOptions`

> `defineOptions()` in <script setup> cannot reference locally declared variables because it will be hoisted outside of the setup() function. If your component options require initialization in the module scope, use a separate normal <script> to export the options instead.

see https://vuejs.org/api/sfc-script-setup.html#defineoptions

an alternative could be:
```ts
defineOptions(
  ((): { tester: RankedTester } => {
    const tester: RankedTester = rankWith(1, isBooleanControl);
    return { tester };
  })()
);
```
but that looks a bit over engineered

packages/vue-vanilla/src/cells/DateTimeCell.vue
davewwww: yes, `props.data` isnt right, i have changed it to `input.cell.value.data`

but this code is required because of the `watch()` on `cell.data`.
in this cell dont use `":value=fromISO(cell.data)"`, i use `v-model="dataTime"`, that fixes the issue #2156

packages/vue-vanilla/src/cells/BooleanCell.vue
davewwww: another option could be a second script tag without the "setup" key next to the first one:
```ts
<script setup lang="ts">
import { rendererProps, useJsonFormsEnumCell } from '@jsonforms/vue';
import type { ControlElement } from '@jsonforms/core';
import { useVanillaCell } from '../util';

const props = defineProps(rendererProps<ControlElement>());

const input = useVanillaCell(
  useJsonFormsCell(props),
  (target) => target.checked
);
const { styles, cell, appliedOptions, onChange, isFocused } = input;
</script>

<script lang=ts>
import { isBooleanControl, type RankedTester, rankWith } from '@jsonforms/core';
export const tester: RankedTester = rankWith(1, isBooleanControl);
</script>
```

but your es-lint dont like it and move the imports to the first script tag

sdirix:
Update: I will re-review this PR again this week

sdirix:
I like the cells but we need to fix the build errors. I started refactoring the cells but did run out of time. I think we could try not using the `setup` script but declaring them like we do the regular components. Then I would remove the `tester` from each `.vue` file and declare them in a separate file. The combined tester in the `.vue` files lead to all kind of errors in the past anyway with Vite.

Issue #2033, opened 2022-10-09
Angular date picker: selected date is off by one day


TheMicDiet:
### Describe the bug

When selecting a date from the date picker the date form field value is set to one day before the selected date. For example, choosing 9.10.2022 from the picker will result in 8.10.2022 being displayed in the date form field.

### Expected behavior

The correct date value should be set, i.e., the date set and displayed in the date form field should match the selected date from the picker.

### Steps to reproduce the issue

1. Open the date picker on a date field.
2. Click on any day.
3. See the value in the date field is off by one day compared to the day previously clicked on.

### In which browser are you experiencing the issue?

Google Chrome 106.0.5249.103

### Framework

Angular

### RendererSet

Material

### Additional context

This issue occurs both in my own app and on the JSONForms Angular Material Playground (https://jsonforms-angular-seed.netlify.app/).

sdirix:
Possible cause: Maybe some time zone shenanigans are interfering here. For example it could be that we set the time to 00:00 of the selected day, but because of some automatic timezone consideration the day is then calculated to the day before.

TheMicDiet:
I think you’re right. If I change my time zone to UTC, the date is correct.

PR #2281, opened 2024-02-12
Fix Angular Datepicker

LukasBoll:
This pull request addresses the "Angular date picker: selected date is off by one day" #2033 issue.
Starting from [this](https://github.com/eclipsesource/jsonforms/pull/2035) implementation, I added the option to provide date formats and views in the uischema.

### DIFF
--- a/packages/angular-material/tsconfig.json
+++ b/packages/angular-material/tsconfig.json
@@ -5,7 +5,6 @@
     "declarationMap": true,
     "inlineSources": true,
     "sourceMap": true,
-    "target": "es6"
   },
   "exclude": ["node_modules"],
   "files": ["src/index.ts"],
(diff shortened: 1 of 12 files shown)


packages/angular-material/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "declarationMap": true,
     "inlineSources": true,
     "sourceMap": true,
+    "target": "es6"
```
sdirix, line 8: Why explicitly target es6? Does es2021 not work for you?

sdirix (changes requested):
The bug seems to be fixed! However I encountered another issue during testing:

It's not possible for me to enter a "schemaBased" date via the keyboard into the textual input. The date always clears.

That's the first input of the "Dates" example

packages/angular-material/tsconfig.json
LukasBoll: I fixed the issues and could revert this change

sdirix:
@LukasBoll Can you resolve the conflicts?

sdirix (approved):
LGTM!

PR #2287, opened 2024-02-17
feat(core/testers): add hasOption tester

zaosoula:
This PR adds a tester name `hasOption` allowing to check the presence of an option without specifying its value


packages/core/src/testers/testers.ts
```diff
@@ -215,6 +216,24 @@ export const optionIs =
     return !isEmpty(options) && options[optionName] === optionValue;
   };
 
+/**
+ * Checks whether the given UI schema has an option with the given
+ * name. If no options property
+ * is set, returns false.
```
sdirix, line 222:
```suggestion
 * name. If no options property is set, returns false.
```

sdirix (approved):
Looks good! Thanks for the contribution

Issue #2288, opened 2024-02-19
React MUI : Allow the translation of enum value used for elementLabelProp


Maxouwell:
### Describe the bug

At the moment, when an enum field is used in a `elementLabelProp` option, the raw enum value is used, without allowing to translate it via the translate function.
![image](https://github.com/eclipsesource/jsonforms/assets/13748802/2f12cbca-7eed-4aeb-9908-61a66dc91ca6)

### Expected behavior

The value should be sent to the translate function for customizing.

### Steps to reproduce the issue

Create a JsonForms with :

Json Schema
```
{
  type: 'array',
  items: {
    type: 'object',
    properties: {
      message: {
        type: 'string',
        maxLength: 3,
      },
      messageType: {
        type: 'string',
        enum: ['MSG_TYPE_1', 'MSG_TYPE_2'],
      },
    },
  },
}
``` 
UI Schema :
```
{
  type: 'Control',
  scope: '#',
  options: {
    elementLabelProp: 'messageType',
    detail: {
      type: 'HorizontalLayout',
      elements: [
        { type: 'Control', scope: '#/properties/message' },
        { type: 'Control', scope: '#/properties/messageType' },
      ],
    },
  },
}
```
Data :
```
[
  {
    message: 'El Barto was here',
    messageType: 'MSG_TYPE_1',
  },
  {
    message: 'Yolo',
  },
]
```

### Which Version of JSON Forms are you using?

v.3.2.1

### Framework

React

### RendererSet

Material

### Additional context

I think an old bug was repurposed to highlight this behavior but I can't find it anymore.

lucas-koehler:
Hi @Maxouwell ,
thanks for bringing this up :) I agree that this would be a nice extension to the behavior of the ExpandPanelRenderer. Would you like to contribute this?

Maxouwell:
Hi @lucas-koehler !
Has you've seen, I already submited a PR for this :)

As a side question, what would be the best way to check with you if something is a bug or not ?
I came across some odd behavior while doing the PR that I'm not sure if it was intended. Should I open a bug for each one ?
(on the top of my mind, MaterialEnumArrayRenderer that could have a FormLabel, and a description, and that does not support const value (but it means having withJsonFormsMultiEnumProps addind const support, and it impact the core and the other type of renderer, which I'm not able to fix/expand))

lucas-koehler:
Hi @Maxouwell ,
[...]

PR #2289, opened 2024-02-19
Allow the translation of enum value used for elementLabelProp in ExpandPanelRenderer

Maxouwell:
PR for #2288


packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
```diff
@@ -296,12 +299,35 @@ export const withContextToExpandPanelProps = (
...
+      childLabel = get(childData, childLabelProp, '');
+
+      const childSchema = Resolve.schema(
+        schema,
+        `#/properties/${encode(childLabelProp)}`,
+        rootSchema
+      );
+      if (
+        childSchema &&
+        childSchema.type === 'string' &&
```
sdirix, line 318: The check for the type is not needed

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
```diff
@@ -296,12 +299,35 @@ export const withContextToExpandPanelProps = (
...
+    const { childLabelProp, schema, rootSchema, path, index, uischemas } =
+      props;
     const childPath = composePaths(path, `${index}`);
     const childData = Resolve.data(ctx.core.data, childPath);
-    const childLabel = childLabelProp
-      ? get(childData, childLabelProp, '')
-      : get(childData, getFirstPrimitiveProp(schema), '');
+
+    let childLabel;
+    if (childLabelProp) {
```
sdirix, line 308: The label determination should be within a `useMemo`

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
```diff
@@ -296,12 +299,35 @@ export const withContextToExpandPanelProps = (
...
+
+      const childSchema = Resolve.schema(
+        schema,
+        `#/properties/${encode(childLabelProp)}`,
+        rootSchema
+      );
+      if (
+        childSchema &&
+        childSchema.type === 'string' &&
+        (childSchema.enum !== undefined || childSchema.const !== undefined)
```
sdirix, line 319: Currently we only check for the `enum` case but we should also cover the "one of enums" which are pairs of const an optional title properties, similar to the testers of the enum and oneofenum controls.

Then we need to call the `enumToEnumOptionMapper` or the `oneOfToEnumOptionMapper` accordingly.

sdirix (commented):
Looks good already. However we should also support the one of enum case

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
Maxouwell: I don't have case of enumOf on my side, I'll look to find some in the examples, and add it.

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
Maxouwell: Fixed

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
Maxouwell: Fixed

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
Maxouwell: Added the enumOf case.

Note: the default behavior when no translating is setup will change (showing `const` value => showing `title` value instead). I think it's better but your mileage may vary.

sdirix: Its definitely better ;)

Maxouwell:
Resolved the conflicts, rebased and squasshed

sdirix (changes requested):
I had a look, improved the types and aligned the function declarations to arrow style as the rest of the code base.

I also noticed the following behavior.

Assume the following JSON Schema in use:
```js
{
  type: 'object',
  properties: {
    comments: {
      type: 'array',
      title: 'Messages',
      items: {
        type: 'object',
        properties: {
          message1: {
            type: 'string',
          },
          message2: {
            type: 'string',
            enum: ['foo', 'bar'],
          },
        },
      },
    },
  },
}
```

Then the auto generated keys for the enum will be `comments.message2.foo` and `comments.message2.bar`.

However for the summary title in the expand panel, the following keys will be used: `comments.foo` and `comments.bar`. Therefore if a proper translation of the enum is used, it will not automatically show up in the expand panel, instead another key needs to be additionally translated.

I assume that this was not on purpose?

Can you fix the issue? Ideally the test case should also be improved. In the moment there is just a mock translator in there which does not check the handed over keys. I assume that is why the issue was overlooked.

Maxouwell:
You were right, the previous code had problems on translation key, and on elementLabelProp on deeper path.

I fixed it and added some tests for the deep cases.

packages/material-renderers/test/renderers/MaterialArrayLayout.test.tsx
```diff
@@ -738,4 +910,236 @@ describe('Material array layout', () => {
...
+    expect(getChildLabel(wrapper, 0)).toBe(
+      'translator.root.messageType.MSG_TYPE_1'
+    );
+    // TODO Possible impact to existing jsonForm setup, previously if the translate fonction was setup,
+    //  the enumOf title was written as-is and not pushed to the translation function with a key.
+    //  The oneOf title is pushed for translation over the const value in renderer.ts/oneOfToEnumOptionMapper(),
+    //  seems a bug to me
+    expect(getChildLabel(wrapper, 1)).toBe(
+      'translator.root.messageType.Second message type'
+    );
```
Maxouwell, line 1027: I came across this while fixing the code, I think it needs your opinion.

sdirix: The title is preferred over the value as in practice the value is often a non-string like an object while the title will always be a string. If that is not wanted, the user can define an `i18n` key in the schema.

packages/material-renderers/test/renderers/MaterialArrayLayout.test.tsx
```diff
@@ -738,4 +910,236 @@ describe('Material array layout', () => {
...
+
+    expect(wrapper.find(MaterialArrayLayout).length).toBeTruthy();
+
+    expect(getChildLabel(wrapper, 0)).toBe(
+      'translator.root.messageType.MSG_TYPE_1'
+    );
+    // TODO Possible impact to existing jsonForm setup, previously if the translate fonction was setup,
+    //  the enumOf title was written as-is and not pushed to the translation function with a key.
+    //  The oneOf title is pushed for translation over the const value in renderer.ts/oneOfToEnumOptionMapper(),
+    //  seems a bug to me
```
sdirix, line 1024:
```suggestion
```

sdirix (commented):
I found one minor and one major issue with the current state. Besides that it looks good!

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
```diff
@@ -334,6 +379,69 @@ export const withContextToExpandPanelProps = (
...
+  childPath: string,
+  childLabelProp: string,
+  schema: JsonSchema,
+  rootSchema: JsonSchema,
+  translateFct: Translator,
+  uiSchema: UISchemaElement
+) => {
+  const childData = Resolve.data(data, childPath);
+
+  if (!childLabelProp) return get(childData, getFirstPrimitiveProp(schema), '');
```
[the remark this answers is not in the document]
Maxouwell: Fixed by https://github.com/eclipsesource/jsonforms/pull/2289/commits/27ad9db083068fe212d765e1b3a07dca750800f8

packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx
```diff
@@ -334,6 +379,69 @@ export const withContextToExpandPanelProps = (
...
+      .split('.')
+      .map((p) => encode(p))
+      .join('/properties/')}`,
+    rootSchema
+  );
+
+  const fallbackI18nKey =
+    getI18nKeyPrefix(schema, uiSchema, childPath) +
+    '.' +
+    getI18nKeyPrefix(schema, uiSchema, childLabelProp);
```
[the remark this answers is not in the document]
Maxouwell: Fixed by:
* https://github.com/eclipsesource/jsonforms/pull/2289/commits/72229bfaae733bd993810a8f441039fc6b02d103 for i18n on schema
* https://github.com/eclipsesource/jsonforms/pull/2289/commits/3091b2726333eae6b6e93566cf20ae09550c44e5 for i18n on uiSchema for enum (and test refactoring)

I didn't add the i18n on UiSchema for oneOf as I'm not sure it is coherent

Maxouwell:
I just saw that the same behaviour exist in ListWithDetail, and that the solution don't support if `elementLabelProp` point to an enum array (or oneOf array if it's a thing).
I propose we finish this PR and I will open others PRs for these other ones.

sdirix:
Update: I will re-review this PR again this week

sdirix (changes requested):

packages/core/src/util/label.ts
```diff
@@ -81,3 +100,69 @@ const labelDescription = (text: string, show: boolean): LabelDescription => ({
...
+  translateFct: Translator,
+  uiSchema: UISchemaElement
+): any => {
+  const childData = Resolve.data(data, childPath);
+
+  if (!childLabelProp) childLabelProp = getFirstPrimitiveProp(schema);
+
+  const currentValue = get(childData, childLabelProp, '');
+
+  if (!childLabelProp) return currentValue;
```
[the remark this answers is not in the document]
sdirix: actually this was correct and I readded it. I misunderstood the purpose of the line here.

PR #2291, opened 2024-02-19
dev: use the given i18n from the current example in vue vanilla example app

davewwww:
As described here https://github.com/eclipsesource/jsonforms/pull/2270#pullrequestreview-1884750920, the `i18n` key was not used in the example app to display the given translations. 

Which led to the fact that here: https://deploy-preview-2270--jsonforms-examples.netlify.app/vue-vanilla/#categorization, no translation can be seen.


sdirix (approved):
LGTM! Thanks!

Issue #2239, opened 2024-01-03
Support different classes for different controls


vanillajonathan:
### Is your feature request related to a problem? Please describe.

I cannot use the Bootstrap classes with the Vanilla renderer because the Vanilla renderer only lets you set one class for all `input` control types.

In Bootstrap the inputs should have the `form-control` class but for checkboxes it should be `form-check-input` instead.

### Describe the solution you'd like

A way to configure different CSS classes for different input types.
https://github.com/eclipsesource/jsonforms/blob/master/packages/vue-vanilla/src/styles/defaultStyles.ts

Perhaps:
```javascript
  control: {
    root: 'control',
    wrapper: 'wrapper',
    label: 'label',
    description: 'description',
    input: 'input',
    error: 'error',
    textarea: 'text-area',
    select: 'select',
    option: 'option',
  },
```

could be:
```javascript
  control: {
    root: 'control',
    wrapper: 'wrapper',
    label: 'label',
    description: 'description',
    input: {
      color: 'form-control form-control-color',
      checkbox: 'form-input-check',
      radio: 'form-input-check',
      default: 'form-control'
    },
    error: 'error',
    textarea: 'text-area',
    select: 'select',
    option: 'option',
  },
```

### Describe alternatives you've considered

Fork the Vanilla renderer.

### Framework

Vue 3

### RendererSet

Vanilla

sdirix:
Hi @vanillajonathan, Thanks for the suggestion. This definitely makes sense to me.

I'm wondering about the approach, whether we should split per property as suggested, i.e.

```
  control: {
    root: 'control',
    wrapper: 'wrapper',
    label: 'label',
    description: 'description',
    input: {
      color: 'form-control form-control-color',
      checkbox: 'form-input-check',
      radio: 'form-input-check',
      default: 'form-control'
    },
    error: 'error',
    textarea: 'text-area',
    select: 'select',
    option: 'option',
  },
```

or split at the root level, i.e.

```
  control: {
    checkbox: {
      root: 'control',
      wrapper: 'wrapper',
      label: 'label',
      description: 'description',
      error: 'error',
      input: 'input'
    },
    text: {
      root: 'control',
      wrapper: 'wrapper',
      label: 'label',
      description: 'description',
      error: 'error',
      input: 'input'
    },
  },
```

davewwww:
Hi,

is there anything i can to do here? i can't see a problem here?

[...]

PR #2292, opened 2024-02-22
Radio Group Renderer for Vue Vanilla (WIP)

davewwww:
Hi,

this is a proposal for the radio group renderer, both enum and oneOf, when format=radio.

As I am also working on the [Cell Renderer Set](https://github.com/eclipsesource/jsonforms/pull/2279) I wondered whether this renderer is actually more of a cell instead of renderer

Furthermore, the question of `Styles` naming, as is currently being discussed here: [Support different classes for different controls](https://github.com/eclipsesource/jsonforms/issues/2239#top)
Is `radioGroup` a child of control or a new type at root level?

What do you think?


lucas-koehler (changes requested):
Hi @davewwww , thanks for the contribution :heart: 
Sorry for the late review, we must have overlooked this PR :sweat_smile: 

I think the `RadioGroupControl.vue` could definitely be a cell. However, I would not do this in this PR but rather contribute this PR first and then potentially refactor it in the cell PR.

Could you add some unit tests for the renderers, please? We aim to have at least basic coverage for all renderers.

Regarding the css classes. Congruent with https://github.com/eclipsesource/jsonforms/issues/2239#issuecomment-1938910293, I think the new radio group properties should be added in a new child object of `control`. E.g.:

```ts
export const defaultStyles: Styles = {
  control: {
    root: 'control',
    wrapper: 'wrapper',
    label: 'label',
    description: 'description',
    input: 'input',
    radioGroup: {
       group: 'radio-group'
       item: 'radio-group-item',
       itemLabel: 'radio-group-item-label',
    }
   // [...]
  }
}

```

lucas-koehler:
Thanks again for the contribution :heart: I close this now as it did not have any activity in over a year. Feel free to open another PR if this is still relevant.
Contributions are always highly appreciated :rocket:

PR #2293, opened 2024-02-22
Upgrade ESLint to `^8.56.0`, regenerate pnpm-lock, consolidate vs code settings

lucas-koehler:
- Upgrade ESLint for all packages to ^8.56.0. This is now possible because we no longer support Angular 12.
- Regenerate pnpm-lock to upgrade all dependencies within their version ranges
- Remove obsolete separate configurations for different JS/TS languages.


sdirix (approved):
LGTM

Issue #2296, opened 2024-02-27
React MUI : MaterialEnumArrayRenderer - Add Form Label and description 


Maxouwell:
### Is your feature request related to a problem? Please describe.

Currently, MaterialEnumArrayRenderer (MultiChoice enum) are not on par with the radio group component.

#### On the label
Multi choice 
![image](https://github.com/eclipsesource/jsonforms/assets/13748802/dd0303ea-3eb9-4ec6-8de6-08b8a0d63456)

Radio Group : 
![image](https://github.com/eclipsesource/jsonforms/assets/13748802/581e7d7a-02a0-4b9c-8860-b53e056cef62)

#### On the description 
Only the error is shown on multi choice enum, while the description is shown on the radio group
![image](https://github.com/eclipsesource/jsonforms/assets/13748802/4f779183-4480-4ea6-900f-fae05dfa7e51)

### Describe the solution you'd like

The component may have a label, as available via MUI
![image](https://github.com/eclipsesource/jsonforms/assets/13748802/ef2ad318-b29b-405b-8b66-42371be7d5f1)

I'm open to do a PR, but should the label be added as is or behind an option to not impact existing forms ?

### Describe alternatives you've considered

N/A

### Framework

React

### RendererSet

Material

PR #2297, opened 2024-02-27
React MUI : MaterialEnumArrayRenderer - Add Form Label and description

Maxouwell:
PR for #2296. with always on label (not conditionnal to an option)

### DIFF
--- a/packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
+++ b/packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
@@ -3,6 +3,7 @@ import {
   ControlProps,
   DispatchPropsOfMultiEnumControl,
   hasType,
+  isDescriptionHidden,
   JsonSchema,
   OwnPropsOfEnum,
   Paths,
@@ -11,6 +12,7 @@ import {
   resolveSchema,
   schemaMatches,
   schemaSubPathMatches,
+  showAsRequired,
   uiTypeIs,
 } from '@jsonforms/core';
 
@@ -21,15 +23,23 @@ import {
   FormControlLabel,
   FormGroup,
   FormHelperText,
+  FormLabel,
   Hidden,
 } from '@mui/material';
 import isEmpty from 'lodash/isEmpty';
 import React from 'react';
+import merge from 'lodash/merge';
+import { useFocus } from '../util';
 
 export const MaterialEnumArrayRenderer = ({
+  config,
+  id,
   schema,
   visible,
   errors,
+  description,
+  label,
+  required,
   path,
   options,
   data,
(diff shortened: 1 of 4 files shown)


packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
```diff
@@ -38,9 +48,35 @@ export const MaterialEnumArrayRenderer = ({
...
+    focused,
+    appliedUiSchemaOptions.showUnfocusedDescription
+  );
+
   return (
     <Hidden xsUp={!visible}>
-      <FormControl component='fieldset'>
+      <FormControl
+        component='fieldset'
+        fullWidth={!appliedUiSchemaOptions.trim}
```
Maxouwell, line 65: Note : I also included the trim option from the RadioGroup component

packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
```diff
@@ -38,9 +48,35 @@ export const MaterialEnumArrayRenderer = ({
...
+        onFocus={onFocus}
+        onBlur={onBlur}
+      >
+        <FormLabel
+          htmlFor={id}
+          error={!isValid}
+          component={'legend' as 'label'}
```
lucas-koehler, line 72: I don't think you need to `as 'label'`. I guess this might have only been necessary in an older mui version with stricter types.
```suggestion
          component={'legend'}
```

packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
```diff
@@ -53,6 +89,7 @@ export const MaterialEnumArrayRenderer = ({
                 key={option.value}
                 control={
                   <MuiCheckbox
+                    id={id}
```
lucas-koehler, line 92: I don't think we should add the id here. With this addition, all checkboxes of this multi enum would have the same id. I'd just remove this again. What do you think?

lucas-koehler (changes requested):
Hi @Maxouwell , thanks a lot for another contribution :heart: 
The code looks already pretty good to me and showing the labels and description works well in the example :)
I just have two small remarks inline. Please have a look.

packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
Maxouwell: It's taken as-is from MaterialRadioGroup. Should I remove it also there ?

lucas-koehler: Sure that would be great. Thanks!

packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
Maxouwell: This one is trickier.

I didn't add the id field, it was already there buried in otherProps.
I surfaced it only because I needed it for the htmlFor field of the FormLabel.

Also, it is required by MuiCheckBox:
image

What I propose is to concat the value as done for the key field:
```
  id={id+ '-checkbox-' + option.value}
  key={'checkbox-' + option.value}
```

But once you do that,
* The `htmlFor={id}` on `FormLabel` has no sense and should be removed
  * _And then_, it make no sense on MaterialRadioGroup where I got it to start with
* The field `id={option.value}` on `FormControlLabel` is also wrong (same id if the option is in another chekbox), and may be `id={id+ '-' + option.value}` or `id={id+ '-checkbox-label-' + option.value}`

Let me know what you think @lucas-koehler !

packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
Maxouwell: Also removed from MaterialSliderControl

Maxouwell: And from this line from MaterialRadioGroup `component={'fieldset' as 'div'}`

Maxouwell: Fixed in commit [20dceb0](https://github.com/eclipsesource/jsonforms/pull/2297/commits/20dceb010e9a1b72be1b570bc97949867836b4e2)
(Unresolved to be sure it's ok for you)

packages/material-renderers/src/complex/MaterialEnumArrayRenderer.tsx
lucas-koehler: > I didn't add the id field, it was already there buried in otherProps.

Ah I missed that, sorry.

I think your suggestion sounds good, to summarize:
- Remove `htmlFor={id}` from `FormLabel`
- Adapt ids of `FormControlLabel` from `id={option.value}` to `id={id+ '-label-' + option.value}` (I think just `label` instead of `checkbox-label` in combination with the value is enough)
- Adapt id on the checkbox to `id={id+ '-' + option.value}`

Maxouwell: Done !
[e32cbae](https://github.com/eclipsesource/jsonforms/pull/2297/commits/e32cbae1d5d389cc6c9249f044108e88a9049afa)

lucas-koehler (approved):
LGTM now, thanks :sparkles:

PR #2298, opened 2024-03-06
Set `type` of `<button>`s in vanilla array renderer

joefreeman:
This updates the `<button>`s in the vanilla array renderer to set the `type` to `"button"`. This allows a `<JsonForms>` to be contained within a `<form>` without these buttons triggering a submit of the form.

I believe this is fixing the same issue as #1731, but for the vanilla renderers.


sdirix (approved):
Thanks!

PR #2303, opened 2024-03-14
Disable vanilla array buttons when readonly

joefreeman:
This updates the vanilla array controls (`TableArrayControl` and `ArrayControlRenderer`) to disable the buttons (for adding/removing/moving items) when the control is set to readonly (otherwise the data can be updated while readonly).

To (manually) test this, I've updated the examples to support setting the `readonly` property of the `<JsonForms>`, and added toggle actions to a couple of the array examples. (Not sure if there's a better way to do this?)

### DIFF
--- a/packages/examples/src/example.ts
+++ b/packages/examples/src/example.ts
@@ -41,6 +41,7 @@ export interface ExampleDescription {
   config?: any;
   actions?: { label: string; apply: (props: StateProps) => any }[];
   i18n?: JsonFormsI18nState;
+  readonly?: boolean;
 }
 
 export interface StateProps {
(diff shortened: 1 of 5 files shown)


sdirix (commented):
Thanks for the contribution :heart:! Works for me!

Adapting the examples is great and allows to easily test the new functionality manually, thanks :+1:!

Please also add a unit test so that the new behavior is safe from regressions.

lucas-koehler:
@sdirix @joefreeman I added tests that check that buttons are rendered with expected enabled and disabled state for enabled/disabled controls.

sdirix (approved):

PR #2307, opened 2024-03-27
fix: update ID on schema change

### LINKED ISSUE #2213: Control.id doesn't get updated properly, if multiple forms are involved
### Describe the bug

I am using Vue2 with the with my own selection of renderers. I have v-cards with an integrated JSONForms object looking like this:
image
The jsonforms object gets updated based on the currently selected tab.

Now the problem: 
Lets imagine there is a tab with 2 checkboxes in the form and another tab with 5 checkboxes. The position of the first two checkboxes inside each form are the same, because there is no other element in the schema. If I switch now between the two tabs (forms), then only the values for the checkboxes where there is a change in the position gets updated, even if all other values for the control get updated correctly.

Example:
When accessing the first tab, the controls show the correct values:
Screenshot from 2023-11-14 22-33-48

But after switching the tab and the other form is shown, the control updates the path and the other values correctly, but doesn't update the id:
Screenshot from 2023-11-14 18-00-06

After reloading the page, the DOM of course shows the correct id, but it should already show the correct value with the form change.

### Expected behavior

After switching from one form to the other, the control.id value updates correctly.

1. Create a json-form component with changable content
2. Create schemas and uischemas with two similar controls where just the prop id is different
3. After creating the properties, look at them in the browser and switch from one content to the other
4. Check if the DOM shows the correct value for the control.id

### Screenshots

_No response_

Mozilla Firefox 119.0.1

3.1.0

### Framework

Vue 2

### RendererSet

### Additional context

Own renderer set but nothing changed in the control.id

LukasBoll:
This commit will update the IDs of all controls if the schema prop handed over to JSON Forms changes.

closes #2213

### DIFF
--- a/packages/react/src/JsonForms.tsx
+++ b/packages/react/src/JsonForms.tsx
@@ -77,6 +77,17 @@ export class JsonFormsDispatchRenderer extends React.Component<
     }
   }
 
+  componentDidUpdate(prevProps: JsonFormsProps) {
+    if (prevProps.schema !== this.props.schema) {
+      removeId(this.state.id);
+      this.setState({
+        id: isControl(this.props.uischema)
+          ? createId(this.props.uischema.scope)
+          : undefined,
+      });
+    }
+  }
+
   render() {
     const {
       schema,
--- a/packages/angular/src/library/abstract-control.ts
+++ b/packages/angular/src/library/abstract-control.ts
@@ -29,6 +29,7 @@ import {
   JsonFormsState,
   JsonSchema,
   OwnPropsOfControl,
+  removeId,
   StatePropsOfControl,
 } from '@jsonforms/core';
 import { Component, Input, OnDestroy, OnInit } from '@angular/core';
@@ -148,6 +149,7 @@ export abstract class JsonFormsAbstractControl<
     if (this.subscription) {
       this.subscription.unsubscribe();
     }
+    removeId(this.id);
   }
 
   isEnabled(): boolean {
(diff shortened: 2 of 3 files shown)


packages/react/src/JsonForms.tsx
```diff
@@ -77,6 +77,17 @@ export class JsonFormsDispatchRenderer extends React.Component<
     }
   }
 
+  componentDidUpdate(prevProps: JsonFormsProps) {
+    if (prevProps.schema != this.props.schema) {
```
lucas-koehler, line 81:
```suggestion
    if (prevProps.schema !== this.props.schema) {
```
Always make sure to use strict inequality (and equality) checks to avoid unexpected bugs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality

LukasBoll: Thank you, I must have missed that here!

LukasBoll:
Hi @lucas-koehler ,
I noticed a bug, where the application crashes when switching between some of the examples if you remove the key attribute in the Vue-example app.
I wanted to confirm, that these errors are not related to this commit before this PR is merged.
Since the same issue occurs on the master branch, I assume the issue is unrelated and will remove the draft status.

lucas-koehler (approved):
@LukasBoll Alright makes sense. I also tested with the vue example app and switching between the person and scope examples successfully updated the id even when removing the key in the example app. Without this fix, this did not work.

PR #2317, opened 2024-04-09
Add context for array operation

eneufeld:
Currently it is very hard to figure out in the middleware what array operation happend to lead to the current array state. There are cases were this information is valuable. To support this all update actions can now have a 'context'. Here we added a specific 'UpdateArrayContext'.

### DIFF
--- a/packages/core/src/actions/actions.ts
+++ b/packages/core/src/actions/actions.ts
@@ -56,6 +56,57 @@ export const UPDATE_I18N = 'jsonforms/UPDATE_I18N' as const;
 export const ADD_DEFAULT_DATA = 'jsonforms/ADD_DEFAULT_DATA' as const;
 export const REMOVE_DEFAULT_DATA = 'jsonforms/REMOVE_DEFAULT_DATA' as const;
 
+export type UpdateArrayContext =
+  | { type: 'ADD'; values: any[] }
+  | { type: 'REMOVE'; indices: number[] }
+  | { type: 'MOVE'; moves: { from: number; to: number }[] };
+
+export const isUpdateArrayContext = (
+  context: object
+): context is UpdateArrayContext => {
+  if (!('type' in context)) {
+    return false;
+  }
+  if (typeof context.type !== 'string') {
+    return false;
+  }
+  switch (context.type) {
+    case 'ADD': {
+      return (
+        'values' in context &&
+        Array.isArray(context.values) &&
+        context.values.length > 0
+      );
+    }
+    case 'REMOVE': {
+      return (
+        'indices' in context &&
+        Array.isArray(context.indices) &&
+        context.indices.length > 0 &&
+        context.indices.every((i) => typeof i === 'number')
+      );
+    }
+    case 'MOVE': {
+      return (
+        'moves' in context &&
+        Array.isArray(context.moves) &&
+        context.moves.length > 0 &&
+        context.moves.every(
+          (m) =>
+            typeof m === 'object' &&
+            m !== null &&
+            'from' in m &&
+            'to' in m &&
+            typeof m.from === 'number' &&
+            typeof m.to === 'number'
+        )
+      );
+    }
+    default:
+      return false;
+  }
+};
+
 export type CoreActions =
   | InitAction
   | UpdateCoreAction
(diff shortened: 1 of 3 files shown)


packages/core/src/actions/actions.ts
```diff
@@ -56,6 +56,66 @@ export const UPDATE_I18N = 'jsonforms/UPDATE_I18N' as const;
 export const ADD_DEFAULT_DATA = 'jsonforms/ADD_DEFAULT_DATA' as const;
 export const REMOVE_DEFAULT_DATA = 'jsonforms/REMOVE_DEFAULT_DATA' as const;
 
+export type UpdateArrayContext =
+  | { type: 'ADD'; values: any[] }
```
sdirix, line 60: Instead of only specifying the values, can we also specify the array index at which they will be added? We already have consumers who customized their array add operations to for example add them at the beginning of the array. That way we could cover them too.

The remove and move information is sufficient :+1:

packages/core/src/actions/actions.ts
```diff
@@ -56,6 +56,66 @@ export const UPDATE_I18N = 'jsonforms/UPDATE_I18N' as const;
...
+  | { type: 'REMOVE'; indeces: number[] }
```
sdirix, line 61: Should be `indexes` or `indices`

packages/core/src/actions/actions.ts
```diff
@@ -56,6 +56,66 @@ export const UPDATE_I18N = 'jsonforms/UPDATE_I18N' as const;
...
+    case 'REMOVE': {
+      if (
+        'indeces' in context &&
+        Array.isArray(context.indeces) &&
+        context.indeces.length > 0 &&
+        context.indeces.every((i) => typeof i === 'number')
+      ) {
+        return true;
+      }
+      return false;
```
sdirix, line 94: We could just return the if-condition instead of returning true and false explicitly ;)

sdirix (changes requested):
Looks good and thanks for adding the tests. Please have a look at my comments.

packages/core/src/actions/actions.ts
eneufeld: I don't think we can, as we don't have that information in the dispatch method. See: https://github.com/eclipsesource/jsonforms/pull/2317/files#diff-a1bd17d367230eda9ec6010e4de56595495fd67e51d4faa5650ec2d800b9e702R836

Am I missing something?

sdirix: Ah yes, you are right, I missed that!

packages/core/src/actions/actions.ts
```diff
@@ -56,6 +56,60 @@ export const UPDATE_I18N = 'jsonforms/UPDATE_I18N' as const;
...
+  }
+  switch (context.type) {
+    case 'ADD': {
+      if (
+        'values' in context &&
+        Array.isArray(context.values) &&
+        context.values.length > 0
+      ) {
+        const firstValueType = typeof context.values[0];
+        return context.values.every((v) => typeof v === firstValueType);
```
sdirix, line 81: I think this check is not necessary. JSON Schema supports different types in arrays.

packages/core/src/actions/actions.ts
```diff
@@ -56,6 +56,60 @@ export const UPDATE_I18N = 'jsonforms/UPDATE_I18N' as const;
...
+    }
+    case 'MOVE': {
+      return (
+        'moves' in context &&
+        Array.isArray(context.moves) &&
+        context.moves.length > 0 &&
+        context.moves.every(
+          (m) =>
+            typeof m === 'object' &&
+            'from' in m &&
```
sdirix, line 101: This will crash in case `m` is `null`. We should add a `!!m` check too.

packages/core/src/actions/actions.ts
eneufeld: but we don't support this in jsonforms, so all newly added values must be of the same type
I'm fine removing this check

packages/core/src/actions/actions.ts
eneufeld: I will add a test case and adapt the code if necessary

packages/core/src/actions/actions.ts
sdirix: We don't support it in JSON Forms in our current renderers but there is no restriction for that in JSON Schema and custom/future renderers might support that.

sdirix (approved):
Works for me!

PR #2318, opened 2024-04-09
deps: Use latest vue version 3.4.21 and recreate pnpm lock

### LINKED ISSUE #2234: Update vue to 3.4+

Updating the pnpm lock file to use vue 3.4.3 causes vue-vanilla's tests to fail with error:

```
Compiling...

  [=========================] 98% (after emitting)

 ERROR  Failed to compile with 1 error

This dependency was not found:

* node:stream in /home/user/jsonforms/node_modules/.pnpm/@vue+server-renderer@3.4.3_vue@3.4.3/node_modules/@vue/server-renderer/dist/server-renderer.cjs.js

To install it, you can run: npm install --save node:stream
  [=========================] 100% (completed)

 WEBPACK  Failed to compile with 1 error(s)

Error in /home/user/jsonforms/node_modules/.pnpm/@vue+server-renderer@3.4.3_vue@3.4.3/node_modules/@vue/server-renderer/dist/server-renderer.cjs.js

  Module not found: 'node:stream' in '/home/user/jsonforms/node_modules/.pnpm/@vue+server-renderer@3.4.3_vue@3.4.3/node_modules/@vue/server-renderer/dist'

 ERROR  mochapack exited with code 1.
 ELIFECYCLE  Test failed. See above for more details.
```

Using Node 18 instead of 16 did not solve the problem.

Upgrade vue peer dependency to 3.4.3+ and fix the vue-vanilla package tests and/or other dependencies to still work.

Do not update vue. This is not a viable long-term solution

### Framework

Vue 3

### RendererSet

Vue Vanilla

### Additional context

_No response_

lucas-koehler:
Fixes #2234 

- Remove pnpm overrides for vue version to use latest version 3.4.21. **Do not touch the peer dependency because no new vue code was introduced => it remains at `^3.2.26`**
- Update vue cli dependencies to latest version ^5.0.8
- Updater @vue.compiler-sfc and @vue/test-utils dependencies to latest versions
- Fix vue vanilla test utils to attach mounted components to document. This is (now) necessary so that click events on checkbox inputs correctly trigger a change event.
- Regenerate pnpm-lock file. This updates all dependencies to the latest version permitted by their version ranges.


packages/angular/package.json
```diff
@@ -55,13 +55,12 @@
...
       "test/**/*"
     ],
-    "compileEnhancements": false,
     "extensions": [
       "ts"
     ],
     "require": [
       "./test-config/ts-node.config.js",
-      "source-map-support/register"
+      "@cspotcode/source-map-support/register.js"
```
sdirix, line 63: Why do we switch to this fork? Is this a mandatory change?

packages/vue-vanilla/package.json
```diff
@@ -90,7 +89,7 @@
     "tslib": "^2.5.0",
     "typedoc": "~0.25.3",
     "typescript": "~4.9.5",
-    "vue": "^3.2.26",
+    "vue": "^3.4.21",
```
sdirix, line 92: As we don't adjust the `peerDependencies`, did you test whether the built package is still compatible with a consumer using `vue` `3.2.26`?

sdirix (commented):
Looks good. Added two questions

packages/vue-vanilla/package.json
lucas-koehler: Good point. I now tested it with the vue seed by packaging the locally built jsonforms packages and installing them there. With this, it worked with vue 3.2.26 :)

packages/angular/package.json
lucas-koehler: Good point. Initially something didn't work in relation to source-map-support and I thought because it is outdated. Turned out, only the `.js` ending was missing. I now reverted back to the original `original-source-map` and only set its dep to the latest version (only a patch version upgrade).

lucas-koehler: I also checked that reverting to original-source-map does not increase the vulnerability count again

sdirix (approved):

Issue #2327, opened 2024-04-29
The delete button should be displayed only when an item is hovered and hidden


stanisev:
### Describe the bug

The delete button should be displayed only when an item is hovered and hidden (via the hide class) by default which was removed In one of the previous MRs

### Expected behavior

We should add back removed `hint` css class

### Steps to reproduce the issue

Use master details renderer. All elements of the master details will have the delete button (delete button should be displayed only on hover)

### Which Version of JSON Forms are you using?

v3.2.1

### Framework

Angular

### RendererSet

Other (please specify in the Additional context field)

JBBianchi:
Little typo ~`hint`~ _`hide`_ 

Discussed in https://github.com/eclipsesource/jsonforms/pull/2229#discussion_r1583038946

PR #2330, opened 2024-05-03
[FIX-#2327] Revert deletion of hide css class in process of angular upgrade

stanisev:
Reverting back .`hide `css class, since delete button should be displayed only when an item is hovered and hidden and not all the time.


lucas-koehler (approved):
LGTM, thanks for the fix :heart:

PR #2341, opened 2024-05-30
initial contribution for vue-vuetify renderer set

kchobantonov:
@sdirix  please review the initial contribution that have the changes that fixes the vue3 integration. Few things to note:

1. The original example data in the vuetify render set was not migrated since we do have an example set with the exceptions for few examples ported - additional properties, login form and etc. The only issue here is that the original examples were having the vuetify UI options that were not included or ported.
2. few minor fixed in the core (issues with undefined in the core)
3. upgrade of `ajv-i18n` since the Vite do not like the require when expecting to use import


kchobantonov:
@sdirix  can you review and check why we have out of memory failure on the build machine ? Perhaps we need a build machine with more memory ?

kchobantonov:
@sdirix  the build is fixed - I had to change 
```
    "build-only": "vite build",
```
to 
```
    "build-only": "NODE_OPTIONS=\"--max-old-space-size=4096\" vite build",
```

in the vuetiy package - if you want you can remove that from there and place it in the ci.yaml the same way how we set that for test to be set on build as well but that is up to you.

sdirix:
Thanks! I plan to take a look at this PR next week.

I was wondering whether we should preserve the history of the other repository, i.e. instead of copying all files over to here, we can add the vue3-branch into this repository as an orphan-branch and then do a merge commit with the content of the first commit here. This would keep the history.

kchobantonov:
I can review and apply those changes - keep in mind that this PR will have almost the same thing as the PR for fixing the vue3 without the recent changes (few weeks ago) that were merged into that repo, and also this PR different from the PR against the older vue 3 vuetify 3 repository since I had to integrate that with the jsonforms main repo and most of the changes are in the example so both example apps are different between this PR and the other one. If you want to keep the history please create the proper PR with the history from the other repo and the either I or you can override that on top and have the history there.

kchobantonov:
@sdirix  the requested PRs are applied to this PR

sdirix:
@kchobantonov Thank you very much!

packages/core/src/util/schema.ts
```diff
@@ -27,7 +27,7 @@ import find from 'lodash/find';
 import { JsonSchema } from '../models';
 
 export const getFirstPrimitiveProp = (schema: any) => {
-  if (schema.properties) {
+  if (schema && schema.properties) {
```
manuelmeister, line 30:
```suggestion
  if (schema?.properties) {
```

Couldn't this be done like this?

sdirix:
I'm looking at the PR. I plan to contribute the migration and finish the review next week.

activeliang:
I'm very eager to use this library, but the date picker issue hasn't been resolved yet. I see that this PR allows the use of the date picker component from Vuetify 3. I hope the code can be merged soon.

sdirix (commented):
Thank you for the contribution :heart:

PR #2349, opened 2024-06-27
feat: support @mui/x-date-pickers v7

### LINKED ISSUE #2328: Support @mui/x-date-pickers v7

I'm using the newest version of MUI X, and I'm getting the warning below:

 WARN  Issues with peer dependencies found
libs/core-react
└─┬ @jsonforms/material-renderers 3.2.1
  └── ✕ unmet peer @mui/x-date-pickers@^6.0.0: found 7.3.1

Support for MUI X v7

At the moment I am ignoring the warning, and I haven't experienced any issues yet.

### Framework

React

### RendererSet

Material

### Additional context

_No response_

seanparmelee:
Closes #2328 

This PR updates the `peerDependencies` of `@jsonforms/material-renderers` to allow for either v6 or v7 of `@mui/x-date-pickers`. I went through the [v6 to v7 migration guide](https://mui.com/x/migration/migration-pickers-v6/) and didn't see any breaking changes that require a code change in `@jsonforms/material-renderers`. I also tried running the [pickers codemod](https://mui.com/x/migration/migration-pickers-v6/#run-codemods) to see if it changes anything, but it didn't. I also checked the material-renderers Dates examples and confirmed they still work.

There is [mention of updating `@mui/material`](https://mui.com/x/migration/migration-pickers-v6/#update-mui-material-package), but the examples seem to work without it. I was trying to better understand why the need for the update (https://github.com/mui/mui-x/pull/12516) and it seemed like it was for a very specific use case (https://github.com/mui/mui-x/issues/12248) that I don't think requires a corresponding peer dependency bump in `@jsonforms/material-renderers`, but happy to do so if you prefer.

In other words, think the material-renderers (as it's currently implemented) can support both `@mui/x-date-pickers` 6 and 7, but this would also mean that any future changes made in the control components would need to be tested against both versions, at least until the next major release of jsonforms where v6 support could be dropped.


packages/material-renderers/package.json
```diff
@@ -100,7 +100,7 @@
     "@jsonforms/react": "workspace:*",
     "@mui/icons-material": "^5.11.16",
     "@mui/material": "~5.13.0",
-    "@mui/x-date-pickers": "^6.5.0",
+    "@mui/x-date-pickers": "^7.7.1",
```
sdirix, line 103: Do we need to adapt the `devDependency` to `x-date-pickers` or would it also work if we keep building against the old one?

sdirix (commented):
Hi @seanparmelee,

Thank you very much for the PR and the testing :heart: 

I just have a minor question

packages/material-renderers/package.json
seanparmelee: Not needed; I can revert this back so it continues to use the old one.

sdirix (approved):

Issue #2366, opened 2024-08-13
isRequired encoded scope paths with slashes not properly matched


adamskeeled:
### Describe the bug

Consider the following schema:
```
{
  "type": "object",
  "unevaluatedProperties": false,
  "properties": {
      "/work/email": {
          "title": "Email with slash",
          "description": "Email address With Slash",
          "type": "string",
          "format": "email"
      }
  },
  "required": [
      "/work/email"
  ],
}
```

Since the path(`/work/email`) contains a forward slash, the final encoded scope looks a bit different (forward slashes get converted to `~1`). When trying to check if the field is required, the most direct path is to use the `required` prop inside the `control` from `mapStateToControlProps`. However, since the path doesn't get decoded, it doesn't match with what is inside the global `required`.

### Expected behavior

The scope path should be decoded and then matched against the `required` field

### Steps to reproduce the issue

1. Use the above schema
2. Use any renderer which matches the above sub-schema and is using the `control.required` to show an asterisk for example
3. Notice that the asterisk won't show up

### Which Version of JSON Forms are you using?

v3.3.0

### Framework

Core

### Additional context

I believe I know where to implement a fix and can even create a PR to help do this.
In the `isRequired` function in the `src/util/renderers.ts` file, the `decode` needs to get imported and then used on the `lastSegment` before `isRequired` returns

adamskeeled:
@sdirix I'm trying to push a PR but it says in the terminal that I don't have permission. Am I missing something?

sdirix:
Thanks for the report! I'll keep the issue open for now until the fix is in upstream JSON Forms ;)

PR #2367, opened 2024-08-14
fix: matching required paths with slashes

adamskeeled:
In JSON Pointers two characters need to be escaped: '~' is converted to '~0' and '/' is converted to '~1'.
When JSON Paths contain one of these two characters, when trying to check whether those path are required, the `isRequired` function is used. That function compares the scoped path (which in this case is the encoded path with the converted symbols) with the closest `required` array (which contains the original unencoded path with either `~` or `/`).

Before checking the `required` array, we need to decode the scoped path so the comparison works properly.

Fixes https://github.com/eclipsesource/jsonforms/issues/2366

### DIFF
--- a/packages/core/src/mappers/renderer.ts
+++ b/packages/core/src/mappers/renderer.ts
@@ -73,7 +73,7 @@ import {
   isVisible,
   Resolve,
   resolveSchema,
-  decode
+  decode,
 } from '../util';
 import {
   Translator,
--- a/packages/core/src/mappers/renderer.ts
+++ b/packages/core/src/mappers/renderer.ts
@@ -73,6 +73,7 @@ import {
   isVisible,
   Resolve,
   resolveSchema,
+  decode
 } from '../util';
 import {
   Translator,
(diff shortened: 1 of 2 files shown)


packages/core/src/mappers/renderer.ts
```diff
@@ -114,7 +115,7 @@ const isRequired = (
   rootSchema: JsonSchema
 ): boolean => {
   const pathSegments = schemaPath.split('/');
-  const lastSegment = pathSegments[pathSegments.length - 1];
+  let lastSegment = pathSegments[pathSegments.length - 1];
```
sdirix, line 118: Let's just decode in place
```suggestion
  const lastSegment = decode(pathSegments[pathSegments.length - 1]);
```

packages/core/src/mappers/renderer.ts
```diff
@@ -127,6 +128,9 @@ const isRequired = (
     rootSchema
   );
 
+  // decode JSON Pointer escape sequences
+  lastSegment = decode(lastSegment);
+
```
sdirix, line 133: then we can remove this again
```suggestion
```

sdirix:
Thanks for the contribution and the included test case :heart: That's what we like to see ;)

packages/core/src/mappers/renderer.ts
adamskeeled: Done 👌

packages/core/src/mappers/renderer.ts
adamskeeled: Done 👌

sdirix (approved):
Thanks for the contribution!

PR #2370, opened 2024-08-27
chore: remove 'matchMedia' mock

sdirix:
The 'matchMedia' mock was previously introduced to workaround the 'Hidden' component of Material UI not rendering in case media queries are not available in the environment.

With JSON Forms 3.3.0 we removed all usages of the 'Hidden' component so the 'matchMedia' mock is also no longer required.


sdirix:
I rebased on `master`, no other changes

PR #2377, opened 2024-09-15
fix(angular): unsubscribe from jsonForm state subscription

### LINKED ISSUE #2354: abstract-control (angular) state subscription is not cleared
### Describe the bug

https://github.com/eclipsesource/jsonforms/blob/master/packages/angular/src/library/abstract-control.ts
ngOnDestroy does unsubscribe, but this.jsonFormsService.$state.subscribe is never actually assigned to the variable.

So every schema change adds new subscriptions and old ones still remain active and trigger.

### Expected behavior

removed controls are unsubscribed

<in description>

### Screenshots

_No response_

3.3.0

### Framework

Angular

### RendererSet

### Additional context

_No response_

LukasBoll:
Previously, the subscription to the jsonForm state in Angular was never unsubscribed. With this commit, all subscriptions will be unsubscribed when the component is destroyed.

Closes #2354

### DIFF
--- a/packages/angular/src/library/base.renderer.ts
+++ b/packages/angular/src/library/base.renderer.ts
@@ -22,18 +22,30 @@
   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
   THE SOFTWARE.
 */
-import { Directive, Input } from '@angular/core';
+import { Directive, Input, OnDestroy } from '@angular/core';
 import {
   JsonSchema,
   OwnPropsOfRenderer,
   UISchemaElement,
 } from '@jsonforms/core';
+import { Subscription } from 'rxjs';
 
 @Directive()
-export class JsonFormsBaseRenderer<T extends UISchemaElement> {
+export class JsonFormsBaseRenderer<T extends UISchemaElement>
+  implements OnDestroy
+{
   @Input() uischema: T;
   @Input() schema: JsonSchema;
   @Input() path: string;
+  subscriptions: Subscription = new Subscription();
+
+  public addSubscription(subscription: Subscription | Subscription[]) {
+    if (Array.isArray(subscription)) {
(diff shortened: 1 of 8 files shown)


packages/angular/src/library/base.renderer.ts
```diff
@@ -22,18 +22,30 @@
...
 
 @Directive()
-export class JsonFormsBaseRenderer<T extends UISchemaElement> {
+export class JsonFormsBaseRenderer<T extends UISchemaElement>
+  implements OnDestroy
+{
   @Input() uischema: T;
   @Input() schema: JsonSchema;
   @Input() path: string;
+  subscriptions: Subscription = new Subscription();
```
lucas-koehler, line 40: The subscriptions variable should only be accessible by subclasses
```suggestion
  protected subscriptions: Subscription = new Subscription();
```

packages/angular/src/library/base.renderer.ts
```diff
@@ -22,18 +22,30 @@
...
+
+  public addSubscription(subscription: Subscription | Subscription[]) {
```
lucas-koehler, line 42: Similar to the variable, only subclasses should add subscriptions.
```suggestion
  protected addSubscription(subscription: Subscription | Subscription[]) {
```

lucas-koehler (commented):
Thanks for this PR @LukasBoll . The changes already look pretty good to me :sparkles: 
I have just two minor suggestions inline to make the base renderer's subscriptions `protected`.

manohara-kp:
ERROR TypeError: this.subscription is undefined
    Angular 7
    JsonFormsRendererComponent_Conditional_0_Template json-forms-renderer.component.html:2
    Angular 58
    RxJS 6
    Angular 25
    TestTreeStatusComponent_Conditional_1_Template test-tree-status.component.html:9
    Angular 4
    TestTreeStatusComponent_Template test-tree-status.component.html:2
    Angular 26 Am facing this error could you please help me ?

PR #2383, opened 2024-10-24
Make HTML ID generation fast

jwueller:
Replaces HTML ID generation with a simple monotonic sequence to avoid several expensive hash lookups for every generated ID.

This entirely eliminates the chance for collisions under regular circumstances, and the default seed even performs better in some adverse circumstances, like multiple instances of the same library sharing an ID namespace on the same page.

The only potential downside is that IDs might not be particularly semantically helpful, but that can be mitigated by just appending that information instead.

This might be a breaking change if this module is considered a public API.

Fixes issues like this: https://jsonforms.discourse.group/t/expensive-id-generation-mechanism-with-createid-and-isuniqueid/1737

### DIFF
--- a/packages/angular/src/library/abstract-control.ts
+++ b/packages/angular/src/library/abstract-control.ts
@@ -29,7 +29,6 @@ import {
   JsonFormsState,
   JsonSchema,
   OwnPropsOfControl,
-  removeId,
   StatePropsOfControl,
 } from '@jsonforms/core';
 import { Component, Input, OnDestroy, OnInit } from '@angular/core';
@@ -149,7 +148,6 @@ export abstract class JsonFormsAbstractControl<
 
   ngOnDestroy() {
     super.ngOnDestroy();
-    removeId(this.id);
   }
 
   isEnabled(): boolean {
--- a/packages/angular/src/library/jsonforms.component.ts
+++ b/packages/angular/src/library/jsonforms.component.ts
@@ -32,7 +32,7 @@ import {
   ViewContainerRef,
 } from '@angular/core';
 import {
-  createId,
+  nextId,
   isControl,
   getConfig,
   JsonFormsProps,
@@ -143,7 +143,7 @@ export class JsonFormsOutlet
         const controlInstance = instance as JsonFormsControl;
         if (controlInstance.id === undefined) {
           const id = isControl(props.uischema)
-            ? createId(props.uischema.scope)
+            ? nextId() + props.uischema.scope
             : undefined;
           (instance as JsonFormsControl).id = id;
         }
--- a/packages/core/src/util/ids.ts
+++ b/packages/core/src/util/ids.ts
@@ -23,30 +23,21 @@
   THE SOFTWARE.
 */
 
-const usedIds: Set<string> = new Set<string>();
+let idNamespace: string;
+let idIndex: number;
 
-const makeId = (idBase: string, iteration: number) =>
-  iteration <= 1 ? idBase : idBase + iteration.toString();
-
-const isUniqueId = (idBase: string, iteration: number) => {
-  const newID = makeId(idBase, iteration);
-  return !usedIds.has(newID);
+export const seedIds = (namespace = 'jsonforms', index = 0) => {
+  idNamespace = namespace;
+  idIndex = index;
 };
 
-export const createId = (proposedId: string) => {
-  if (proposedId === undefined) {
-    // failsafe to avoid endless loops in error cases
-    proposedId = 'undefined';
-  }
-  let tries = 0;
-  while (!isUniqueId(proposedId, tries)) {
-    tries++;
-  }
-  const newID = makeId(proposedId, tries);
-  usedIds.add(newID);
-  return newID;
-};
+// This has a salt in the unlikely case that someone bundled multiple instances
+// of this module on a single site.
+seedIds(`jsonforms${Math.random().toString(32).slice(2)}`);
 
-export const removeId = (id: string) => usedIds.delete(id);
-
-export const clearAllIds = () => usedIds.clear();
+/**
+ * Generate an ID that's unique within the current execution context. This is
+ * intended for HTML ID generation. Does not guarantee stability across SSR
+ * boundaries!
+ */
+export const nextId = () => `:${idNamespace}:${(idIndex++).toString(36)}:`;
--- a/packages/core/test/mappers/cell.test.ts
+++ b/packages/core/test/mappers/cell.test.ts
@@ -25,7 +25,7 @@
 import test from 'ava';
 import * as _ from 'lodash';
 import * as Redux from 'redux';
-import { clearAllIds, createAjv, validate } from '../../src/util';
+import { seedIds, createAjv, validate } from '../../src/util';
 import { UPDATE_DATA, UpdateAction } from '../../src/actions';
 import configureStore from 'redux-mock-store';
 import { JsonFormsState } from '../../src/store';
@@ -263,7 +263,7 @@ test('mapStateToCellProps - data', (t) => {
 });
 
 test('mapStateToCellProps - id', (t) => {
-  clearAllIds();
+  seedIds();
   const ownProps = {
     uischema: coreUISchema,
     id: '#/properties/firstName',
--- a/packages/core/test/mappers/renderer.test.ts
+++ b/packages/core/test/mappers/renderer.test.ts
@@ -64,7 +64,7 @@ import {
   mapStateToOneOfEnumControlProps,
   mapStateToOneOfProps,
 } from '../../src/mappers';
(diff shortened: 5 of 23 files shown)


lucas-koehler:
Hello @jwueller, thanks for the PR :heart: and sorry for the late reply!
Unfortunately, we cannot accept this behavior change because it is an intentional feature that generated IDs are predictable.


## Alternative solution

We discussed this with the team and came up with a potential solution that would allow you and others to override the ID generation when necessary:

Currently, the id-generation related methods `createId`, `removeId`, `clearAllIds` are directly imported from `packages/core/src/util/ids.ts`. This makes it currently impossible to override the implementations for adopters.

To make them overridable, we propose exporting a constant singleton object `Id` from `packages/core/src/util/ids.ts` and then calling the methods from the object. This allows to import the object and override its methods.

This requires adapting all current calls of the id-methods to call them from the new `Id` object

For this, introduce the object in `packages/core/src/util/ids.ts`:
```typescript
// [...]
export const Id = {
  createId: createId,
  removeId: removeId,
  clearAllIds: clearAllIds
}
```

Callers of the Id methods then import the `Id` object and call the methods from there, like so:

```typescript
// OLD
import { createId } from '@jsonforms/core';
// [...]
const newId = createId();

// NEW
import { Id } from '@jsonforms/core';
// [...]
const newId = Id.createId();
```

To override and implementation, i.e. of `createId`:
```typescript
import { Id } from '@jsonforms/core';

function myCreateId() {
  // [...]
}

Id.createId = myCreateId;
```

If you want to contribute this, we would gladly accept this refactoring. It should then allow you to provide your own random-based ID generation in your project :)

jwueller:
> Unfortunately, we cannot accept this behavior change because it is an intentional feature that generated IDs are predictable. Predictable IDs are required because some downstream adopters depend on them for their code or CSS.

@lucas-koehler Maybe you can help me understand, because I'm slightly confused. The IDs are still very much predictable. Or are we talking about the randomized namespace prefix? That can be overwritten, like I do in the tests. The randomized default namespace just reduces the risk of naming collisions in edge cases, but is not strictly required.

Or are you saying that IDs need to retain their exact structure? Because that structure inherently can't scale well, because it requires coordination between IDs, which will always be slow. Also, relying on IDs to have a certain structure seems like a significant anti-pattern, considering they have to be unique on the entire page, and can't be predicted reliably across multiple forms already anyway.

> ## Clean solution

Yes, passing the generator function into the config would certainly be the best solution, but is also the most disruptive. I just wanted fast generation that we can actually use ASAP, because the current ID generator scales so poorly, that it takes up the majority of the entire application compute time once you have enough forms on the page.

> To make them overridable, we propose exporting a constant singleton object `Id` from `packages/core/src/util/ids.ts` and then calling the methods from the object. This allows to import the object and override its methods.

This is an okay compromise, but will still be very slow by default. I can make a separate PR for that if it's the only acceptable solution, since the main goal of this one is to make it fast.

sdirix:
Hi @jwueller,

You're right, the suggested `id` mechanism is deterministic.

>Or are you saying that IDs need to retain their exact structure? Because that structure inherently can't scale well, because it requires coordination between IDs, which will always be slow. Also, relying on IDs to have a certain structure seems like a significant anti-pattern, considering they have to be unique on the entire page, and can't be predicted reliably across multiple forms already anyway.

Yes, we can't change the ID generation in a minor version as this **will** break downstream in an unexpected manner. We know adopters who rely on the current unique ids to programmatically identify and/or style special inputs.

In many use cases this is not an anti-pattern as adopters who rely on this mechanism are using JSON Forms in a predictable manner.

---

In practice, the use case of rendering thousands of inputs at the same time in which the hash lookup penalty becomes apparent is the actual niche use case. In the more complex renderer sets, rendering is then already so expensive that the use case is out of question already anyway.

I did a quick performance test:
 - Invoking `createId` with 1000 very long identical ids completes in ~120 ms
 - Invoking `createId` with 10000 very long identical ids completes in ~12s
 - Invoking `createId` with 10000 very long ids of which 1000 are spread out duplicates completes in ~125 ms
 - Invoking `createId` with 200 very long unique ids completes in ~0.5ms
 - Invoking `createId` with 200 very long unique ids, each time handed over 10 times, completes in ~5ms

So even with 1000 duplicates, the penalty is barely noticeable compared to the rendering time needed to render 1000 inputs. The only very bad case is once you have over 1000+ identical duplicates. This is certainly an niche use case. Therefore it's not a good enough reason to completely change the id behavior within a minor version.

I'm fine with changing the id generation with the next major version though, especially when making it configurable.

---

> Yes, passing the generator function into the config would certainly be the best solution, but is also the most disruptive. I just wanted fast generation that we can actually use ASAP, because the current ID generator scales so poorly, that it takes up the majority of the entire application compute time once you have enough forms on the page.

This is really only true in use cases where you have >1000 exactly duplicated ids. What kind of use case do you have?

> This is an okay compromise, but will still be very slow by default. I can make a separate PR for that if it's the only acceptable solution, since the main goal of this one is to make it fast.

We would happily accept that PR as this just makes JSON Forms better for everyone by adding more options.

---

There is also the other alternative to speed up the current implementation without changing its behavior. 

For example using the following improved implementation we avoid long string hashes and instead use a more performant number set

```ts
const idMap = new Map<string, Set<number>>();

function createId(proposedId) {
  if (!idMap.has(proposedId)) {
    idMap.set(proposedId, new Set());
  }

  const usedIndices = idMap.get(proposedId);

  let index = 0;
  while (usedIndices.has(index)) {
    index++;
  }

  const newId = index === 0 ? proposedId : `${proposedId}${index}`;

  usedIndices.add(index);

  return newId;
}
```

Using this approach, I get the following performance numbers:

 - 1000 very long identical ids completes in ~5ms
 - 10000 very long identical ids completes in ~450ms
 - 10000 very long ids of which 1000 are spread out duplicates completes in ~15ms
 - 200 unique ids completes in ~0.1ms
 - 200 unique ids, each time handed over 10 times, completes in ~3.5ms

The worst cases is significantly improved by a factor of 2400% from 12 seconds to 450ms. Of course the delete use case becomes a bit more complicated.

So if this is sufficient performance for you, feel free to contribute this improvement with testcases, as it's more complicated than before.

jwueller:
Hi @sdirix,

> Yes, we can't change the ID generation in a minor version as this **will** break downstream in an unexpected manner. We know adopters who rely on the current unique ids to programmatically identify and/or style special inputs.

Yes, this is definitely a breaking change if you consider IDs part of the public API. Although if data about the property name is required by CSS or other post-processing, then it might be better to move those into `data-*` attributes, or similar in the future, since those don't have to deal with collisions.

> In many use cases this is not an anti-pattern as adopters who rely on this mechanism are using JSON Forms in a predictable manner.

Are they really that predictable, though? The fact that IDs depend on the order that they were generated in makes them pretty hard to predict if you have more than one form on a page.

For example, since the IDs basically always boil down to just a property name and a numeric suffix, something as simple as `#/properties/name` would be very likely to collide between unrelated schemas, leading to unpredictable IDs.

> In practice, the use case of rendering thousands of inputs at the same time is already a nice use case.

I agree that this might not be the primary use-case, but in our case we have a lot of list items, each containing rather simple forms, but they tend do be similar in structure. This means that there is also a lot of ID collisions, because most of them happen to use common names for their properties.

> I did a quick performance test:

It seems that you're getting much better numbers than this example did, which is surprising to me:
https://jsonforms.discourse.group/t/expensive-id-generation-mechanism-with-createid-and-isuniqueid/1737

> This is really only true in use cases where you have >1000 identical ids. What kind of use case do you have?

Lots of small but similar forms in a larger list structure, as described above.

> There is also the other alternative to speed up the current implementation without changing its behavior.

Yes, it's definitely possible to speed up the current externally visible behavior. Would you prefer a separate PR for a backwards-compatible solution?

sdirix:
> I agree that this might not be the primary use-case, but in our case we have a lot of list items, each containing rather simple forms, but they tend do be similar in structure. This means that there is also a lot of ID collisions, because most of them happen to use common names for their properties.

Interesting, you are rendering all these forms at the same time? I would have expected more of a tree-master with detail approach in use cases like that.

> It seems that you're getting much better numbers than this example did, which is surprising to me:
https://jsonforms.discourse.group/t/expensive-id-generation-mechanism-with-createid-and-isuniqueid/1737

The discussion post is not very detailed. "1000 forms" could mean anything, from 1000 identical ids up to 10000, 100000 etc. Besides that it obviously depends on the executing machine too.

> Yes, it's definitely possible to speed up the current externally visible behavior. Would you prefer a separate PR for a backwards-compatible solution?

I'm fine either way, i.e. a new PR or force-pushing here.

First I would recommend opening a PR just changing the id structure, i.e. with the suggested object approach. That would already allow overwriting for your case.

PR #2387, opened 2024-11-11
feat: improve support for `minItems` and `maxItems` for array layout and control

Typiqally:
Added improved support for the `minItems` and `maxItems` properties, to allow for more fine-grained control and validation in ArrayControl and ArrayLayout applications. I have also added additional tests to cover these properties.

----

While working on this, I realized that the current handling of array (or items) schemas is too restrictive for custom implementations. Specifically, the schema is being completely replaced by the items schema, which typically only includes a basic type description and hides important details like the `min` and `max` item properties from the original array schema.

https://github.com/eclipsesource/jsonforms/blob/fcef896c8927269d3f3ed609a7255651085b193d/packages/core/src/mappers/renderer.ts#L798-L818

To address these limitations, I see two possible approaches:

1. Add a `parentSchema` property to preserve the original schema, keeping the existing behavior while easing the constraints. This might introduce some confusion around naming.
2. Store the original schema in the `schema` property and put the resolved schema in a new `arraySchema` or `itemsSchema` property. However, this would change the current behavior, requiring migration.

I’d appreciate your thoughts on these options.

### DIFF
--- a/packages/core/src/mappers/renderer.ts
+++ b/packages/core/src/mappers/renderer.ts
@@ -785,6 +785,7 @@ export interface ControlWithDetailProps
  */
 export interface StatePropsOfArrayControl
   extends StatePropsOfControlWithDetail {
+  arraySchema: JsonSchema;
   childErrors?: ErrorObject[];
   minItems?: number;
   maxItems?: number;
@@ -813,6 +814,7 @@ export const mapStateToArrayControlProps = (
     path,
     uischema,
     schema: resolvedSchema,
+    arraySchema: schema,
     childErrors,
     renderers: ownProps.renderers || getRenderers(state),
     cells: ownProps.cells || getCells(state),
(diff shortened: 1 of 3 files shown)


sdirix:
I agree that the current situation is not optimal. I would like to suggest to add a new prop `arraySchema` in which the schema referring to the `array` is handed over. For now we should not adapt the `schema` prop, as this will certainly lead to backward compatibility issues with custom renderers.

We could think of a migration path with a major release or just keep the current prop naming, even if it's not ideal.

sdirix:
Yes, let's refactor our code so we use `arraySchema` where appropriate. If a binding did hand over `minItems` then we should still do that to not break anyone on the update.

Typiqally:
@sdirix, I have removed the props and deprecated the original `minItems` property in favour of using `arraySchema.minItems`. I don't see an usage of this property anywhere in the internal packages. Only the Vuetify package seems to be using something related to `minItems`, but they get around these props by resolving the schema manually in the component itself.

sdirix (approved):
Thanks :heart:

Issue #2389, opened 2024-11-20
Allow jsonforms core to properly resolve errorsAt when errors are provided by ajv-errors plugin 


kchobantonov:
### Is your feature request related to a problem? Please describe.

When the ajv-errors plugin is installed with ajv then the following schema 

```
{
  type: 'object',
  properties: {
    username: {
      type: 'string',
      title: 'username',
      description: 'Login Name',
    },
    password: {
      type: 'string',
      title: 'password',
      format: 'password',
      description: 'Login password',
    },
  },
  required: ['username', 'password'],
  errorMessage: {
    required: {
      username: 'username is required',
      password: 'password is required',
    },
  },
}
```

will produces the errors for the username and password when those are missing but the corresponding UI fields for username and password won't be ever marked as invalid.

### Describe the solution you'd like

Change the errorsAt function to return the errors wrapped by ajv-errors but substituted with the correct messsage ( e.g. the message from the ErrorObject with keyword errorMessage

### Describe alternatives you've considered

none identified so far

### Package

Core

### Additional context

will create PR with demo

PR #2390, opened 2024-11-20
fix (#2389)

kchobantonov:
fix: include any wrapped errors from ajv-errors plugin that are related to the instancePath (#2389)


kchobantonov:
@sdirix please review - the demo is with this example - [Login](https://deploy-preview-2390--jsonforms-examples.netlify.app/vue-vuetify/#login) - you can edit the schema and provide different error message and once applied then then the error message on the UI will change as well - I was also thinking about the i18n but the errorAt is low level function and perhaps handling the i18n could be done at another AJV plugin level (like the ajv-i18n integrated with jsonforms translator)

kchobantonov:
@sdirix yes I think probably going via a middleware might be better for even one more reason - to support custom keywords where the keyword and its validation is for specific field like the required but then the change in the jsonschema needs to somehow pass the middleware object into the getErrorAt function somehow. For now yes I can go with a ajv plugin and adjust the errors in the case of errorMessage and how it is implemented but in the future we might need to be able to provide the getErrorAt function in the jsonforms middleware so that it can be overriden in the cases where errors with custom keywords are generated for specific field and jsonforms core needs to be pick those as well. 

What do you think - adjusting the getErrorAt to be able to invoke a jsonforms middleware new function that by default will do what it does today ?

kchobantonov:
@sdirix here is how I'm doing that https://github.com/kchobantonov/jsonforms-vuetify-webcomponent/blob/93842c9ab85a1e0b62b6a66e5d7292bfb28e3af4/packages/jsonforms-vuetify-renderers/src/core/ajv-i18n/index.ts#L33

It is done as AJV plugin that overrides ajv compile function in order to intercept any additional errors that are procuded by plugins such as ajv-errors - also the plugin combines the ajv-i18n as well as the jsonform translate function

PR #2391, opened 2024-11-22
[FIX-eclipsesource#2386] Add support for Angular 19

laurentgrangier:


laurentgrangier:
@lucas-koehler or @sdirix Could you please take a look at the merge request? I don't understand what are the problems with the build.

sdirix:
The issue seems to be [`here`](https://github.com/eclipsesource/jsonforms/blob/fcef896c8927269d3f3ed609a7255651085b193d/packages/angular-material/build-example.js#L49-L64). The example app build does not enter the `success` state and therefore returns `1`, indicating an error.

We're using an Angular builder here, it likely changed a bit between versions.

PR #2392, opened 2024-11-26
fix complex renderers

kchobantonov:
update demo app, fix oneOf, allow passing slots to the dispatcher renderer, use vuetify widgets for integer and number inputs


kchobantonov:
@sdirix please review - this fixes oneOf renderer where it could clear out the whole data - check for example - oneOf - Preselection for Primitives (Issue 1273 ) example and try to switch between diferent oneOf when they are populated -once the dialog is clicked to apply the selectio the data will not be an object

the other changes are some stuff that I'm looking to do and are part of the excercise - be able to have a generic editor for jsonschema - e.g. the schema to be the jsonschema meta model and then to be able to create a new jsonschema with that - other are other stuff that I'm working on fixing but are not yet ready - like having a MixedControl which can handle mixed types by asking you to select the type of the element before showing you the actual control to work with the data when creating a new data and if the data is present then automatically selecting the correct type.

One thing that I'm having an issue is with the jsonforms core resolveSchema method which when it can't find the property it will return undefined and in the case of dynamic schema you do not really have a structure 

`
{
"type: "object",
"additionalProperties": true
} 
`

and we try to find a property in that schema then the resolveSchema does not give me the schema that I have provided. Locally I'm doing some hacks which I do not really like for example attaching a Symbol keys to the schema that points to a function that can be then invoke by the resolveSchema in that particular case but that looks too hacky. If you have any suggestions please let me know.

Also are you interested in such renderer - MixedRenderer or maybe we can call it UnionRenderer - anyway let me know if such one could go to the jsonform vuetify render repo - otherwise I won't spent much time to even prepare such PR

packages/vue-vuetify/src/util/composition.ts
```diff
@@ -141,7 +142,7 @@ export const useVuetifyControl = <
   },
   I extends {
     control: ComputedRef<T>;
-  } & DispatchPropsOfControl,
+  } & (DispatchPropsOfControl | DispatchPropsOfMultiEnumControl),
```
sdirix, line 145: Seems weird to add this type here. What is the rationale?

packages/vue/src/components/JsonForms.vue
```diff
@@ -52,8 +52,9 @@ export default defineComponent({
   },
   props: {
     data: {
-      required: true,
+      required: false,
       type: [String, Number, Boolean, Array, Object] as PropType<any>,
+      default: undefined, // allow for cases like mixed schema and the data in not yet defined
```
sdirix, line 57: Is this default needed? If not required, `undefined` is the default anyways, right?

sdirix (changes requested):
The changes look mostly good! Can you take a look at the CI checks and fix them?

packages/vue/src/components/JsonForms.vue
kchobantonov: yes that is correct let me adjust that and remove the comment there as well then, althought we have default: undefined for uischema as well as schema - do you want me to clean that up as well ?

kchobantonov: when I try to remove the default then I got some lint warning -so just cleaned up the comment since it is more noise than clarity 

Prop 'data' requires default value to be set.eslint[vue/require-default-prop](https://eslint.vuejs.org/rules/require-default-prop.html)

kchobantonov:
@sdirix all fixed and those that needed clarification please find those in the corresponding sections.

packages/vue-vuetify/src/util/composition.ts
sdirix: Hmm I see. The core issue is that we do not expose `handleChange` to these controls although I would have expected it. See for example the React renderers in which we not only do the specialized `addIterm` etc. but also expose the usual `handleChange`. 

I'm fine with this for this PR but we should rework in the future I think

sdirix (approved):
Works for me

PR #2400, opened 2024-11-29
React material table cell validationMode

### LINKED ISSUE #2398: ValidateAndHide only partially working for react material table cell
### Describe the bug

Validation mode `'ValidateAndHide'` does not hide error messages in array object cells (rendered by the React MaterialTableControl)

### Expected behavior

The error should be hidden

1. Create a schema with an array of objects, where the object has required properties

   e.g.

   ```json
   {
       "type": "object",
       "properties": {
           "someArray": {
               "type": "array",
               "items": {
                   "type": "object",
                   "properties": {
                       "name": {"type": "string"},
                       "age": {"type": "number"},
                   },
                   "required": ["name"],
               },
           },
       },
   }
   ```

2. Render the schema with the react material renderers and validation mode `'ValidateAndHide'`

   ```jsx
   <JsonForms
     {...otherProps}
     validationMode={'ValidateAndHide'}
   />
   ```

### Screenshots

| validationMode | screenshot |
| --- | --- |
| NoValidation | NoValidation shows grey box input box and no red error message |
| ValidateAndHide | ValidateAndHide shows grey input box with red error message |
| ValidateAndShow | ValidateAndShow shows red input box with red error message |

v3.4.1

### Package

Core, React Material Renderers

### Additional context

I'm not very familiar with the codebase so apologies if this is a red herring, but it looks like the control code which correctly hides the error [uses getErrorAt](https://github.com/eclipsesource/jsonforms/blob/fcef896c8927269d3f3ed609a7255651085b193d/packages/core/src/mappers/renderer.ts#L532) from [src/store/jsonFormsCore](https://github.com/eclipsesource/jsonforms/blob/fcef896c8927269d3f3ed609a7255651085b193d/packages/core/src/store/jsonFormsCore.ts#L40) and this respects validation mode.

Whereas the [cell error label](https://github.com/eclipsesource/jsonforms/blob/fcef896c8927269d3f3ed609a7255651085b193d/packages/material-renderers/src/complex/MaterialTableControl.tsx#L253) instead [uses errorsAt](https://github.com/eclipsesource/jsonforms/blob/fcef896c8927269d3f3ed609a7255651085b193d/packages/material-renderers/src/complex/MaterialTableControl.tsx#L180) which comes from [src/util/errors](https://github.com/eclipsesource/jsonforms/blob/fcef896c8927269d3f3ed609a7255651085b193d/packages/core/src/util/errors.ts#L66).

There may be a good reason for that of course which I don't understand - but I'd expect cells and controls both to respect that `validationMode` property

0mpurdy:
Fix #2398

### DIFF
--- a/packages/material-renderers/test/renderers/MaterialTableControl.test.tsx
+++ b/packages/material-renderers/test/renderers/MaterialTableControl.test.tsx
@@ -1,0 +1,136 @@
+/*
+  The MIT License
+
+  Copyright (c) 2017-2019 EclipseSource Munich
+  https://github.com/eclipsesource/jsonforms
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
+  furnished to do so, subject to the following conditions:
+
+  The above copyright notice and this permission notice shall be included in
+  all copies or substantial portions of the Software.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+import { JsonSchema7 } from '@jsonforms/core';
+import * as React from 'react';
+
+import { materialCells, materialRenderers } from '../../src';
+import Enzyme, { mount, ReactWrapper } from 'enzyme';
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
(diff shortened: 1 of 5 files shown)


0mpurdy:
| Before | After |
| --- | --- |
| before | after |

packages/material-renderers/test/renderers/MaterialTableControl.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+        schema={schemaWithRequired}
+        uischema={uischema}
+        renderers={materialRenderers}
+        cells={materialCells}
+        validationMode='ValidateAndShow'
+      />
+    );
+    const messageFormHelperText = wrapper.find(FormHelperText).at(0);
+    expect(messageFormHelperText.text()).toBe(
+      "must have required property 'message'"
```
0mpurdy, line 89: @sdirix Thought I'd at least attempt to add relevant unit tests - there didn't seem to be any for `MaterialTableControl`

This fails - not sure if validation just doesn't run for some reason

If you know why let me know and I'll fix it, otherwise I can either skip this test, or remove this file all together.

Testing manually with the examples seems to be working as expected

0mpurdy: Interestingly, the test works with `message: { type: 'string', minLength: 3 }`

So that's another option, but doesn't really explain why required is not being calculated correctly in this context

lucas-koehler: In the unit test, did you add the required property as an empty string or leave out the property completely?
If it's empty, the required check does not fail. In the ui, the renderer unsets the field to undefined when it is empty. That could be a reason for different results.

I also tested it in the example app and there showing the required error on an array item still worked fine (I locally extended the arrays.ts example for this).

Could you re-add a unit test for this? If it still doesn't run as expected, I can have a look then :)

lucas-koehler:
@0mpurdy Thanks for the contribution :heart: 
The changes already look quite good to me. Please have a look at my answer to your comment https://github.com/eclipsesource/jsonforms/pull/2400#discussion_r1869718844

lucas-koehler (approved):
LGTM now! Thanks for the contribution and all the tests :sparkles:

Issue #2401, opened 2024-12-01
createDefaultValue doesn't follow anyOf, oneOf, allOf


kchobantonov:
### Describe the bug

1. createDefaultValue returns empty object in the "anyOf" example when new element is added in "Addresses or Users (AnyOf Array Items)"
2. createDefaultValue should also pass the rootSchema when it tries to resolve the the schema based on schema.$ref

### Expected behavior

the createDefaultValue should have returned empty array rather than empty object

we get 

```
{
  "addressOrUser": {
    "street_address": "1600 Pennsylvania Avenue NW",
    "city": "Washington",
    "state": "DC"
  },
  "addressesOrUsersAnyOfItems": [
    {}
  ]
}

```
instead of 

```
{
  "addressOrUser": {
    "street_address": "1600 Pennsylvania Avenue NW",
    "city": "Washington",
    "state": "DC"
  },
  "addressesOrUsersAnyOfItems": [
    []
  ]
}
```

### Steps to reproduce the issue

1. Go to generated example app
2. Select Example 'anyOf'
3. Scroll down to 'Addresses or Users (AnyOf Array Items)'
4. Click on the '+' to add new element in "Addresses or Users (AnyOf Array Items)" array
5. See that added element under addressesOrUsersAnyOfItems is an empty object rather than empty array

### Which Version of JSON Forms are you using?

3.4.1

### Package

Core

PR #2402, opened 2024-12-01
fixes createDefaultValue doesn't follow anyOf, oneOf, allOf #2401

kchobantonov:
fixing issue #2401


sdirix (changes requested):
> @sdirix check the fixed example here https://deploy-preview-2402--jsonforms-examples.netlify.app/react-material/#anyOf

I don't see any difference in the example between `master` and this PR

packages/core/src/mappers/renderer.ts
```diff
@@ -196,8 +200,44 @@ export const createDefaultValue = (
...
   } else if (hasType(resolvedSchema, 'null')) {
     return null;
+  } else if (
+    schema.oneOf &&
+    Array.isArray(schema.oneOf) &&
+    schema.oneOf.length > 0
+  ) {
+    for (const combineSchema of schema.oneOf) {
+      const result = createDefaultValue(combineSchema, rootSchema, undefined);
+      if (result !== undefined) {
```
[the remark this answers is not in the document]
kchobantonov: yes you are right - overlooked this one - let me see how to fix it without making the change not backward compatible

packages/core/src/mappers/renderer.ts
```diff
@@ -196,8 +200,44 @@ export const createDefaultValue = (
...
+        // return the first one that have type information
+        return result;
```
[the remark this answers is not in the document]
kchobantonov: yes I agree with that

kchobantonov:
@sdirix please review again

packages/core/src/mappers/renderer.ts
```diff
@@ -196,9 +200,62 @@ export const createDefaultValue = (
     return extractDefaults(resolvedSchema, rootSchema);
   } else if (hasType(resolvedSchema, 'null')) {
     return null;
-  } else {
-    return {};
   }
+
+  let combinatorDefault = undefined;
+
+  combinatorDefault = createDefaultValueForCombinatorSchema(
```
sdirix, line 207: Instead of this additional method we could create a `doCreateDefaultValue` instead which behaves the same as the current one but returns `undefined` if there is no default value.

In `createDefaultValue` we then only call the `doCreateDefaultValue` and return `{}` instead of `undefined`.

This keeps backward compatibility but avoids the magic symbol

kchobantonov: sure will do

kchobantonov: fixed

sdirix (approved):

PR #2406, opened 2024-12-12
Allow using projects to properly tree-shake lodash

FlorianSW:
Instead of importing top-level module, consistently use deep-imports for lodash. This allows projects to properly tree-shake lodash in their build is substantially reduce the size of the resulting bundle.


lucas-koehler (approved):
Hi @FlorianSW , thanks for this improvement :heart: 
The changes LGTM :+1:

PR #2408, opened 2024-12-17
customizations for arraylist, fix oneOf tab renderer disabled property

kchobantonov:
add arrayList.title missed from the migration, add additional customizations for array renderers, fix one tab disable property, fix calling Generate.uiSchema


kchobantonov:
@sdirix take a look

sdirix (approved):
LGTM. Thanks!

PR #2409, opened 2024-12-30
Mixed support

kchobantonov:
1. provide support for mixed types
2. fix clearing of inputs when not under an object - e.g. string once cleared will become empty string rather than to return undefined to remove the property from the parent object
3. fix date controls, now the date can show year, year/month and etc. depending on the control option views
4. improve the performance to convert dayjs format into maska format function
5. fixing the :key to at least have the collection size into the key so when the size changes the vue will not reuse html elements leaving attributes like id unchanged
6. add more examples to show how the mixed control can be used
7. use arraySchema from the mapped values rather than using Resolve
8. minor UI changes to use less space
9. core changes : use the provided schema when the resolve failed, make sure that boolean is valid schema as well - e.g. items: true is valid schema (Note: the last change for the items to be boolean was reverted back since more code should be adjusted - maybe a future improvement)
10. introduce a new config - allowAdditionalPropertiesIfMissing - this allows to show the additioanalProperties UI when the additionalProperties and patternProperties do not exist. The reason is that by default if the additionalProperties is missing from the schema the default should be considered as true but the UI component does not respect that default - e.g. it needs to be explicitly states as additionalProperties: true if the UI control is needed without the allowAdditionalPropertiesIfMissing config


kchobantonov:
@sdirix  please review - checkout the new examples
1. https://deploy-preview-2409--jsonforms-examples.netlify.app/vue-vuetify/#mixed
2. https://deploy-preview-2409--jsonforms-examples.netlify.app/vue-vuetify/#mixed-object
3. https://deploy-preview-2409--jsonforms-examples.netlify.app/vue-vuetify/#json-editor
4. https://deploy-preview-2409--jsonforms-examples.netlify.app/vue-vuetify/#jsonschema

For example 3 the easier way to see that is to copy any example json data, provided it in the example ( Data tab - save it) and see the result in the Demo tab

sdirix:
Thanks for the contribution :heart: I'll take a look within January :+1:

packages/core/src/mappers/renderer.ts
```diff
@@ -186,7 +186,7 @@ export const doCreateDefaultValue = (
   rootSchema: JsonSchema
 ) => {
   const resolvedSchema =
-    typeof schema.$ref === 'string'
+    typeof schema?.$ref === 'string'
```
sdirix, line 189: schema should not be `undefined` here. If it is. then we should adjust the parameter type

packages/core/src/util/resolvers.ts
```diff
@@ -123,12 +123,20 @@ const resolveSchemaWithSegments = (
...
+    return schema;
+  }
+
   if (isEmpty(schema)) {
     return undefined;
   }
 
   // use typeof because schema can by of any type - check singleSegmentResolveSchema below
-  if (typeof schema.$ref === 'string') {
+  if (typeof schema?.$ref === 'string') {
```
sdirix, line 139: `schema` should exist here. If it doesn't we need to adjust the parameter type

packages/core/src/util/resolvers.ts
```diff
@@ -123,12 +123,20 @@ const resolveSchemaWithSegments = (
   pathSegments: string[],
   rootSchema: JsonSchema
 ): JsonSchema => {
+  if (
+    typeof schema === 'boolean' &&
+    (!pathSegments || pathSegments.length === 0)
+  ) {
+    // add the case where the schema can be true value for example: items: true or additionalProperties: false
+    return schema;
+  }
```
sdirix, line 132: What is the use case of this? Who hands over segments to resolve if the schema is just a boolean already anyway?

packages/vue-vuetify/src/complex/AllOfRenderer.vue
```diff
@@ -20,7 +20,7 @@
         />
         <dispatch-renderer
           v-for="(allOfRenderInfo, allOfIndex) in allOfRenderInfos"
-          :key="`${control.path}-${allOfIndex}`"
+          :key="`${control.path}-${allOfRenderInfos.length}-${allOfIndex}`"
```
sdirix, line 23: Why adapt this key?

packages/vue-vuetify/src/complex/ArrayControlRenderer.vue
```diff
@@ -78,14 +78,17 @@
             <tbody>
               <tr
                 v-for="(element, index) in control.data"
-                :key="`${control.path}-${index}`"
+                :key="`${control.path}-${control.data.length}-${index}`"
```
sdirix, line 81: Why adapt this key?

packages/vue-vuetify/src/complex/ArrayControlRenderer.vue
```diff
@@ -78,14 +78,17 @@
...
               >
                 <td
-                  v-for="propName in getValidColumnProps(control.schema)"
+                  v-for="propName in validColumnProps"
                   :key="
                     composePaths(
-                      composePaths(control.path, `${index}`),
+                      composePaths(
+                        control.path,
+                        `${validColumnProps.length}-${index}`,
```
sdirix, line 90: Why add the columnProps length to the path

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -0,0 +1,491 @@
...
+      // return only if we were able to determine all types, otherwise return the default
+      return Array.from(enumTypes).filter((type) => type !== null) as string[];
+    }
+  }
+
+  // return any
+  return ['array', 'boolean', 'integer', 'null', 'number', 'object', 'string'];
+}
+
+const createMixedRenderInfos = (
```
sdirix, line 206: This function seems very dangerous as it potentially modifies the handed over `schema`. We should never modify the `schema` handed over by the user as this is unexpected behavior.

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -0,0 +1,491 @@
...
+
+    return {
+      schema,
+      uischema,
+      label: `${schema.type}`,
+    };
+  });
+};
+
+export function getJsonDataType(value: any): string | null {
```
sdirix, line 310: I think we already have a util function like this somewhere in the code base. Likely in the `schema` generator.

packages/vue-vuetify/src/complex/ObjectRenderer.vue
```diff
@@ -10,13 +10,14 @@
       :cells="control.cells"
     />
     <additional-properties
-      v-if="hasAdditionalProperties && showAdditionalProperties"
+      v-if="hasAdditionalProperties || showAdditionalProperties"
```
sdirix, line 13: If `showAdditionalProperties === false` then we should still hide them, right?

packages/vue-vuetify/src/util/composition.ts
```diff
@@ -481,3 +484,18 @@ export const useIcons = () => {
     current: iconSet,
   };
 };
+
+export const determineClearValue = (
+  props: RendererProps<ControlElement & Scopable>,
```
sdirix, line 489: The props are not used

packages/vue-vuetify/src/controls/EnumControlRenderer.vue
```diff
@@ -56,8 +56,9 @@ const controlRenderer = defineComponent({
     ...rendererProps<ControlElement>(),
   },
   setup(props: RendererProps<ControlElement>) {
+    const clearValue = determineClearValue(props, '');
```
sdirix, line 59: We don't know whether `string` is even an allowed value in the `enums`

packages/vue-vuetify/src/extended/AutocompleteEnumControlRenderer.vue
```diff
@@ -81,9 +81,10 @@ const controlRenderer = defineComponent({
     ...rendererProps<ControlElement>(),
   },
   setup(props: RendererProps<ControlElement>) {
+    const clearValue = determineClearValue(props, '');
```
sdirix, line 84: same here, we don't know whether string is appropriate

packages/vue-vuetify/src/extended/AutocompleteOneOfEnumControlRenderer.vue
```diff
@@ -81,9 +81,10 @@ const controlRenderer = defineComponent({
     ...rendererProps<ControlElement>(),
   },
   setup(props: RendererProps<ControlElement>) {
+    const clearValue = determineClearValue(props, '');
```
sdirix, line 84: same here

packages/vue-vuetify/src/controls/AnyOfStringOrEnumControlRenderer.vue
```diff
@@ -62,9 +62,9 @@ const controlRenderer = defineComponent({
     ...rendererProps<ControlElement>(),
   },
   setup(props: RendererProps<ControlElement>) {
-    return useVuetifyControl(
-      useJsonFormsControl(props),
-      (value) => value || undefined,
+    const clearValue = determineClearValue(props, '');
```
sdirix, line 65: same here

sdirix (commented):
Hi @kchobantonov,

I had a first look at the implementation. The examples work great and are really impressive :heart: 

I added some comments, please have a look

packages/vue-vuetify/src/complex/AllOfRenderer.vue
kchobantonov: because if the size of the allOfRendererInfo changes then the key is not stable enough to guarantee that previous keys will point the exact same elements - e.g. we do not have a primary key for the info that we can use - usually this happens when we change the UI schema and the schema have to be applied again.

packages/vue-vuetify/src/complex/ArrayControlRenderer.vue
kchobantonov: same as above - to make sure that if the data size changes then the key will be unique enough so that it does not reuse UI elements that is not supposed to

packages/core/src/util/resolvers.ts
kchobantonov: that is the resolve the case where the value of the last path segment is equals to true. For example items: true and you want to bind that items then you need to resolve the jsonschema which can be boolean but in the code we ignore that this can be ever the case.

packages/vue-vuetify/src/util/composition.ts
kchobantonov: fixed

packages/vue-vuetify/src/extended/AutocompleteEnumControlRenderer.vue
kchobantonov: not sure at the moment how to handle this case indeed you are right that the string value nor the empty value in this case can be a valid enum value but then it will preserve the field at least in the case of the addtional properties which should be as of now the only option when preserving that field. For example for date and time we have similar thing where empty string is not exactly a date or time field and will require you to provide a valid value but at least won't remove the field althogther unless you remove that field using the delete button from the additional properties control. Open to any other suggestions here I guess.

packages/vue-vuetify/src/extended/AutocompleteOneOfEnumControlRenderer.vue
kchobantonov: same comment as the above enum

packages/core/src/mappers/renderer.ts
kchobantonov: fixed

packages/core/src/util/resolvers.ts
kchobantonov: fixed

kchobantonov:
@sdirix check my responses and also updated repo but it looks like the build failed perhaps because netlify is using greater than pnpm version 8 now ?

packages/vue-vuetify/src/complex/MixedRenderer.entry.ts
```diff
@@ -0,0 +1,74 @@
...
+  return false;
+};
+
+export const isMixedControl = (
+  uischema: UISchemaElement,
+  schema: JsonSchema,
+  context: TesterContext,
+) =>
+  isMixedSchema(uischema, schema, context) &&
+  (isControl(uischema) || isDefaultGenUiSchema(uischema));
```
sdirix, line 54: Why do we check `isDefaultGenUiSchema`? Note also that we have adopters who adopt the default UI Schema, so this might not apply to them.

sdirix (commented):
I have concerns about the proposed changes to the core functionality.

Please have a look at my new comments and the unresolved comments from my previous review.

packages/vue-vuetify/src/complex/AllOfRenderer.vue
kchobantonov: I agree as well but then we do need that on the core level - for now at least that is fixing the issue that I was having when I change the uischema from the monaco editor in our vue examples. So once this is available at the core - maybe some id for the uischema changes then we can utilize that.

packages/core/src/mappers/renderer.ts
```diff
@@ -871,7 +872,21 @@ export const mapStateToArrayControlProps = (
...
+        'array',
+        'boolean',
+        'integer',
+        'null',
+        'number',
+        'object',
+        'string',
+      ],
+    };
+  }
```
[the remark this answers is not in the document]
kchobantonov: ok good point - will see how to adjust that

packages/core/src/mappers/renderer.ts
```diff
@@ -871,7 +872,21 @@ export const mapStateToArrayControlProps = (
   const { path, schema, uischema, label, ...props } =
     mapStateToControlWithDetailProps(state, ownProps);
 
-  const resolvedSchema = Resolve.schema(schema, 'items', props.rootSchema);
+  let resolvedSchema = Resolve.schema(schema, 'items', props.rootSchema);
+  if ((resolvedSchema as any) === true) {
```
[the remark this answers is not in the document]
kchobantonov: yes that was again the case where the true value is not considered as a valid jsonschema by the metamodel used in jsonforms while in fact true/false is a valid jsonschema but to support that as typing means more code changes so that is why I didn't do that this way

packages/vue-vuetify/src/complex/MixedRenderer.vue
kchobantonov: the schema is not modified but we pass down a schema that is for the specific type that was selected by the user - e.g. when the user selects integer from a type that is [integer, object] then we adjust the type of the schema that we want to render to integer and pass along all associate properties for that - e.g. for example if there is a properties that is relevant only for type object then we remove that for integer - again the schema that is passed i a new schema based on the user selection so that we can delegate the rendering of that schema to the jsonforms

packages/vue-vuetify/src/complex/MixedRenderer.vue
kchobantonov: there is similar function in the core for that but return a schema while here I want the actual type not the schema and the function there is a method - maybe we could exposed that function into the jsonforms core and then reuse that for both cases perhaps

packages/vue-vuetify/src/complex/MixedRenderer.entry.ts
kchobantonov: it was needed because I wanted to make sure that besides the schema check we also check the uischema and the thing is that initially the code was
 `isMixedSchema(uischema, schema, context) && isControl(uischema)`
but then we weren't capturing the case where when the uischema is not provided by generated then the MixedRenderer was not involved - also it shouldn't affect others I guess since the mixed types are in general at the moment not really supported by the core / other renderers - I just didn't want the check to be just on the schema but to follow more what other testers were having as checks

packages/vue-vuetify/src/complex/ObjectRenderer.vue
kchobantonov: that is more like show additional properties even when additionalProperties is missing - this is because by jsonschema spec the default value is to allow additional properties while here in this implementation that needs to be explicitly set to true or an object - e.g. missing additionalProperties won't cause the object renderer to render the additional properties UI but if you specify that in the UI schema then it will

kchobantonov:
@sdirix please review again - here are the outstanding items that I need to check with you if you are ok with those as they are at the moment with the explanation that was provided.

1. change in the packages/core/src/util/resolvers.ts - I have added the reason why is that but I haven't yet resolved the conversation
2. the changes to the :key to include the size of the layout items is still there at least until there is some support from the core to know when the uischema is changes to use the proper key that will work on all changes to the layout data - e.g. reordering and etc.

packages/core/src/util/resolvers.ts
sdirix: Should we just switch the order of the checks? We could place the `isEmpty` check to after the pathSegments check. Then your case would work. There would only be a behavior change for empty schemas called to be resolved with an empty path. They would get the empty schema back instead of `undefined`.

packages/vue/src/components/JsonForms.vue
```diff
@@ -182,6 +182,7 @@ export default defineComponent({
       return {
         data: this.jsonforms.core.data,
         errors: this.jsonforms.core.errors,
+        additionalErrors: this.jsonforms.core.additionalErrors,
```
sdirix, line 185: Why emit `additionalErrors`? We don't emit them in the other renderer sets. As the user already hands them in they should be aware of them already.

packages/core/src/util/resolvers.ts
sdirix: @lucas-koehler what do you prefer? This special case for boolean schemas or changing the code from

```ts
  if (isEmpty(schema)) {
    return undefined;
  }

  // use typeof because schema can by of any type - check singleSegmentResolveSchema below
  if (typeof schema.$ref === 'string') {
    schema = resolveSchema(rootSchema, schema.$ref, rootSchema);
  }

  if (!pathSegments || pathSegments.length === 0) {
    return schema;
  }
```

to
```ts
  // use typeof because schema can by of any type - check singleSegmentResolveSchema below
  if (typeof schema?.$ref === 'string') { // additional ? required
    schema = resolveSchema(rootSchema, schema.$ref, rootSchema);
  }

  if (!pathSegments || pathSegments.length === 0) {
    return schema;
  }
  
  if (isEmpty(schema)) {
    return undefined;
  }
```

I quickly tested the change. All tests are successful and I don't see a regression in the test app

sdirix (commented):
Works good for me!

I'm waiting on Lucas' (or your) input to [this suggestion](https://github.com/eclipsesource/jsonforms/pull/2409/files#r2098368296) and the handling of the [additionalErrors](https://github.com/eclipsesource/jsonforms/pull/2409/files#r2098373585). When they are resolved I'm happy to merge.

Thanks for your great work and patience ;)

Nitpick: the "Mixed" and "Json Editor" examples in the Vue Vuetify app do not render as nicely as the other examples. Does not need to be resolved for the merge.

packages/core/src/util/resolvers.ts
lucas-koehler: I prefer changing the code like you suggested, @sdirix . It seems more stable to me - especially if there are similar cases that we did not think of, yet.

sdirix:
@kchobantonov Please adapt the resolver as suggested and the additionalErrors, then we can merge

kchobantonov:
@sdirix both changes that were request are now done - please review again and let me know if anything else is needed. Thanks for your feedback.

packages/core/src/util/resolvers.ts
```diff
@@ -123,19 +123,20 @@ const resolveSchemaWithSegments = (
...
   rootSchema: JsonSchema
 ): JsonSchema => {
-  if (isEmpty(schema)) {
-    return undefined;
-  }
-
   // use typeof because schema can by of any type - check singleSegmentResolveSchema below
-  if (typeof schema.$ref === 'string') {
+  if (typeof schema?.$ref === 'string') {
+    // additional ? required
```
sdirix, line 128:
```suggestion
```

packages/vue/src/types.ts
```diff
@@ -12,6 +12,9 @@ export interface InjectJsonFormsDispatch {
   dispatch: Dispatch<CoreActions>;
 }
 
-export type JsonFormsChangeEvent = Pick<JsonFormsCore, 'data' | 'errors'>;
+export type JsonFormsChangeEvent = Pick<
+  JsonFormsCore,
+  'data' | 'errors' | 'additionalErrors'
+>;
```
sdirix, line 18:
```suggestion
export type JsonFormsChangeEvent = Pick<JsonFormsCore, 'data' | 'errors'>;
```

kchobantonov:
@sdirix please review again - all concerns should be addresses and the last commit is fixing an issue where in the jsonschema example when you try to add new example field as an object and you do enable from the "Allow Additional Properties By Default" option then to correctly determine the type of the new properties that are going to be added to this new object example.

sdirix (approved):
Works for me! Thanks for the great contribution :heart:

PR #2413, opened 2025-01-31
core: Remove AJV usage from combinator mappers

### LINKED ISSUE #2371: Remove AJV usage from combinator mappers
Direct AJV usage in renderers (i.e. oneof renderer to determine applicable sub schema) should be removed completely as this is a hacky attempt to select the suitable subschema.
In general, we want to get rid of mandatory ajv usage in general.

For combinator renderers, AJV is invoked in `mapStateToCombinatorRendererProps`. This should be re-written to use a heuristic instead:
- If there is an type/kind/id (in this order) attribute: use to select initial entry. See if we can match this against a const in the schema. Otherwise try with first string and/or number property.
- id property should be configurable via a custom JSON Schema attribute (we already use a custom i18n attribute). The new attribute should start with x-jsf-. I.e. x-jsf-type-property
- allOf: Investigate whether calling mapStateToCombinatorRendererProps in mapStateToAllOfProps  is necessary

lucas-koehler:
- Adapt algorithm to determine the fitting schema index for combinators to no longer use AJV
- New heuristic uses identifying properties that should match a const value in the schema
- Adapt MaterialOneOfRenderer.test.tsx to fit new heuristic
- Describe changes and add examples to migration guide
- Adapt some of the anyOf and oneOf examples to custom and new identification properties

In contrast to oneOf and anyOf rendererer, allOf renderers do not need the `indexOfFittingSchema` because all schemas apply at once.
Thus, remove using mapStateToCombinatorRendererProps from mapStateToAllOfProps and, with this, the unnecessary calculation of the index.

fix #2371

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,117 @@
 # Migration guide
 
+## Migrating to JSON Forms 3.6
+
+### Combinator (anyOf & oneOf) index selection now uses a heuristic instead of AJV
+
+In this update, we have eliminated the direct usage of AJV to determine the selected subschema for combinator renderers.
+To achieve this, the algorithm in `getCombinatorIndexOfFittingSchema` and with this `mapStateToCombinatorRendererProps` was changed.
+Thus, custom renderers using either method might have behavior changes.
+This rework is part of an ongoing effort to remove mandatory usage of AJV from JSON Forms.
+
+Before this change, AJV was used to validate the current data against all schemas of the combinator.
+This was replaced by using a heuristic which tries to match the schema via an identification property
+against a `const` entry in the schema.
+
+The identification property is determined as follows in descending order of priority:
+
+1. The schema contains a new custom property `x-jsf-type-property` next to the combinator to define the identification property.
+2. The data has any of these properties: `type`, `kind`, `id`. They are considered in the listed order.
+3. The data has any string or number property. The first encountered one is used.
+
+If no combinator schema can be matched, fallback to the first one as before this update.
+
+Note that this approach can not determine a subschema for non-object subschemas (e.g. ones only defining a primitive property).
+Furthermore, subschemas can no longer automatically be selected based on validation results like
+produced by different required properties between subschemas.
+
+#### Example 1: Custom identification property
+
+Use custom property `x-jsf-type-property` to define which property's content identifies the subschema to select.
+In this case, `mytype` is defined as the property to use. The two subschemas in the `anyOf` each define a `const` value for this property.
+Meaning a data object with property `mytype: 'user'` results in the second subschema being selected.
+The `default` keyword can be used to tell JSON Forms to automatically initialize the property.
+
+```ts
+const schema = {
+  $schema: 'http://json-schema.org/draft-07/schema#',
+  type: 'object',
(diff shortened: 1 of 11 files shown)


MIGRATION.md
```diff
@@ -1,5 +1,117 @@
...
+
+Before this change, AJV was used to validate the current data against all schemas of the combinator.
+This was replaced by using a heuristic which tries to match the schema via an identification property
+against a `const` entry in the schema.
+
+The identification property is determined as follows in descending order of priority:
+
+1. The schema contains a new custom property `x-jsf-type-property` next to the combinator to define the identification property.
+2. The data has any of these properties: `type`, `kind`, `id`. They are considered in the listed order.
+3. The data has any string or number property. The first encountered one is used.
```
sdirix, line 20: We should remove this as this is kind of arbitrary. Potentially we could search for a `const` property with limitations on nested level etc. Could also be a followup as this is not mission critical.

lucas-koehler: I agree :+1: 
As discussed, we should also remove the `id` property from the second point because it seems likely that this is used for other purposes.

lucas-koehler:
@sdirix The tests are fixed and the review comments applied :)

MIGRATION.md
```diff
@@ -2,6 +2,116 @@
...
+The identification property is determined as follows in descending order of priority:
+
+1. The schema contains a new custom property `x-jsf-type-property` next to the combinator to define the identification property.
+2. At least one of the combinator schemas has this property with a const declaration: `type`, `kind`. They are considered in the listed order.
+
+If no combinator schema can be matched, fallback to the first one as before this update.
+
+Note that this approach can not determine a subschema for non-object subschemas (e.g. ones only defining a primitive property).
+Furthermore, subschemas can no longer automatically be selected based on validation results like
+produced by different required properties between subschemas.
```
sdirix, line 24:
```suggestion
Furthermore, subschemas can no longer automatically be selected based on validation results like produced by different required properties between subschemas.
```

packages/core/src/mappers/combinators.ts
```diff
@@ -36,6 +36,12 @@ export interface CombinatorSubSchemaRenderInfo {
 
 export type CombinatorKeyword = 'anyOf' | 'oneOf' | 'allOf';
 
+/** Custom schema keyword to define the property identifying different combinator schemas. */
+export const COMBINATOR_TYPE_PROPERTY = 'x-jsf-type-property';
```
sdirix, line 40: We could place these two properties as well  as `getCombinatorIndexOfFittingSchema` into a `Combinators` object. Then in all our code we should always refer to `Combinators.COMBINATOR_TYPE_PROPERTY` etc. This would allow adopters to easily globally override these two properties as well the actual determination function for easier customization.

MIGRATION.md
```diff
@@ -2,6 +2,116 @@
 
 ## Migrating to JSON Forms 3.6
 
+### Combinator (anyOf & oneOf) index selection now uses a heuristic instead of AJV
```
sdirix, line 5: I think we should place this for JSON Forms 4.0 as this will break adopters

PR #2414, opened 2025-01-31
create OneOfEnumCell for select inputs

edkahara:
Context: https://jsonforms.discourse.group/t/oneof-not-producing-select-field-or-multiple-choice-checklist-in-vanilla-renderers/2574

1. Create and export `OneOfEnumCell` cell that creates select inputs in vanella-renderers.
2. Adds `hideEmptyOption` that when `true`, removes the empty option so the select input only has the options the user passes in `oneOf`


lucas-koehler:
Hi @edkahara , thank you for the contribution :heart: 

Please note that you need to sign the contributor license agreement (CLA) before we can accept your contribution. See this comment: https://github.com/eclipsesource/jsonforms/pull/2414#issuecomment-2628557966

lucas-koehler (approved):
Hi @edkahara thanks again for the contribution and adding these extensive tests :heart: 

The code now looks good to me!
Building the example app fails here. However, this is not caused by this PR and building and testing it locally worked fine. Thus, we can still merge this :)

PR #2420, opened 2025-02-17
Default JSON schema values for visibility rules

### LINKED ISSUE #2419: Visibility rules should use default values in the JSON schema when being evaluated
### Describe the bug

Default values in the JSON schema aren't used when evaluating visibility rules when the AJV [useDefaults](https://ajv.js.org/options.html#usedefaults) option is enabled.

### Expected behavior

If the schema has a default value and the data object doesn't contain a value for the scope referenced in the rule, the JSON schema's default value should be used.

I was able to reproduce it with a test in the core package.

```diff
diff --git a/packages/core/test/mappers/renderer.test.ts b/packages/core/test/mappers/renderer.test.ts
index 9d0dd516..80860fe4 100644
--- a/packages/core/test/mappers/renderer.test.ts
+++ b/packages/core/test/mappers/renderer.test.ts
@@ -85,6 +85,17 @@ const mockDispatch = (
   return [getCore, dispatch];
 };
 
+const showRule = {
+  effect: RuleEffect.SHOW,
+  condition: {
+    type: 'SCHEMA',
+    scope: '#/properties/firstName',
+    schema: {
+      const: 'Homer',
+    },
+  },
+};
+
 const hideRule = {
   effect: RuleEffect.HIDE,
   condition: {
@@ -1210,6 +1221,38 @@ test('mapStateToLayoutProps - hidden via state with path from ownProps ', (t) =>
   t.false(props.visible);
 });
 
+test('mapStateToLayoutProps - show via state from default values from schema ', (t) => {
+  const uischema = {
+    type: 'VerticalLayout',
+    elements: [coreUISchema],
+    rule: showRule,
+  };
+  const ownProps = {
+    uischema,
+  };
+  const state = {
+    jsonforms: {
+      core: {
+        ajv: createAjv({
+          useDefaults: true,
+        }),
+        schema: {
+          type: 'object',
+          properties: {
+            firstName: { type: 'string', default: 'Homer' },
+            lastName: { type: 'string' },
+          },
+        },
+        data: { lastName: 'Simpson' },
+        uischema,
+        errors: [] as ErrorObject[],
+      },
+    },
+  };
+  const props = mapStateToLayoutProps(state, ownProps);
+  t.true(props.visible);
+});
+
 test("mapStateToOneOfProps - indexOfFittingSchema should not select schema if enum doesn't match", (t) => {
   const uischema: ControlElement = {
     type: 'Control',

```

### Screenshots

_No response_

v3.5.1

### Package

Core

### Additional context

_No response_

danielronnkvist:
Uses the AJV from state to [assign default values](https://ajv.js.org/guide/modifying-data.html#assigning-defaults) if it has been configured to do so for the JSON schema. Included tests to ensure that the `SHOW` rule uses the default value when it's not present in the data, and that if the `useDefaults` option isn't set the visibility rule should return false.

Fixes #2419

### DIFF
--- a/packages/core/src/mappers/renderer.ts
+++ b/packages/core/src/mappers/renderer.ts
@@ -1038,9 +1038,12 @@ export const mapStateToLayoutProps = (
   state: JsonFormsState,
   ownProps: OwnPropsOfLayout
 ): LayoutProps => {
-  const rootData = getData(state);
+  const originalData = getData(state);
+  // Create a new object to avoid modifying the original data
+  const rootData = { ...originalData };
   const ajv = getAjv(state);
-  ajv?.compile(getSchema(state))(rootData);
+  const schema = getSchema(state);
+  ajv?.compile(schema)(rootData);
   const { uischema } = ownProps;
   const visible: boolean =
     ownProps.visible === undefined || hasShowRule(uischema)
--- a/packages/core/src/mappers/renderer.ts
+++ b/packages/core/src/mappers/renderer.ts
@@ -1039,10 +1039,12 @@ export const mapStateToLayoutProps = (
   ownProps: OwnPropsOfLayout
 ): LayoutProps => {
   const rootData = getData(state);
+  const ajv = getAjv(state);
+  ajv?.compile(getSchema(state))(rootData);
   const { uischema } = ownProps;
   const visible: boolean =
     ownProps.visible === undefined || hasShowRule(uischema)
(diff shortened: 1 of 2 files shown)


sdirix (changes requested):
The PR can't be merged as is as this would lead to massive performance decline. Please also see [this message](https://github.com/eclipsesource/jsonforms/issues/2419#issuecomment-2665168312). I'm not convinced that this fixes a case that could happen during "real-world" usage. The unit test fails because the constructed JSON Forms state in the test is not correct.

PR #2421, opened 2025-02-19
Fix MIT license

SamuelMarks:
You currently show:
image

What you want:
image


sdirix (changes requested):

Please revert all changes. The only change we *really* need is to remove the https://github.com/eclipsesource/jsonforms line from the License. Then GH will detect the `MIT` license.

- We don't need a `LICENSE-MIT` file as we have a single license
- We don't want to maintain an updated year in the license files and headers as this is not a legal requirement and produces unneeded churn.

sdirix:
I created a follow up with the suggested changes (https://github.com/eclipsesource/jsonforms/pull/2431). Thanks for the contribution :heart:

PR #2431, opened 2025-03-24
chore: update LICENSE text

sdirix:
Updates the LICENSE text to
- refer to EclipseSource instead of EclipseSource Munich
- remove the link to the repository, enabling GH auto detection


lucas-koehler (commented):
The change LGTM. Should we also adapt the license headers in the code files?

sdirix:
Good idea, let's handle that in a follow up at some point (low prio)

PR #2432, opened 2025-03-24
[ENG-1114] Add support for preserving data on hide

nhcathcart:
This adds a flag to preserve data when the field is hidden. This will be necessary for when we want to store data in hidden fields. The default clearing behavior will remain, but when this flag is present we will not clear data. There will be a follow up to add a toggle in the form builder to set this flag.


sdirix:
@nhcathcart this looks like a PR which should have been opened against a [fork of JSON Forms](https://github.com/mosaic-avantos/jsonforms) ;)

PR #2433, opened 2025-04-04
fix(material-renderers): add missing Mui-error class for some controls

vishnutsivan:
- Passing error prop for missing controls

### DIFF
--- a/packages/material-renderers/package.json
+++ b/packages/material-renderers/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/material-renderers",
-  "version": "3.5.1",
+  "version": "3.5.2",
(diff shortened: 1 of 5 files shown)


lucas-koehler (changes requested):
Thanks for this contribution aligning this with the other controls :heart: 

Please remove the version update as explained inline :)

lucas-koehler (approved):
@vishnutsivan Thanks for the update! This LGTM now.

The next regular release will probably be around end of May or in June. There is no fixed date, yet. However, we could do a pre-release (e.g. 3.6.0-alpha.0) this week.

lucas-koehler:
@IceFreez3r I published version `3.6.0-alpha.0`

PR #2440, opened 2025-04-15
refactor(material-renderers): flushable debounce

### LINKED ISSUE #2260: Unpredictable data change
### Describe the bug

When I quickly click a button, sometimes the previous state is logged to the console. I can never be sure that I'm accessing actual form data (or at least I don't know how to get to them since there is no real reference to them or to the form)

```tsx
const initialData = {
  name: 'Max Power',
};

const ClearFormExample = () => {
  const [data, setData] = useState(initialData);

  return (
    <div>
      <JsonForms
        schema={schema}
        uischema={uischema}
        data={data}
        renderers={materialRenderers}
        cells={materialCells}
        onChange={({ data, _errors }) => setData(data)}
      />
      <Button onClick={() => console.log(data)} color='primary'>
        Print data
      </Button>
    </div>
  );
};
```

### Expected behavior

It would be nice to be able to directly access the internal state of the form by reference

1. Just try the modified example from the docs

### Screenshots

_No response_

v3.1.0

### Framework

React

### RendererSet

_No response_

### Additional context

_No response_

johannesss:
this is just a poc to see if this path is viable

fixes #2260

### DIFF
--- a/packages/material-renderers/src/util/debounce.ts
+++ b/packages/material-renderers/src/util/debounce.ts
@@ -1,19 +1,19 @@
 /*
   The MIT License
-  
+
   Copyright (c) 2021 EclipseSource Munich
   https://github.com/eclipsesource/jsonforms
-  
+
   Permission is hereby granted, free of charge, to any person obtaining a copy
   of this software and associated documentation files (the "Software"), to deal
   in the Software without restriction, including without limitation the rights
   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
   copies of the Software, and to permit persons to whom the Software is
   furnished to do so, subject to the following conditions:
-  
+
   The above copyright notice and this permission notice shall be included in
   all copies or substantial portions of the Software.
-  
+
   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


johannesss:
> Conceptually this makes sense to me. In which state is the PR now, shall we do a detailed review before you continue?

Yes, sure!

packages/material-renderers/src/util/debounce.ts
```diff
@@ -25,15 +25,27 @@
...
+  handleChange: (path: string, value: any) => void;
+  data: any;
+  path: string;
+  eventToValue?: (ev: any) => any;
+  defaultValue?: any;
+  flushOnBlur?: boolean;
+  focused?: boolean;
+  timeout?: number;
+}
+export const useDebouncedChange = ({
```
sdirix, line 39: I know it's a lot of parameters already, but I would prefer to keep the old function and just add more optional parameters to it. The reason is that this change needlessly breaks backward compatibility for adopters.

sdirix (changes requested):
Works for me! I just have a minor change request

packages/material-renderers/src/util/debounce.ts
johannesss: Sure thing. I'll fix this asap!

sdirix (approved):
Works for me! Thanks for the contribution :heart:

PR #2441, opened 2025-04-23
feat(core): Add validate function based rule condition

lucas-koehler:
Add a new rule condition `ValidateFunctionCondition` using a given `validate` function to evaluate the condition result. This allows using arbitrary custom logic to evaluate condition results.
This facilitates not using schema-conditions to be able to only use one pre-compiled AJV for the data schema at a later stage.

### DIFF
--- a/packages/core/src/models/uischema.ts
+++ b/packages/core/src/models/uischema.ts
@@ -150,6 +150,16 @@ export interface SchemaBasedCondition extends BaseCondition, Scoped {
   failWhenUndefined?: boolean;
 }
 
+/** A condition using a validation function to determine its fulfillment. */
+export interface ValidateFunctionCondition extends BaseCondition, Scoped {
+  /**
+   * Validates whether the condition is fulfilled.
+   *
+   * @param data The data as resolved via the scope.
+   * @returns `true` if the condition is fulfilled */
+  validate: (data: unknown) => boolean;
+}
+
 /**
  * A composable condition.
  */
@@ -179,7 +189,8 @@ export type Condition =
   | LeafCondition
   | OrCondition
   | AndCondition
-  | SchemaBasedCondition;
+  | SchemaBasedCondition
+  | ValidateFunctionCondition;
 
 /**
  * Common base interface for any UI schema element.
--- a/packages/core/src/util/runtime.ts
+++ b/packages/core/src/util/runtime.ts
@@ -33,6 +33,7 @@ import {
   SchemaBasedCondition,
   Scopable,
   UISchemaElement,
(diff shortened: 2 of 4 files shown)


packages/core/src/models/uischema.ts
```diff
@@ -150,6 +150,16 @@ export interface SchemaBasedCondition extends BaseCondition, Scoped {
...
 }
 
+/** A condition using a validation function to determine its fulfillment. */
+export interface ValidateFunctionCondition extends BaseCondition, Scoped {
+  /**
+   * Validates whether the condition is fulfilled.
+   *
+   * @param data The data as resolved via the scope.
+   * @returns `true` if the condition is fulfilled */
+  validate: (data: unknown) => boolean;
```
sdirix, line 160: I think we should hand in much more data here, i.e. whatever we have available at call time.

For example
```suggestion
  validate: (rule: {
    scopedData: unknown;
    schema: JSONSchema,
    uischema: UISchemaElement,
    path: string; // or string[]
    // etc.
    }) => boolean;
```

lucas-koehler: Generally, I agree with your sentiment. However, we do not have that much available without breaking the public API:
The `evaluateCondition` method has these parameters:
```ts
data: any, // This is the full data
condition: Condition,
path: string,
ajv: Ajv
```

Tracing upwards in the call history to the first publicly exposed methods using this - `evalVisibility` and `evalEnablement` - they have these parameters available (the same is true for `isVisible` and `isEnabled`):

```
uischema: UISchemaElement, // ui schema element containing the rule
data: any, // full data
path: string = undefined, // Optional instance path for nested controls
ajv: Ajv
```

Thus, without breaking the API of there we could additionally hand in:
- the UISchemaElement containing the condition
- the full data
- the optional instance path for nested controls

I omitted handing in the full data because this can be covered by using the root scope `#`. Thinking about it again, this is then only the nested root data.
Thus, I suggest handing in the three listed properties in addition to the scoped data. I'd like to avoid breaking the API. WDYT?

lucas-koehler:
@sdirix I now added the additional context information being available without breaking API as described here: https://github.com/eclipsesource/jsonforms/pull/2441#discussion_r2056168335

packages/core/src/models/uischema.ts
sdirix: Makes sense to me :+1:

sdirix (changes requested):
Works for me, but there is an issue with the example

lucas-koehler:
@sdirix Thanks for the review and suggestion :) The example should be fixed now.

PR #2443, opened 2025-04-24
draft: allow control-wrapper to be overridden by provided component

yaffol:
This draft PR shows an initial method of allowing the control-wrappper component to be overridden, by inject(...)ing the wrapper component from the application that consumes JSON Forms.

My use case for this is that I have a custom rendering layout which moves the field label to the left of the field and I also show a customised information box in some cases above the field.

This allows me to use the stock renderers but perform this customisation, so I don't have to update custom renderers for every field type where I only want to make relatively minor wrapper layout changes.

### DIFF
--- a/packages/vue-vuetify/src/controls/DateControlRenderer.vue
+++ b/packages/vue-vuetify/src/controls/DateControlRenderer.vue
@@ -1,6 +1,8 @@
 <template>
-  <control-wrapper
+  <component
+    :is="resolvedWrapper"
     v-bind="controlWrapper"
+    :control="control"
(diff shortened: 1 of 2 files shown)


sdirix (commented):
In general this change makes a lot of sense to me. I think it would be nicer if the `control-wrapper` was already returned by `useVuetifyControl`, so we don't need a separate injection.

I didn't check: If we have many `useVuetifyControl` which don't need the wrapper, then I'm fine with a separate injection. However I would prefer to use `useControlWrapper` instead of `controlWrapperOverride`.

PR #2446, opened 2025-05-11
allow using lazy loadable renderers

dsl400:
proposal to add possibility to use lazy loadable renderers

### DIFF
--- a/packages/angular/src/library/jsonforms.component.ts
+++ b/packages/angular/src/library/jsonforms.component.ts
@@ -24,6 +24,7 @@
 */
 import maxBy from 'lodash/maxBy';
 import {
+  ComponentFactory,
   ComponentFactoryResolver,
   Directive,
   Input,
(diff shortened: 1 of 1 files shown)


dsl400:
I was hoping to migrate our existing code by loading some components as custom renderers

sdirix:
I don't think this PR is necessary. Instead of within JSON Forms you could await the renderers outside of JSON Forms and only render JSON Forms once they are resolved.

dsl400:
> I don't think this PR is necessary. Instead of within JSON Forms you could await the renderers outside of JSON Forms and only render JSON Forms once they are resolved.

Can you provide some technical arguments about why ? 
Maybe it is something that I do not yet understand.

sdirix:
Hi @dsl400, I’m not sure this needs to be built into JSON Forms. Since you're writing custom renderers anyway, you can wrap them to handle async loading yourself. This keeps the framework cleaner and avoids adding complexity.

sdirix:
I discussed this again with @lucas-koehler  and we came to the conclusion that this feature could be useful for multiple adopters, so we would accept a polished contribution.
Let’s just keep the tone respectful so we can focus on the technical discussion.

packages/angular/src/library/jsonforms.component.ts
```diff
@@ -126,11 +127,32 @@ export class JsonFormsOutlet
...
+        });
+        return;
+      } else {
+        bestComponent = renderer.renderer;
+      }
     }
 
     const componentFactory =
       this.componentFactoryResolver.resolveComponentFactory(bestComponent);
+    this.renderComponent(componentFactory, uischema, schema, props);
```
sdirix, line 147: For async components, this will lead to the `UnknownRenderer` being rendered first which is misleading and might look like an error to the user. Instead we should render a `LoadingRenderer` or something similar.

packages/angular/src/library/jsonforms.component.ts
```diff
@@ -126,11 +127,32 @@ export class JsonFormsOutlet
       renderer !== undefined &&
       renderer.tester(uischema, schema, testerContext) !== -1
     ) {
-      bestComponent = renderer.renderer;
+      if (renderer.renderer instanceof Promise) {
+        renderer.renderer.then((resolvedRenderer) => {
```
sdirix, line 131: There can be a race condition here:
1. The outlet is invoked, however the renderer (Renderer A) is a promise and needs some time to complete
2. There is an update in the mean time, another renderer (Renderer B) is required and it is awaited too
3. The Renderer B resolves first and renders
4. Now finally Renderer A resolves and will be used for rendering, although Renderer B is the one which should be used.

sdirix (changes requested):

dsl400:
@sdirix  Thank you very much for reconsidering this PR, I will come back with updates

PR #2448, opened 2025-05-11
repository configuration for angular packages dev mode

dsl400:
proposal of angular packages configuration  that uses `ng serve`  for development mode

when running `npm run dev` in `packages/angular-material`  the packages `core`  `angular`  `angular-material` and `examples`  are referenced from source code and changes are applied immediately


packages/angular-material/package.json
```diff
@@ -134,7 +134,7 @@
     "ts-loader": "^9.5.1",
     "tslib": "^2.5.0",
     "typedoc": "~0.25.3",
-    "typescript": "~5.5.0",
+    "typescript": "~5.4.2",
```
lucas-koehler, line 137: I don't think reducing the typescript version is necessary here. I tested it locally without this and it still worked. Please revert this change. Then, re-generating the pnpm lockfile should also not be necessary.

lucas-koehler (changes requested):
Hi @dsl400 .
thanks for the contribution :heart: !
Testing this locally, changes in the angular and core bundles were immediately recognized. This is a great improvement for development. Thanks!

I have one inline remark about the change of the Typescript version as I don't think it's required. Please have a look.

packages/angular-material/package.json
dsl400: Will do that. I tested a few methods for this solution and forgot to undo this modification

lucas-koehler (approved):
@dsl400 Thanks for the update. I also reverted the larger lock file change that stemmed from the typescript dependency change. With this, it LGTM :)
Thanks again for the contribution!

PR #2454, opened 2025-05-30
fix: use translated description in Angular Material

sdirix:
The control description is already translated via JSON Forms core. The potentially translated description is now used instead of manually resolving the description from the schema.

Adds testcases.


lucas-koehler (approved):
LGTM, thanks!

PR #2455, opened 2025-05-30
fix: append seconds in Vue Vanilla time

sdirix:
Appends seconds if necessary in Vue Vanilla time renderer. This makes sure that we do not produce validation errors.

Also allows to configure the "step" attribute via UI Schema options.

Adds test cases.


lucas-koehler (approved):
LGTM! Thanks for the fix :)

PR #2456, opened 2025-06-01
Validation Icon on ExpandPanel and ListWithDetail

vishnutsivan:
Validation icons on List with details item
image

Validation icons on Expand panel item
image


packages/core/src/mappers/renderer.ts
```diff
@@ -823,6 +833,7 @@ export interface OwnPropsOfMasterListItem {
 
 export interface StatePropsOfMasterItem extends OwnPropsOfMasterListItem {
   childLabel: string;
+  childErrors: string;
```
lucas-koehler, line 836: This should be the raw errors of type `ErrorObject[]` instead of the processed string for two reasons:
1. To be consistent with the behavior of the array mapper in `mapStateToArrayControlProps`
2. To allow renderers to decided for themselves how to show the errors.

PR #2468, opened 2025-07-18
deps: Update to Vuetify ^3.9 and Vue ^3.5

lucas-koehler:
- Update to latest Vuetify 3.9.0
- Update to latest Vue 3.5.17
- Update Vue peer dependency to ^3.5.0
- Fix imports for Vuetify promoted from labs
- Fix typings for update events in Vuetify DateControlRenderer
- Regenerate Vuetify test snapshots to fit changes in Vuetify (e.g. setting values in HTML element)
- Add visualViewport mock for Vuetify tests as this seems to be needed after dependency update
- In JsonForms.vue explicitly type provided state as `JsonFormsSubStates` to avoid TS7056

### DIFF
--- a/packages/vue-vuetify/src/controls/DateTimeControlRenderer.vue
+++ b/packages/vue-vuetify/src/controls/DateTimeControlRenderer.vue
@@ -88,6 +88,19 @@
                             proxyModel.value.time = val;
                           } else {
                             pickerValue.time = val;
+                          }
+                        }
+                      "
+                      @update:minute="
+                        () => {
+                          if (!showActions && !useSeconds) {
+                            showMenu = false;
+                          }
+                        }
+                      "
+                      @update:second="
+                        () => {
+                          if (!showActions && useSeconds) {
                             showMenu = false;
                           }
                         }
@@ -145,7 +158,19 @@
                               date: pickerValue.date,
                               time: val,
                             };
-
+                          }
+                        }
+                      "
+                      @update:minute="
+                        () => {
+                          if (!showActions && !useSeconds) {
+                            showMenu = false;
+                          }
+                        }
+                      "
+                      @update:second="
+                        () => {
+                          if (!showActions && useSeconds) {
                             showMenu = false;
                           }
                         }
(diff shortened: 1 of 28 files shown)


packages/vue-vuetify/src/controls/IntegerControlRenderer.vue
```diff
@@ -36,7 +36,7 @@ import {
   type RendererProps,
 } from '@jsonforms/vue';
 import { defineComponent } from 'vue';
-import { VNumberInput } from 'vuetify/labs/VNumberInput';
+import { VNumberInput } from 'vuetify/components';
```
sdirix, line 39: Something must have changed. The `#numbers` example of the Vue Vuetify demo app no longer works properly

Image

packages/vue-vuetify/src/controls/DateTimeControlRenderer.vue
```diff
@@ -197,10 +197,10 @@ import {
   VTab,
   VTabs,
   VTextField,
+  VTimePicker,
```
sdirix, line 225: The timer breaks, I can no longer select minutes

Image

See `#dates` example of Vue Vuetify

sdirix (changes requested):
I can successfully build and use the Vue Vuetify demo application. However it breaks for me for the components which were moved from lab to the regular lib. Likely they changed behavior a bit.

lucas-koehler:
Hi @sdirix , thanks for the review and valuable feedback. I adapted the following renderers:
- TimeControlRenderer: Close selector after the smallest shown unit (i.e. minutes or seconds) is chosen. This should reflect the previous behavior
- DateTimeControlRenderer: Close selector after the smallest shown unit (i.e. minutes or seconds) is chosen. This differs from the previous behavior in regard to day selection. Previously, the selector also auto-closed after selecting a day. This is no longer the case. The reason being that there is no event to listen to changes to the day value. Thus, this would require more complicated logic. However, I think the new behavior is fine as it follows the same logic as the pure date or time controls: It closes after selecting the smallest unit.
- NumberControlRenderer: Set the precision in addition to the step: With this, decimals are correctly shown. The behavior should now be the same as before the Vuetify upgrade
  - I also added a step with scientific notation to the numbers example to be able to see the behavior for this in the example

sdirix (approved):
All reported regressions are fixed 👍 Thanks!

PR #2472, opened 2025-07-24
fix: root $ref resolving

### LINKED ISSUE #2471: Resolver crashes when encountering a `$ref` at root level
### Describe the bug

When a JSON Schema specifies a `$ref` at the root level of the schema, our resolver often ends up in an endless recursion until a maximum stack error is thrown.

Example JSON Schema:

```json
{
 "definitions": {
   "foo": {
     "type": "string"
   }
 },
 "$ref": "#/definitions/foo"
}
```

The issue is caused by [this check](https://github.com/eclipsesource/jsonforms/blob/883f108817ff7ba3bdd2976269f0a109c6f37659/packages/core/src/util/resolvers.ts#L127-L129).

### Expected behavior

The resolver should not end up in an endless recursive "loop" when the actual schema is loop-free.

1. Use JSON Forms to render the schema above
2. Observe console logs

### Screenshots

_No response_

3.6.0

### Package

Core

### Additional context

Related discussion: https://jsonforms.discourse.group/t/recursive-schema-error/2746

sdirix:
Reworks schema resolving to be able to handle $refs in the root of the JSON Schema. Also adds circle reference detection, preventing runtime crashes.

Fixes #2471

Also adjusts core ava settings to fix incorrect source mappings.


lucas-koehler (approved):
LGTM, thanks for this extensive improvement of the resolving algorithm :)

PR #2478, opened 2025-08-12
preserve the form data so when we update the additionalErrors the dat…

kchobantonov:
…a is not reset to the initial state


kchobantonov:
@sdirix  please review, to check the issue in the current master branch just apply the change for ExampleView.vue and then try to test the additional-errors example, e.g. change the form once it is loaded and then click on the "Add Additonal Error" button and observe that the form data is reset, then check the  https://deploy-preview-2478--jsonforms-examples.netlify.app/vue-vuetify/#additional-errors

sdirix:
So the error occurs only if the user of JSON Forms:
 - hands over **new** `additionalErrors` to JSON Forms component
 - does **not** hand over an updated state of the the `data`

I don't think that this is a valid use case. Either the props of the JSON Forms component are "frozen" and the management is done by JSON Forms, or the props are always live. The error case is a weird mix which is not really valid.

The same issue will occur in Angular and React if only parts of the props are updated. Either they should never be updated by the user, i.e. uncontrolled variant, or they should all be updated, i.e. controlled variant.

packages/vue/src/components/JsonForms.vue
```diff
@@ -189,29 +189,23 @@ export default defineComponent({
...
-          this.schemaToUse,
-          undefined,
-          undefined,
-          this.schemaToUse
-        );
+      this.uischemaToUse = newUischema ?? generateUISchema(this.schemaToUse);
     },
     data(newData) {
-      this.dataToUse = newData;
+      this.dataToUse = newData === undefined ? {} : newData;
```
lucas-koehler: Why is this necessary? For instance, we don't do this in React

packages/vue/src/components/JsonForms.vue
```diff
@@ -26,21 +26,24 @@ import {
...
-
 const EMPTY: ErrorObject[] = reactive([]);
 
+const createDefaultLayout = (): Layout => ({
+  type: 'VerticalLayout',
+  elements: [],
+});
+const generateUISchema = (schema: JsonSchema) =>
+  Generate.uiSchema(schema, undefined, undefined, schema) ??
+  createDefaultLayout();
```
lucas-koehler: This is behavior that shouldn't be defined specifically here. If the Generator does not handle this already, we might fix it there for all.

packages/vue-vuetify/dev/views/ExampleView.vue
```diff
@@ -91,7 +91,6 @@ const onChange = (event: JsonFormsChangeEvent): void => {
       monaco.Uri.parse(toDataUri(props.example.name)),
       event.data !== undefined ? JSON.stringify(event.data, null, 2) : '',
     );
-    state.data = event.data;
```
lucas-koehler, line 97: I get that this is introduced to test the new behavior but even if we were to merge it, I don't think the example app use this unrecommended mixed approach. Thus, this should be reverted

lucas-koehler (changes requested):
Hi @kchobantonov ,
Thanks for the updates! I have some comments inline unrelated to the setting of `this.dataToUse = newEvent.data;`.

I am still a bit concerned about this:
- It makes the data propagation less intuitive:
  - In the controlled case, the data is set to dataToUse again after it has already been updated. Granted, this does not trigger an additional invocation of `coreDataToUpdate()`.
  - In the uncontrolled case, this triggers `coreDataToUpdate` an additional time without it being needed. This has a possible performance penalty because it invokes the middleware again
  - In the mixed case, it triggers the `coreDataToUpdate` like in the uncontrolled case. From a data point of view that should also not be necessary because the data is uncontrolled. Granted, this is acceptable considering the issues that this tries to fix.

As this does make the state handling less intuitive and lowers performance in uncontrolled mode as far as I can tell, I am leaning towards not introducing supporting this mixed case.

Do you have a concrete use case in mind where this is necessary? For a diffuse robustness increase, I prefer to not introduce this because consumers can always control the data in the component rendering the form. Even if the state handling is more complex for other properties.
We can of course add this to the documentation though.

packages/vue/src/components/JsonForms.vue
kchobantonov: since we do generate the json schema based on the data when the jsonschema is not provided then if the data is undefined then the whole jsonforms will throw an error so this update of the data follows what we have initially on the data property initialization. If jsonforms works works without errors with undefined data when not providing the jsonschema then that code can be removed

packages/vue/src/components/JsonForms.vue
kchobantonov: yes since jsonforms expects not undefined ui schema this code works to handle the cases where the uiSchema returns null - if we do not wrap that then jsonforms throws error since it does expect valid uiSchema. I haven't modified the uiSchema generation since that might break some expectation - just wasn't sure if that is the case but yes if you think that modifying the generateUiSchema to never return null/undefined or change the jsonforms to handle null/undefined uiSchema then that code can be removed

packages/vue-vuetify/dev/views/ExampleView.vue
kchobantonov: well it uses that in the sense that when you try to apply any changes on the ui schema then you do not want to change the data back to the initial example data but to preserve the modified data if that was the case

kchobantonov:
@lucas-koehler thanks for the review - please check my comments

lucas-koehler:
@sdirix Thanks for the explanation. I agree with your sentiment. Both suggested solutions are fine with me but I prefer the second one to keep the API of the JSON Forms component concise. I.e.:

> Alternatively we could also not create a new modelValue prop but just offer an update:data event for custom v-model:data support. In fact that's even better to avoid introducing a redundant variable.

packages/core/src/generators/schema.ts
```diff
@@ -177,5 +177,5 @@ export const generateJsonSchema = (
 
   const gen = new Gen(findOption);
 
-  return gen.schemaObject(instance);
+  return gen.property(instance);
```
sdirix, line 176: No tests cover the new behavior. All existing tests in [`schema.test.ts`](https://github.com/eclipsesource/jsonforms/blob/3a9bc7449e96414026b9e7f2572b4d0d54000a89/packages/core/test/generators/schema.test.ts) pass objects. Tests for `generateJsonSchema("hello")`, `generateJsonSchema(42)`, `generateJsonSchema([1,2])` etc. are needed.

packages/vue/src/components/JsonForms.vue
```diff
@@ -189,29 +189,23 @@ export default defineComponent({
...
-          undefined,
-          this.schemaToUse
-        );
+      this.uischemaToUse = newUischema ?? generateUISchema(this.schemaToUse);
     },
     data(newData) {
-      this.dataToUse = newData;
+      this.dataToUse = newData === undefined ? {} : newData;
+
+      if (!this.schema) {
```
sdirix: Combined with the other change sof the PR, this means that on every user edit (in schema-less mode) the schema and uischema are regenerated from the core's data, which is expensive and likely unnecessary since the shape of the data typically doesn't change between edits the user can make through JSON Forms

We need to differentiate between outside and user based changes and only regenerate on outside changes.

sdirix (commented):
Yes, I think we need to cleanly differentiate between "controlled" data and "uncontrolled" data. This becomes even more clear now with the schema regeneration.

Alternative for the schema generation: We could also think about NOT handling the schema regeneration within JSON Forms and expecting the caller of JSON Forms to remount us, e.g. via a key, so we don't need to handle that case within JSON Forms. However I would be fine to keep it within JSON Forms but only for the "controlled" mode, so that we don't regenerate on simple user edits.

packages/core/src/generators/schema.ts
kchobantonov: fixed in the latest commit

packages/vue/src/components/JsonForms.vue
kchobantonov: resolved with the latest commit

packages/vue/src/components/JsonForms.vue
sdirix: I agree with @lucas-koehler that this fallback belongs into the generator, so React and Angular profit as well: `Generate.uiSchema` already declares a non-null return type. Wrapping only the final result of `generateDefaultUISchema` keeps the per-property `null` semantics intact and lets us drop this local fallback.

kchobantonov:
> Yes, I think we need to cleanly differentiate between "controlled" data and "uncontrolled" data. This becomes even more clear now with the schema regeneration.
> 
> Alternative for the schema generation: We could also think about NOT handling the schema regeneration within JSON Forms and expecting the caller of JSON Forms to remount us, e.g. via a key, so we don't need to handle that case within JSON Forms. However I would be fine to keep it within JSON Forms but only for the "controlled" mode, so that we don't regenerate on simple user edits.



> Yes, I think we need to cleanly differentiate between "controlled" data and "uncontrolled" data. This becomes even more clear now with the schema regeneration.
> 
> Alternative for the schema generation: We could also think about NOT handling the schema regeneration within JSON Forms and expecting the caller of JSON Forms to remount us, e.g. via a key, so we don't need to handle that case within JSON Forms. However I would be fine to keep it within JSON Forms but only for the "controlled" mode, so that we don't regenerate on simple user edits.

There are now checks for schema-less usage: when the data prop changes, we compare the newly generated schema with the current one and only update the schema/UI schema if the data structure actually changed.

kchobantonov:
@sdirix Do we still plan to merge this PR, or have we decided not to pursue this approach?

I'm asking because, besides the additionalErrors issue with form data, this PR also includes a fix for schema generation. Instead of:

```return gen.schemaObject(instance);```

it uses:

```return gen.property(instance);```

which is more correct, as it allows the generator to properly detect whether the input is a simple type, an array, or an object and generate the appropriate schema.

If we're no longer considering merging this PR, I can create a separate PR containing just the schema generation fix and then close this one.

**Comment by sdirix** [comment]
I will re-review this PR today. Let's see.

File: packages/vue/src/components/JsonForms.vue [hunk 6]
```diff
@@ -251,6 +255,9 @@ export default defineComponent({
       );
     },
     eventToEmit(newEvent) {
+      // update the data so if we change the additionalErrors this won't reset the form data
+      this.dataToUse = newEvent.data;
```
sdirix: This sync silently changes the `data` prop semantics and diverges from React and Angular, where the prop stays authoritative (https://github.com/eclipsesource/jsonforms/blob/6db5bc87e763922c6853d76b166885978e280c74/packages/react/src/JsonFormsContext.tsx#L159-L173). It also dispatches a redundant `updateCore` through the middleware after every user edit in the uncontrolled case. Remove it: with an `update:data` event, `v-model:data` keeps the prop current and solves the `additionalErrors` use case without changing the contract of `data`.

packages/vue/src/components/JsonForms.vue
```diff
@@ -124,12 +129,9 @@ export default defineComponent({
   emits: ['change'],
```
sdirix: Add `update:data` here and emit it alongside `change` in the `eventToEmit` watcher, as agreed in https://github.com/eclipsesource/jsonforms/pull/2478#issuecomment-3646236151. With `v-model:data` the parent's data is never stale, so updating `additionalErrors` no longer resets the form, and that replaces the `dataToUse` sync below.

packages/core/test/generators/schema.test.ts
```diff
@@ -26,6 +26,37 @@ import test from 'ava';
...
+  });
+  t.deepEqual(generateJsonSchema(3.14), {
+    type: 'number',
+  });
+  t.deepEqual(generateJsonSchema(true), {
+    type: 'boolean',
+  });
+  t.deepEqual(generateJsonSchema(null), {
+    type: 'null',
+  });
```
sdirix, line 45: `generateJsonSchema(undefined)` now returns `{}` instead of throwing, a welcome fix that Angular directly relies on because it hands potentially undefined data straight in. Add a test for the `undefined` case too so this behavior can't silently regress.

packages/vue/src/components/JsonForms.vue
```diff
@@ -189,29 +191,31 @@ export default defineComponent({
...
-        Generate.uiSchema(
-          this.schemaToUse,
-          undefined,
-          undefined,
-          this.schemaToUse
-        );
+      this.uischemaToUse = newUischema ?? generateUISchema(this.schemaToUse);
     },
     data(newData) {
+      const isSameAsCurrentData = newData === this.dataToUse;
```
sdirix: Compare against `this.jsonforms.core.data` instead of `this.dataToUse`. That way echo detection keeps working once the sync in `eventToEmit` is removed, and `v-model:data` users don't pay schema generation plus `isEqual` on every edit.

sdirix (changes requested):

sdirix:
> @sdirix Do we still plan to merge this PR, or have we decided not to pursue this approach?
> 
> I'm asking because, besides the additionalErrors issue with form data, this PR also includes a fix for schema generation. Instead of:
> 
> `return gen.schemaObject(instance);`
> 
> it uses:
> 
> `return gen.property(instance);`
> 
> which is more correct, as it allows the generator to properly detect whether the input is a simple type, an array, or an object and generate the appropriate schema.
> 
> If we're no longer considering merging this PR, I can create a separate PR containing just the schema generation fix and then close this one.

@kchobantonov I re-reviewed everything today, including testing the behavior in the Vuetify dev app. Most of it is in good shape and I would recommend to merge it.

- The `generateJsonSchema` fix is correct and now well covered by tests.
- The generated-UI-schema fallback works. We would just like it moved into `generateDefaultUISchema` itself so React and Angular benefit as well.

The one part we don't want to merge as is: `this.dataToUse = newEvent.data`. It silently changes the semantics of the data prop for all Vue users and makes Vue diverge from React and Angular. As I wrote before we need to cleanly differentiate between "controlled" and "uncontrolled" data.

Concretely that means: keep the data prop semantics as they are and add the `update:data` event from the previous discussion, so `v-model:data` works out of the box. The `additionalErrors` use case is then solved by binding `v-model:data`, without JSON Forms mixing controlled and uncontrolled state internally. I left inline comments with the details.

If you'd rather land the schema generation fix quickly, splitting it into its own PR is fine with us and should merge fast. But if you're up for the `update:data` change, we can just as well finish everything here. Thanks again for driving this!

packages/vue/src/components/JsonForms.vue
kchobantonov: fixed in the latest update

packages/core/test/generators/schema.test.ts
kchobantonov: fixed in the latest update

packages/vue/src/components/JsonForms.vue
kchobantonov: fixed in the latest update

packages/vue/src/components/JsonForms.vue
kchobantonov: fixed in the latest update

kchobantonov:
@sdirix all reported issues are now addressed - please review.

packages/vue/src/components/JsonForms.vue
```diff
@@ -189,29 +185,31 @@ export default defineComponent({
...
-          undefined,
-          this.schemaToUse
-        );
+      this.uischemaToUse = newUischema ?? generateUISchema(this.schemaToUse);
     },
     data(newData) {
+      const isSameAsCurrentData = newData === this.jsonforms.core.data;
       this.dataToUse = newData;
+
+      if (!this.schema && !isSameAsCurrentData) {
```
EclipseSourceAI, line 202: `!this.schema` breaks `schema: false`, which is a legal JSON Schema and allowed by the prop type. Init uses `??` ([master](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/vue/src/components/JsonForms.vue#L128-L129)), so `schemaToUse` starts out as `false` and then gets silently replaced by a generated object schema on the first `data` prop change. Use `this.schema === undefined` to stay consistent.

packages/vue/src/components/JsonForms.vue
```diff
@@ -7,40 +7,39 @@
...
 
-// TODO fix @typescript-eslint/ban-types
-// eslint-disable-next-line @typescript-eslint/ban-types
-const isObject = (elem: any): elem is Object => {
-  return elem && typeof elem === 'object';
-};
-
 const EMPTY: ErrorObject[] = reactive([]);
 
+const getSchemaGeneratorInput = (data: any) => (data === undefined ? {} : data);
```
EclipseSourceAI, line 39: With `generateJsonSchema(undefined)` now returning `{}`, this local `undefined -> {}` mapping is dead weight and keeps Vue diverging from React, which passes `data` straight through ([master](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/react/src/JsonForms.tsx#L220-L222)). Same argument that moved the uischema fallback into the generator applies here: drop the helper and pass `dataToUse`.

packages/vue/src/components/JsonForms.vue
```diff
@@ -251,6 +249,7 @@ export default defineComponent({
       );
     },
     eventToEmit(newEvent) {
+      this.$emit('update:data', newEvent.data);
```
EclipseSourceAI, line 252: `mounted()` emits only `change`, so the initial core state never reaches a `v-model:data` parent. A middleware that transforms data inside `Actions.init` is then lost on the next unrelated prop change (`additionalErrors` hands `dataToUse` back into `updateCore`), which is the same bug class this PR fixes. Emit `update:data` in `mounted()` too.

packages/core/src/generators/uischema.ts
```diff
@@ -225,4 +225,4 @@ export const generateDefaultUISchema = (
   wrapInLayoutIfNecessary(
     generateUISchema(jsonSchema, [], prefix, '', layoutType, rootSchema),
     layoutType
-  );
+  ) ?? createLayout(layoutType);
```
EclipseSourceAI, line 228: `generateDefaultUISchema` returning an empty layout instead of `null` is a public API behavior change and deserves a MIGRATION.md entry under the existing 3.9 section ([master](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/MIGRATION.md#L3-L5)). It also makes the `foundUISchema !== null` branches unreachable in both `CombinatorProperties` implementations ([react](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/material-renderers/src/complex/CombinatorProperties.tsx#L62), [vue](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/vue-vuetify/src/complex/components/CombinatorProperties.vue#L68)), which could be cleaned up here.

packages/core/src/generators/schema.ts
```diff
@@ -177,5 +173,5 @@ export const generateJsonSchema = (
...
```
EclipseSourceAI, line 176: Good fix, but it changes results for existing callers: `generateJsonSchema([1, 2])` used to return an object schema with `"0"`/`"1"` properties and now returns `{type: 'array', items: {type: 'integer'}}`. Worth a MIGRATION.md entry alongside the uischema change.

packages/vue/README.md
```diff
@@ -82,7 +83,7 @@ export default defineComponent({
   },
   methods: {
     onChange(event: JsonFormsChangeEvent) {
-      this.data = event.data;
+      console.log(event);
```
EclipseSourceAI, line 86: `console.log(event)` doesn't show a reader anything. Either drop the handler from the sample or use it for something real (e.g. storing `event.errors`). The `data` prop docs above also still don't say it is controlled and must be kept in sync, which is what makes `v-model:data` the recommended binding.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
> This review was done by an AI agent and therefore may contain mistakes. Feel free to ignore any comment you disagree with. Noting *why* helps, since replies are read and taken into account in follow-up reviews.
>
>
>

Fixes the `additionalErrors` resetting form data in the Vue `JsonForms` component by adding an `update:data` event for `v-model:data` support, plus schema/uischema regeneration when the `data` prop changes externally in schema-less mode. Also carries two `@jsonforms/core` generator fixes: `generateJsonSchema` now handles non-object roots (primitives, arrays, `undefined`, `null`) via `gen.property`, and `generateDefaultUISchema` no longer returns `null`.

All previous review points from `sdirix` and `lucas-koehler` look addressed: the `dataToUse` sync is gone, the uischema fallback moved into the generator so React and Angular benefit, echo detection compares against `jsonforms.core.data`, and the generator tests cover the new cases.

Verified locally: build and lint clean, all package tests pass (the `angular-material` failure in my first run was a missing `CHROME_BIN`, not the PR). I also drove the Vuetify dev app: the `additional-errors` example keeps edited data across repeated "Add additional error" clicks, `#dynamic` regenerates from `Name` to `Id` on external data change, and root-level primitive data (`42`, `"hello"`) renders a single working control with no console errors.

Where to focus: the new `data` watcher in `packages/vue/src/components/JsonForms.vue`, specifically the `!this.schema` guard (breaks the legal `schema: false` case, verified), and whether the two public core generator behavior changes should get MIGRATION.md entries under the existing 3.9 section. The rest of the inline comments are smaller cleanups.

kchobantonov:
@sdirix now the EclipseSourceAI comments/review are resolved with the last commit

PR #2481, opened 2025-08-21
fix how we handle empty values when not in dynamic context (e.g. addi…

kchobantonov:
…tionalProperties case)

### DIFF
--- a/packages/vue-vuetify/src/controls/EnumControlRenderer.vue
+++ b/packages/vue-vuetify/src/controls/EnumControlRenderer.vue
@@ -58,7 +58,7 @@ const controlRenderer = defineComponent({
   setup(props: RendererProps<ControlElement>) {
     const clearValue = determineClearValue('');
     return useVuetifyControl(useJsonFormsEnumControl(props), (value) =>
-      value === null ? clearValue : value,
+      value !== null ? value : clearValue,
     );
   },
 });
--- a/packages/vue-vuetify/src/extended/AutocompleteEnumControlRenderer.vue
+++ b/packages/vue-vuetify/src/extended/AutocompleteEnumControlRenderer.vue
@@ -84,7 +84,7 @@ const controlRenderer = defineComponent({
     const clearValue = determineClearValue('');
     return useVuetifyControl(
       useJsonFormsEnumControl(props),
-      (value) => (value === null ? clearValue : value),
+      (value) => (value !== null ? value : clearValue),
       300,
     );
   },
--- a/packages/vue-vuetify/src/extended/AutocompleteOneOfEnumControlRenderer.vue
+++ b/packages/vue-vuetify/src/extended/AutocompleteOneOfEnumControlRenderer.vue
@@ -84,7 +84,7 @@ const controlRenderer = defineComponent({
     const clearValue = determineClearValue('');
     return useVuetifyControl(
       useJsonFormsOneOfEnumControl(props),
-      (value) => (value === null ? clearValue : value),
+      (value) => (value !== null ? value : clearValue),
       300,
     );
   },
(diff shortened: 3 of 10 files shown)


kchobantonov:
@sdirix please review - when we update the handling of empty values it looks like deleting (not using the x - clear ) was setting the value to empty string rather to remove the property when we are not under dynamic context (e.g. additionalProperties where we need to preserve the key)

packages/vue-vuetify/src/controls/EnumControlRenderer.vue
```diff
@@ -58,7 +58,7 @@ const controlRenderer = defineComponent({
   setup(props: RendererProps<ControlElement>) {
     const clearValue = determineClearValue('');
     return useVuetifyControl(useJsonFormsEnumControl(props), (value) =>
-      value === null ? clearValue : value,
+      value !== null ? value : clearValue,
```
sdirix: This is not an actual change.

packages/vue-vuetify/src/extended/AutocompleteEnumControlRenderer.vue
```diff
@@ -84,7 +84,7 @@ const controlRenderer = defineComponent({
     const clearValue = determineClearValue('');
     return useVuetifyControl(
       useJsonFormsEnumControl(props),
-      (value) => (value === null ? clearValue : value),
+      (value) => (value !== null ? value : clearValue),
```
sdirix: Thinking about it, this behavior is a bit dangerous for all enum related controls. `null` could be a valid enum value. Why are we not using `undefined` as the marker for the `clearValue` again?

packages/vue-vuetify/src/extended/AutocompleteOneOfEnumControlRenderer.vue
```diff
@@ -84,7 +84,7 @@ const controlRenderer = defineComponent({
     const clearValue = determineClearValue('');
     return useVuetifyControl(
       useJsonFormsOneOfEnumControl(props),
-      (value) => (value === null ? clearValue : value),
+      (value) => (value !== null ? value : clearValue),
```
sdirix: Not a change

packages/vue-vuetify/src/extended/AutocompleteEnumControlRenderer.vue
sdirix: Not a change

sdirix (commented):
Changes make sense to me in general. However there are a number of "non-changes" in the PR. Also I just noticed that we might have an overall enum issue.

kchobantonov:
@sdirix it is ready to be reviewed again - check my comments regarding the null value cases.

lucas-koehler (changes requested):
Hi @kchobantonov , thanks for the updates. The code looks pretty good to me. However, it seems to break something with the enum value display. See the video below. On master the same text is shown in the dropdown and the enum renderer's text field. Please have a look

[jf-vuetify-enums.webm](https://github.com/user-attachments/assets/093ba202-1251-4135-ace1-0f783a9b2f11)

kchobantonov:
will revert the return-object change and will wait until this issue with vuetify https://github.com/vuetifyjs/vuetify/issues/22126 is fixed to change the comparison for the null value to perhaps undefined

packages/vue-vuetify/src/controls/OneOfEnumControlRenderer.vue
```diff
@@ -56,8 +56,10 @@ const controlRenderer = defineComponent({
     ...rendererProps<ControlElement>(),
   },
   setup(props: RendererProps<ControlElement>) {
+    const clearValue = determineClearValue('');
+
     return useVuetifyControl(useJsonFormsOneOfEnumControl(props), (value) =>
-      value !== null ? value : undefined,
+      value === null ? clearValue : value,
```
lucas-koehler, line 62: We still have the problem of this not working with values that are `null` but should not be cleared. However, as this was also the case before this PR, this is fine for me and could potentially be fixed in a follow up.

PR #2483, opened 2025-08-23
feat(validator): add config to ValidateFunctionContext

kchobantonov:
The context now includes a `config` field that can be passed down to `ValidateFunction`.

This allows validator functions to use configuration data for their logic, such as delegating validation to an externally defined validator function if one is provided in the JSON Forms config object.


sdirix (approved):
Makes sense to me!

PR #2493, opened 2025-09-23
react-material: Update MUI X date pickers from v7 to v8

lucas-koehler:
- Update dev and peer dependency of `@mui/x-date-pickers` from v7 to v8.
- Adapt input variant handling for picker text fields
- Adapt picker variant handling
- Adapt textfield input property handling
- Fix pickers to still autoclose on selecting the smallest unit.
- Make autoclose configurable via uischema option `closeOnSelect`

**NOTE:**
The pickers needed to be updated to handle onChange in addition to onAccept. Otherwise, the value in the open picker is not updated. This results in data updates after each date/time component selection (e.g. hours).

### DIFF
--- a/packages/material-renderers/package.json
+++ b/packages/material-renderers/package.json
@@ -90,7 +90,7 @@
     "@jsonforms/react": "3.7.0-alpha.1",
     "@mui/icons-material": "^7.0.0",
     "@mui/material": "^7.0.0",
-    "@mui/x-date-pickers": "^7.28.0",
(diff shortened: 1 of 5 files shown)


sdirix (approved):
Works for me! I tested the variants and `closeOnSelect`.

The only weird thing for me is the behavior when textually inputting years. The ui does not change until the full year is entered:

Image

You can see the year changes at once (i.e. after 4! key presses) while month and day react immediately

PR #2504, opened 2025-11-20
feat: add prepend slot to date/time control renderers

yaffol:
Adds optional `prepend` slot to DateControlRenderer, DateTimeControlRenderer, and TimeControlRenderer to allow custom content before the input field.

**Benefits:**

- Enables custom icons/buttons before input field
- Follows existing Vuetify slot pattern (prepend-inner already exists)
- Backward compatible (slot is optional)
- Consistent across all date/time renderers

**Use Case:** Allows adding info/help icons or other custom content in the prepend area of v-text-field.

**Testing:** All existing tests pass. The conditional application with `v-if="$slots.prepend"` prevents and empty `<div class="v-input__prepend">` being injected to the rendered html.

**Extension:** If this pattern is acceptable, I could extend it across all the renderers, to allow access to the `prepend` slot for customisation.

### DIFF
--- a/packages/vue-vuetify/dev/renderers/TemperatureRenderer.vue
+++ b/packages/vue-vuetify/dev/renderers/TemperatureRenderer.vue
@@ -35,7 +35,7 @@ const { control } = useJsonFormsControl(props);
 // Determine temperature category (Celsius)
 const category = computed(() => {
   const temp = control.value.data || 0;
-  if (temp <= 4) return 'freezing';
+  if (temp <= 5) return 'freezing';
   if (temp < 15) return 'cold';
   if (temp < 25) return 'comfortable';
   if (temp < 35) return 'hot';


yaffol:
Hi @lucas-koehler,

Yes, that is correct. This allows custom renderers which wrap the upstream controls - such as a tooltip or copy to clipboard icon - which should be rendered in the field's `prepend` slot. The reason is useful to do this via the existing slots is that the elements line up very well across the form, and it allows you to add common customisations before or after an input, without needing to completely copy and maintain a custom renderer. If there is nothing passed in, no effect is applied to the renderer.

I have extended the support to all the renderers, and added `append` slots as well (apart from `PasswordRenderer.vue` which already uses the append slot via the built-in use of `append-icon` for show/hide password value. This allows for before/after content with any of the off-the-shelf renderers.

I've also created an example `Prepend/Append Slots (Basic)` in the example app, demonstrating some common use-cases.

packages/vue-vuetify/dev/renderers/TemperatureRenderer.vue
```diff
@@ -0,0 +1,77 @@
...
+import { VIcon, VTooltip } from 'vuetify/components';
+import { computed } from 'vue';
+
+const props = defineProps(rendererProps<ControlElement>());
+const { control } = useJsonFormsControl(props);
+
+// Determine temperature category (Celsius)
+const category = computed(() => {
+  const temp = control.value.data || 0;
+  if (temp <= 4) return 'freezing';
```
lucas-koehler: Minor: Should match the labels below
```suggestion
  if (temp <= 5) return 'freezing';
```

lucas-koehler (dismissed):
Hi @yaffol , Thanks for this improvement :heart: ! It adds a lot of flexibility and the examples showcase using this well!

lucas-koehler (approved):

PR #2520, opened 2025-12-18
Fix oneOf renderers not clearing some primitive values

### LINKED ISSUE #2506: Vue Vanilla OneOf renderer fails to properly switch between boolean and array values
### Describe the bug

Given the following JSON schema, if the user sets a boolean value then switches the oneOf control to the array schema and adds an array item, the error `TypeError: array.push is not a function` is given.

```json
{
  "oneOf": [
    {
      "title": "Boolean",
      "type": "boolean"
    },
    {
      "title": "Array",
      "type": "array",
        "items": {
        "type": "string"
      }
    }
  ]
}
```

### Expected behavior

The stored value should be reset and turned into an array so items can be added.

1. Setup an app with @jsonforms/vue-vanilla renderers
2. Supply the schema listed in the description above
3. Set an array value
4. Set a boolean value (notice the prompt to clear the data)
5. Set an array value (notice there is no prompt to clear the data)
6. Add an array item (notice the form does not change and an error is now present in the console)

Note: I don't think there needs to be prompt when switching away from a boolean value, but the data should be cleared.

### Screenshots

_No response_

3.6.0

### Package

Vue Vanilla Renderers

### Additional context

I've recreated a OneOfRenderer component and found the issue can be resolved by the following change. It will need more in-depth testing, but this works for me.

```js
// wrap useJsonFormOneOfControl with useVanillaControl
const input = useVanillaControl(useJsonFormsOneOfControl(props as ControlProps));
...
return {
  // remove useVanillaControl wrapper from the returned input
  ...input,
  ...
}
```

lucas-koehler:
Replace isEmpty check with check against undefined because isEmpty always returns true for boolean values, 0, empty strings and objects without props. This lead to errors when switching from any of these to an array and trying to add elements there.

Also adds unit tests for this.

fixes #2506

### DIFF
--- a/packages/material-renderers/src/complex/MaterialOneOfRenderer.tsx
+++ b/packages/material-renderers/src/complex/MaterialOneOfRenderer.tsx
@@ -23,7 +23,6 @@
   THE SOFTWARE.
 */
 import React, { useCallback, useMemo, useState } from 'react';
-import isEmpty from 'lodash/isEmpty';
 
 import { TabSwitchConfirmDialog } from './TabSwitchConfirmDialog';
 
@@ -98,7 +97,7 @@ export const MaterialOneOfRenderer = ({
   const handleTabChange = useCallback(
     (_event: any, newOneOfIndex: number) => {
       setNewSelectedIndex(newOneOfIndex);
-      if (isEmpty(data)) {
+      if (data === undefined) {
         openNewTab(newOneOfIndex);
       } else {
         setConfirmDialogOpen(true);


packages/vue-vanilla/src/complex/OneOfRenderer.vue
```diff
@@ -158,7 +157,7 @@ const controlRenderer = defineComponent({
       const target = event.target as any;
       this.selectIndex = target.value;
 
-      if (this.control.enabled && !isEmpty(this.control.data)) {
+      if (this.control.enabled && this.control.data !== undefined) {
```
sdirix, line 160: Should we also adapt the `isEmpty` in these three other renderers?
- [vue-vuetify OneOfRenderer.vue L177](https://github.com/eclipsesource/jsonforms/blob/3a9bc7449e96414026b9e7f2572b4d0d54000a89/packages/vue-vuetify/src/complex/OneOfRenderer.vue#L177)
- [vue-vuetify OneOfTabRenderer.vue L153](https://github.com/eclipsesource/jsonforms/blob/3a9bc7449e96414026b9e7f2572b4d0d54000a89/packages/vue-vuetify/src/complex/OneOfTabRenderer.vue#L153)
- [material-renderers MaterialOneOfRenderer.tsx L101](https://github.com/eclipsesource/jsonforms/blob/3a9bc7449e96414026b9e7f2572b4d0d54000a89/packages/material-renderers/src/complex/MaterialOneOfRenderer.tsx#L101)

packages/vue-vanilla/tests/unit/util/index.ts
```diff
@@ -1 +1,2 @@
+import '../setup';
```
sdirix: Nit: Importing here only makes the "setup" available if the tests also happen to import from util, i.e. it's a bit unintuitive. Personally I would prefer an explicit import or an adapted test setup.

PR #2522, opened 2025-12-28
Fix categorization stepper step order index, introduce new composition function for vue, fix config to be properly passed to test functions

kchobantonov:
1. Make sure that category stepper is working correctly when making categories visible/hidden.
2. Introduce new composition functions in vue package (moved and enhanced from the vue-vuetify package) that are useful for vue and not related to vuetify
3. Fix the DispatchRenderer and DispatchCell to use the resolved config and pass that to the TestContext - otherwise since there is not mapping for the config whenever the <dispatch-renderer> or <dispatch-cell> are called the config inside the testers is always undefined.


kchobantonov:
@sdirix please review - this is fixing an issue with categorization where when we make categories visible/hidden then the tag/view is switched - the example is here - https://deploy-preview-2522--jsonforms-examples.netlify.app/vue-vuetify/#categorization-stepper-nav-buttons - note that when you select/deselect "Provide Address" or "Vegetarian" you are still on the same tab and that does not change 

Also some type information for the Date/Time/DateTime renderers are fixed since the build was failing and also the vuetify library is updated

packages/vue-vuetify/src/controls/DateControlRenderer.vue
```diff
@@ -158,15 +157,18 @@ const controlRenderer = defineComponent({
...
       masked: '',
       unmasked: '',
       completed: false,
     });
-    const options = useMask
+
+    const state = computed(() => convertDayjsToMaskaFormat(dateFormat.value));
+    const locale = useLocale();
+
+    const options = useMask.value
```
sdirix: `useMask` is now a `computed`, but `useMask.value` is immediately evaluated to decide whether `options` is a `computed(...)` or `null`. So the options will not become reactive later on when `useMask` changes while the component is rendering.

packages/vue-vuetify/src/layouts/CategorizationStepperRenderer.vue
```diff
@@ -9,15 +9,15 @@
       :hide-actions="!appliedOptions.showNavButtons"
     >
       <v-stepper-vertical-item
-        :title="visibleCategoryLabels[index]"
-        v-for="(element, index) in visibleCategories"
-        :value="index + 1"
-        :key="`${layout.path}-${visibleCategories.length}-${index}`"
+        v-for="entry in visibleCategoriesWithIndex"
+        :title="entry.category.value.label"
+        :value="entry.originalIndex + 1"
```
sdirix: The stepper fix looks good. Using `originalIndex` to maintain stable `v-model` values when categories are hidden/shown is the right approach. The old code used filtered-array indices which would shift when visibility changed, causing the active step to jump.

packages/vue/src/jsonFormsCompositions.ts
```diff
@@ -528,3 +517,83 @@ export const useJsonFormsCategorization = (props: LayoutProps) => {
...
+  return translate;
+}
+
+/**
+ * Extracts Ajv from JSON Forms
+ */
+export function useAjv(): Ajv;
+export function useAjv(optional: true): Ajv | undefined;
+export function useAjv(optional?: true) {
+  const jsonforms = optional ? useJsonForms(true) : useJsonForms();
```
sdirix: can't you just hand over optional to `useJsonForms`?

packages/vue/src/jsonFormsCompositions.ts
```diff
@@ -528,3 +517,83 @@ export const useJsonFormsCategorization = (props: LayoutProps) => {
...
+/**
+ * Provides access to the JsonForms translator function.
+ * @returns The JsonForms translator function.
+ */
+export function useTranslator(): ComputedRef<Translator>;
+export function useTranslator(
+  optional: true
+): ComputedRef<Translator> | undefined;
+export function useTranslator(optional?: true) {
+  const jsonforms = optional ? useJsonForms(true) : useJsonForms();
```
sdirix: same here

sdirix (changes requested):
The changes overall look good!

Note that the PR title says "Fix categorization stepper step order index, introduce new composition function for vue, fix config to be properly passed to test functions" but it actually also includes:
- Vuetify version bump (3.9 -> 3.11)
- `useMask` reactivity fix in Date/Time/DateTime controls
- `clearable` behavior change (AND with `enabled`)
- `controlWrapper` source change (using `overwrittenControl` instead of `input.control`)
- `undefined` -> `null` type fixes in picker values
- `activator="parent"` removal from menus
- DateTime `onPickerChange` rewrite
- min/max exclusive fix for DateTimeControlRenderer

These are all independent fixes. Splitting into separate PRs would make each easier to review. This PR contained so much again that I pushed reviewing too later to when I have more time.

packages/vue/src/jsonFormsCompositions.ts
kchobantonov: that was because of the type restriction - will fix that

packages/vue/src/jsonFormsCompositions.ts
kchobantonov: type restriction again will fix that

packages/vue/src/jsonFormsCompositions.ts
kchobantonov: try to do that but if I pass the useJsonForms(optional) then I had to change what type this returns which make the useJsonForms() to return JsonFormsSubStates or undefined and we want that to always return JsonFormsSubStates and never undefined. so revering the change back to use 
`  const jsonforms = optional ? useJsonForms(true) : useJsonForms();`

packages/vue/src/jsonFormsCompositions.ts
```diff
@@ -38,7 +38,9 @@ import {
   Categorization,
   isControl,
   Scopable,
+  Translator,
 } from '@jsonforms/core';
+import Ajv from 'ajv';
```
sdirix: `Ajv` is only used as a type here, so this should be `import type Ajv from 'ajv';`.

packages/vue-vuetify/src/layouts/CategorizationStepperRenderer.vue
```diff
@@ -36,29 +36,29 @@
...
         <v-stepper-header>
           <template
-            v-for="(_, index) in visibleCategories"
-            :key="`${layout.path}-${visibleCategories.length}-${index}`"
+            v-for="(entry, index) in visibleCategoriesWithIndex"
+            :key="`${layout.path}-${entry.originalIndex}`"
           >
-            <v-stepper-item :value="index + 1" editable>
-              {{ visibleCategoryLabels[index] }}
+            <v-stepper-item :value="entry.originalIndex + 1" editable>
```
sdirix, line 43: `activeCategory` is initialized to `1` in [setup](https://github.com/eclipsesource/jsonforms/blob/31b9a7aaf5e2193bdd33dd4cb023bd43613f18d2/packages/vue-vuetify/src/layouts/CategorizationStepperRenderer.vue#L121-L131), but with this change the value bound on each step is `originalIndex + 1`. If the first category is hidden via a rule on initial render, no step will match `activeCategory === 1` and the stepper renders without an active step. The default needs to be the first *visible* `originalIndex + 1`.

sdirix, line 43: Using `originalIndex + 1` means the step badges shown to the user are not sequential when categories in the middle are hidden (e.g. badges "1", "3" with category 2 hidden). Is that on purpose? I guess it depends on the use case whether 2 will ever be shown or not?

sdirix (changes requested):

packages/vue-vuetify/src/layouts/CategorizationStepperRenderer.vue
kchobantonov: Good point. Both behaviors can make sense: keeping the original index communicates the stable category position, while sequential numbering keeps the currently visible steps visually consistent. Since the previous implementation showed sequential step numbers for visible categories, I’ll keep that behavior and only use the original index internally for the stepper value. This preserves the fix for not switching categories when visibility changes, while keeping the displayed badges sequential.

kchobantonov:
@sdirix changes are done - please review again - thanks

packages/vue-vuetify/src/layouts/CategorizationRenderer.vue
```diff
@@ -14,10 +14,10 @@
           direction="vertical"
         >
           <v-tab
-            v-for="(_, index) in visibleCategories"
-            :key="`${layout.path}-${visibleCategories.length}-${index}`"
+            v-for="entry in visibleCategoriesWithIndex"
```
sdirix, line 17: When the active tab's neighbour becomes hidden, the tab strip ends up with no visually selected tab because this is still position based.

packages/vue-vuetify/src/controls/DateTimeControlRenderer.vue
```diff
@@ -401,7 +403,7 @@ const controlRenderer = defineComponent({
         let date = parseDateTime(schema.formatExclusiveMinimum, this.formats);
         if (date) {
           // the format is exclusive
-          date = date.add(1, 'second');
+          date = date.add(1, this.useSeconds ? 'second' : 'minutes');
```
sdirix: `'minutes'` (plural) mixed with `'second'` (singular) is inconsistent with the rest of this file, which uses `'minute'` and `'second'`.

packages/vue/src/jsonFormsCompositions.ts
```diff
@@ -534,3 +523,83 @@ export const useJsonFormsCategorization = (props: LayoutProps) => {
...
+ * @returns The JsonForms sub state.
+ */
+export function useJsonForms(): JsonFormsSubStates;
+export function useJsonForms(optional: true): JsonFormsSubStates | undefined;
+export function useJsonForms(optional?: true) {
+  const jsonforms = inject<JsonFormsSubStates>('jsonforms');
+
+  if (!jsonforms && !optional) {
+    throw new Error(
+      "'jsonforms couldn't be injected. Are you within JSON Forms?"
```
sdirix: Pre-existing typo carried over from [vue-vuetify](https://github.com/eclipsesource/jsonforms/blob/53918bc8f9536bed3ded81f803b89e9be4137ae1/packages/vue-vuetify/src/util/composition.ts#L335): missing closing single quote after `jsonforms`. Same problem on lines 558, 578, 600 for `useDispatch`, `useTranslator`, `useAjv`. Worth fixing while you're already touching this code.

packages/vue/src/jsonFormsCompositions.ts
```diff
@@ -534,3 +523,83 @@ export const useJsonFormsCategorization = (props: LayoutProps) => {
...
+      "'jsonforms i18n couldn't be injected. Are you within JSON Forms?"
+    );
+  }
+
+  const translate = computed(() => {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return jsonforms!.i18n!.translate!;
+  });
+
+  return translate;
```
sdirix, line 596: When `optional=true` and no jsonforms is injected, this still returns the computed (the overload says `ComputedRef<Translator> | undefined`, but the body always returns `translate`). Accessing `.value` then dereferences `jsonforms!.i18n!.translate!` on `undefined` and throws. That breaks the use case in [`ValidationBadge.vue`](https://github.com/eclipsesource/jsonforms/blob/53918bc8f9536bed3ded81f803b89e9be4137ae1/packages/vue-vuetify/src/controls/components/ValidationBadge.vue#L121-L132), which previously got a clean `null` and now will throw the first time the template evaluates `t`. Return `undefined` early when jsonforms (or its translator) is missing in the optional path.

sdirix (changes requested):

kchobantonov:
@sdirix All reported issues have now been resolved, please review again - thank you.

sdirix (approved):
Thank you!

PR #2531, opened 2026-01-14
fix(core): Fix validation errors for nested props of same name

### LINKED ISSUE #2521: Required validation missing for nested property of same name as parent
### Describe the bug

Used libraries:
@jsonforms/core: ^3.7.0.
@jsonforms/react: ^3.7.0.
@jsonforms/material-renderers: ^3.7.0.

If a nested object has the same property name as its parent object, the "is a required property" message is not displayed.

### Expected behavior

An error message should be displayed ("is a required property").

Use these:

1. Schema:
```
{
    type: 'object',
    properties: {
        name: {
            type: 'object',
            properties: { name: { type: 'string' } },
            required: ['name']
        }
    }
}
```

2. UI Schema `{ type: 'Control', scope: '#/properties/name/properties/name', label: 'Name' }`
3. Data `{ name: {} }`
4. JsonForms
```
<JsonForms
    data={data}
    schema={dataSchema}
    uischema={uiSchema}
    renderers={materialRenderers}
    cells={materialCells}
/>
```

Out of interest, this issue goes away if you rename the inner `name` property to something other than 'name'.

### Screenshots

_No response_

v3.7.0

### Package

_No response_

### Additional context

_No response_

lucas-koehler:
- Remove obsolete safeguard against adding a property name a second time to a validation error's control path.
- Add comment in errors.ts to clarify behavior
- Add unit tests for getControlPath and errorAt
- Add example for nested required property of same name

fixes #2521

### DIFF
--- a/packages/core/src/util/errors.ts
+++ b/packages/core/src/util/errors.ts
@@ -32,7 +32,7 @@ import isEqual from 'lodash/isEqual';
 
 /**
  * Checks for an additionally specified property that the error relates to.
- * This may be added to an error's instancePath to show it add the violating property's control.
+ * This may be added to an error's instancePath to show it at the violating property's control.
  * For example, for required property errors, the instancePath points to the object containing the required property.
  * The missing property's name is specified in the error's params.missingProperty field and returned by this function.
  *
--- a/packages/core/test/reducers/core.test.ts
+++ b/packages/core/test/reducers/core.test.ts
@@ -2201,7 +2201,7 @@ test('getControlPath - nested required property', (t) => {
   t.is(controlPath, 'parent.child');
 });
 
-test('getControlPath - prevents duplicate property when path already ends with property', (t) => {
+test('getControlPath - same-named nested properties in different parent', (t) => {
   const errorObject = {
     instancePath: '/parent/child',
     keyword: 'required',
--- a/packages/examples/src/index.ts
+++ b/packages/examples/src/index.ts
@@ -79,7 +79,7 @@ import * as mixed from './examples/mixed';
 import * as mixedObject from './examples/mixed-object';
 import * as string from './examples/string';
 import * as prependAppendSlots from './examples/prepend-append-slots';
-import * as validationNestedSameNameValidation from './examples/validation-nested-same-name';
+import * as validationNestedSameName from './examples/validation-nested-same-name';
 export * from './register';
 export * from './example';
 
@@ -148,5 +148,5 @@ export {
   arrayWithDefaults,
   string,
   prependAppendSlots,
-  validationNestedSameNameValidation,
+  validationNestedSameName,
 };


packages/core/src/util/errors.ts
```diff
@@ -30,6 +30,15 @@ import { isOneOfEnumSchema } from './schema';
 import filter from 'lodash/filter';
 import isEqual from 'lodash/isEqual';
 
+/**
+ * Checks for an additionally specified property that the error relates to.
+ * This may be added to an error's instancePath to show it add the violating property's control.
```
sdirix: Typo: "to show it add the violating" should be "to show it at the violating".

packages/core/src/util/errors.ts
```diff
@@ -51,12 +60,12 @@ export const getControlPath = (error: ErrorObject) => {
...
 
   const invalidProperty = getInvalidProperty(error);
-  if (invalidProperty !== undefined && !controlPath.endsWith(invalidProperty)) {
+  if (invalidProperty !== undefined) {
     controlPath = `${controlPath}.${invalidProperty}`;
   }
 
   // remove '.' chars at the beginning of paths
-  controlPath = controlPath.replace(/^./, '');
+  controlPath = controlPath.replace(/^\./, '');
```
sdirix, line 68: Good catch 👍

packages/core/test/reducers/core.test.ts
```diff
@@ -1870,3 +1870,432 @@ test('core reducer helpers - getControlPath - decodes JSON Pointer escape sequen
...
+    instancePath: '/parent',
+    keyword: 'required',
+    params: { missingProperty: 'child' },
+    schemaPath: DUMMY_SCHEMA_PATH,
+  } as ErrorObject;
+  const controlPath = getControlPath(errorObject);
+  t.is(controlPath, 'parent.child');
+});
+
+test('getControlPath - prevents duplicate property when path already ends with property', (t) => {
```
sdirix: This test name is misleading. It says "prevents duplicate" but this is not what we want?

packages/examples/src/index.ts
```diff
@@ -79,6 +79,7 @@ import * as mixed from './examples/mixed';
 import * as mixedObject from './examples/mixed-object';
 import * as string from './examples/string';
 import * as prependAppendSlots from './examples/prepend-append-slots';
+import * as validationNestedSameNameValidation from './examples/validation-nested-same-name';
 export * from './register';
```
sdirix, line 83: nit: `validationNestedSameNameValidation` has redundant "Validation" in the name. `validationNestedSameName` would be enough.

packages/core/src/util/errors.ts
```diff
@@ -51,12 +60,12 @@ export const getControlPath = (error: ErrorObject) => {
   controlPath = controlPath.replace(/\//g, '.');
 
   const invalidProperty = getInvalidProperty(error);
-  if (invalidProperty !== undefined && !controlPath.endsWith(invalidProperty)) {
+  if (invalidProperty !== undefined) {
```
sdirix, line 63: Do you know why we originally added the `endsWith` check? There must have been a reason

lucas-koehler: I'm not completely sure. Three reasons come to mind:
- AJV calculated error paths differently in the past
- This was a precaution that was never actually necessary
- We handled more cases in `getInvalidProperty` explicitly in the past. However, the current three cases all need the `invalidProperty` attached anyway because they point to the parent object. See code snipped below

```typescript
const getInvalidProperty = (error: ErrorObject): string | undefined => {
  switch (error.keyword) {
    case 'required':
    case 'dependencies':
      return error.params.missingProperty;
    case 'additionalProperties':
      return error.params.additionalProperty;
    default:
      return undefined;
  }
};
```

PR #2532, opened 2026-01-15
feat: add separateReadonlyFromDisabled config option

### LINKED ISSUE #2479: Allow custom renderers to handle the readonly behaviour.

When using a custom textbox renderer and trying to make it readonly via options, JSON forms is disabling the textbox by passing the `enabled` prop as false to the renderer. 

Ideally this should be upto the renderer to decide whether or not to disable the field based on the readonly option.

As discussed here: https://github.com/eclipsesource/jsonforms/issues/2283 - 

>It seems that the core runtime is converting readonly to enabled/disabled and not the renderer sets, so it's not possible to override this from renderer sets reliably.

>https://github.com/eclipsesource/jsonforms/issues/1735#issuecomment-856632817
> https://github.com/eclipsesource/jsonforms/blob/master/packages/core/src/mappers/util.ts#L24
 
Readonly and disabled states are semantically different and has different meanings, behaviour and styles.

I would like for a custom renderer to receive the enabled prop based only on the rules and not based on any other options defined for the ui schema element. The custom renderer should be able to recieve the enabled prop as well as the uischema options, global config and then decide how to render the field.

I am considering renaming the option `readonly` to `isReadOnly` to avoid the issue in the current version of JSON forms. but would love to see a proper solution than this workaround.

### Package

Core

### Additional context

Feature request raised as per this comment..

https://github.com/eclipsesource/jsonforms/issues/2283#issuecomment-3182800446

kchobantonov:
Allows independent control of readonly and enabled states. Adds READONLY/WRITABLE rule effects. Backward compatible (default: false).

fixes #2479

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "3.7.0",
+  "version": "3.8.0-alpha.0",
   "npmClient": "pnpm",
   "$schema": "node_modules/lerna/schemas/lerna-schema.json"
 }
--- a/packages/angular-material/package.json
+++ b/packages/angular-material/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jsonforms/angular-material",
-  "version": "3.7.0",
+  "version": "3.8.0-alpha.0",
   "description": "Material Renderer Set for Angular module of JSON Forms",
   "repository": {
     "type": "git",
@@ -72,8 +72,8 @@
     "@angular/material": "^19.0.0 || ^20.0.0 || ^21.0.0",
     "@angular/platform-browser": "^19.0.0 || ^20.0.0 || ^21.0.0",
     "@angular/router": "^19.0.0 || ^20.0.0 || ^21.0.0",
-    "@jsonforms/angular": "3.7.0",
-    "@jsonforms/core": "3.7.0",
+    "@jsonforms/angular": "3.8.0-alpha.0",
+    "@jsonforms/core": "3.8.0-alpha.0",
     "dayjs": "^1.11.10",
     "rxjs": "^6.6.0 || ^7.4.0"
   },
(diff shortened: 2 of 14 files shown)


kchobantonov:
@lucas-koehler please review

packages/vue-vuetify/src/additional/ListWithDetailRenderer.vue
```diff
@@ -124,7 +125,11 @@
                         small
                         class="ma-0"
                         :aria-label="control.translations.downAriaLabel"
-                        :disabled="index >= dataLength - 1 || !control.enabled"
+                        :disabled="
+                          index >= dataLength - 1 ||
+                          !control.enabled ||
+                          control.readonly
```
sdirix: The move-up button on line 107 uses `isControlEditable(control)`, but this move-down button (and the add/delete buttons) inline `!control.enabled || control.readonly`. Use `isControlEditable` everywhere for consistency. Same pattern appears in `ArrayControlRenderer.vue` and `ArrayLayoutRenderer.vue`.

packages/vue-vuetify/src/complex/OneOfRenderer.vue
```diff
@@ -102,6 +104,7 @@ import {
 } from 'vuetify/components';
 import { DisabledIconFocus } from '../controls/directives';
 import {
+  isControlEditable,
```
sdirix: Unused import.

kchobantonov: thanks for spotting those... will fix

packages/vue-vuetify/src/additional/ListWithDetailRenderer.vue
kchobantonov: will fix those thanks

kchobantonov:
@sdirix @lucas-koehler  please review again - thanks

PR #2533, opened 2026-01-19
refactor: Improve spacing behavior for nested layouts with conditional visibility

kchobantonov:
1. The vertical, horizontal renderers were adding unwanted padding and spacing even when child elements were hidden or not rendered. This caused layout issues with:
- Empty spaces appearing when controls had `visible: false`
- Excessive padding accumulating in deeply nested layouts
- Inconsistent spacing between visible elements

2. Make the group layout and array layouts/control to use collapsable panels to make it easy to deal with array with large number of elements if we want to collapse the whole array UI into just one row.

3. Make the space usage similar to react - e.g. remove exta margins and/or paddings

4. Other minor fixes

### DIFF
--- a/packages/angular-material/src/library/controls/autocomplete.renderer.ts
+++ b/packages/angular-material/src/library/controls/autocomplete.renderer.ts
@@ -47,28 +47,6 @@ import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatInputModule } from '@angular/material/input';
 import { MatAutocompleteModule } from '@angular/material/autocomplete';
 
-/**
- * To use this component you will need to add your own tester:
- * <pre><code>
- * ...
- * export const AutocompleteControlRendererTester: RankedTester = rankWith(2, isEnumControl);
- * ...
- * </code></pre>
- * Add the tester and renderer to JSONForms registry:
- * <pre><code>
- * ...
- * { tester: AutocompleteControlRendererTester, renderer: AutocompleteControlRenderer },
- * ...
- * </code></pre>
- * Furthermore you need to update your module.
- * <pre><code>
- * ...
- * imports: [JsonFormsAngularMaterialModule, MatAutocompleteModule],
- * declarations: [AutocompleteControlRenderer]
- * ...
- * </code></pre>
- *
- */
 @Component({
   selector: 'AutocompleteControlRenderer',
   template: `


kchobantonov:
@sdirix @lucas-koehler please review - new demo example [employee-registration](https://deploy-preview-2533--jsonforms-examples.netlify.app/vue-vuetify/#employee-registration)

packages/vue-vuetify/src/layouts/GroupRenderer.vue
```diff
@@ -74,6 +100,28 @@ const layoutRenderer = defineComponent({
...
       return classes.join(' ');
     },
+    visibleElementsWithIndex(): { element: UISchemaElement; index: number }[] {
+      return this.layout.uischema.elements
+        .map((element, index) => ({ element, index }))
+        .filter(({ element }) => this.isVisible(element, this.layout.path));
+    },
+  },
+  methods: {
+    isVisible(uischema: UISchemaElement, path: string): boolean {
```
sdirix, line 110: This `isVisible` method is copy-pasted across 3 files: HorizontalLayoutRenderer, VerticalLayoutRenderer, GroupRenderer. Maybe this should be a shareable composition? Maybe the same for `visibleElementsWithIndex`? This wouild also allow to unify the behavior between Horizontal and Vertical rendering as they implement this logic differently now.

packages/vue/src/components/UnknownRenderer.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>No applicable renderer found.</div>
+  <div :style="{ color: 'red' }">No applicable renderer found.</div>
```
sdirix, line 2: We usually avoid hard coded styling in the non-renderer and Vanilla packages. This should get a proper class so it can be styled as required by adopters.

I know we do it in React too but this is more of an old historical oversight.

I would prefer using a class over inline styling.

pnpm-lock.yaml
sdirix, line 1: This change is unnecessary

packages/vue-vuetify/src/layouts/VerticalLayoutRenderer.vue
```diff
@@ -2,16 +2,18 @@
...
   >
     <v-row
       v-for="(element, index) in (layout.uischema as Layout).elements"
       :key="`${layout.path}-${(layout.uischema as Layout).elements.length}-${index}`"
-      v-bind="vuetifyProps(`v-row[${index}]`)"
+      v-bind="{ ...vuetifyProps(`v-row`), ...vuetifyProps(`v-row[${index}]`) }"
     >
       <v-col
         cols="12"
+        v-if="isVisible(element, layout.path)"
```
sdirix, line 16: Instead of this v-if we could already iterate over the filtered list like in the other renderers

packages/vue-vuetify/src/layouts/VerticalLayoutRenderer.vue
```diff
@@ -52,7 +62,26 @@ const layoutRenderer = defineComponent({
     ...rendererProps<Layout>(),
   },
   setup(props: RendererProps<Layout>) {
-    return useVuetifyLayout(useJsonFormsLayout(props));
+    const jsonforms = useJsonForms();
+    return { ...useVuetifyLayout(useJsonFormsLayout(props)), jsonforms };
+  },
+  methods: {
+    // helper method to check visibility and if not visible to remove the v-col since otherwise it will add extra padding
```
sdirix, line 69: Not sure the comment is needed here

packages/vue-vuetify/src/layouts/GroupRenderer.vue
```diff
@@ -1,61 +1,87 @@
...
-      v-if="layout.label"
-      :class="styles.group.label"
-      v-bind="vuetifyProps('v-card-title')"
-      >{{ layout.label }}</v-card-title
-    >
+    <v-expansion-panel>
+      <v-expansion-panel-title
+        :class="styles.group.label"
+        v-bind="vuetifyProps('v-expansion-panel-title')"
+        >{{ layout.label }}</v-expansion-panel-title
```
sdirix, line 14: Now that Groups are clickable it's a bit weird that the label can be empty. However I also don't have a good suggestion for this case. Maybe add an icon if the label is empty? Or revert to the old v-card grouping? Not sure. See the `rule` example to check how this looks like in the UI.

sdirix (changes requested):
Thanks for the contribution ❤️ 

Again this is difficult to review. This PR has 6+ different concerns wrapped into one pull request. Separate small PRs would make reviewing easier and would reduce the wait time for you to get feedback.

This PR addresses:
 - Layout visibility filtering (main goal?)
 - Converting v-card to v-expansion-panels
 - Spacing/padding cleanups
 - Unrelated control fixes
 - Dev changes
 - Unrelated Vue package changes, not directly tied to Vue-Vuetify

There are so many functionality and layout changes that it makes it really hard to test all of them besides a shallow "looks fine to me".

Please split them in the future.

---

Functionality wise most changes look fine.

The `v-expansion-panels` refactoring is an obvious breaking change for all adopters using them, in behavior as well as in configuration (e.g. when they used 'v-card' props to customize them). So this should be mentioned in the Breaking Changes / Migration guide. Again this would be easier as a separate PR.

Overall I don't have confidence for all the layout changes, I might have overlooked something.

Please have a look at the comments.

packages/vue-vuetify/src/layouts/HorizontalLayoutRenderer.vue
```diff
@@ -1,16 +1,20 @@
...
+    :class="`pa-0 ${styles.horizontalLayout.root}`"
+    fluid
     v-bind="vuetifyProps('v-container')"
   >
     <v-row v-bind="vuetifyProps('v-row')">
       <v-col
-        v-for="(element, index) in (layout.uischema as Layout).elements"
-        :key="`${layout.path}-${(layout.uischema as Layout).elements.length}-${index}`"
+        v-for="{ element, index } in visibleElementsWithIndex"
+        :key="`${layout.path}-${visibleElementsWithIndex.length}-${index}`"
```
sdirix, line 11: Using `visibleElementsWithIndex.length` here makes all keys change whenever _any_ child's visibility toggles, forcing Vue to unmount and remount every child component.

packages/vue-vuetify/src/layouts/GroupRenderer.vue
```diff
@@ -1,61 +1,87 @@
...
-        :path="layout.path"
-        :enabled="layout.enabled"
-        :renderers="layout.renderers"
-        :cells="layout.cells"
-      />
-    </v-card-text>
-  </v-card>
+      <v-expansion-panel-text
+        v-for="{ element, index } in visibleElementsWithIndex"
+        :key="`${layout.path}-${visibleElementsWithIndex.length}-${index}`"
```
sdirix, line 19: Same issue here

packages/vue-vuetify/src/layouts/VerticalLayoutRenderer.vue
```diff
@@ -2,16 +2,18 @@
   <v-container
     v-if="layout.visible && (layout.uischema as Layout).elements.length > 0"
     fill-height
-    :class="`${styles.verticalLayout.root}`"
+    fluid
+    :class="`pa-0 ${styles.verticalLayout.root}`"
     v-bind="vuetifyProps('v-container')"
   >
     <v-row
```
sdirix, line 9: Should the v-row also be skipped if it's empty?

packages/vue-vuetify/src/controls/RadioGroupControlRenderer.vue
```diff
@@ -19,6 +19,7 @@
       :required="control.required"
       :error-messages="control.errors"
       :model-value="control.data"
+      :inline="inline"
```
sdirix: This is a breaking change. On master, no `inline` was set, so Vuetify's default (vertical stacking) applied. Now all radio groups default to horizontal. If this is intentional, it needs to go in the migration notes.

packages/vue-vuetify/src/layouts/HorizontalLayoutRenderer.vue
```diff
@@ -1,16 +1,20 @@
 <template>
   <v-container
     v-if="layout.visible && (layout.uischema as Layout).elements.length > 0"
-    :class="`${styles.horizontalLayout.root}`"
+    :class="`pa-0 ${styles.horizontalLayout.root}`"
```
sdirix, line 4: This changes *all* HorizontalLayout renderers, not just deeply nested ones.

packages/vue-vuetify/src/layouts/ArrayLayoutRenderer.vue
```diff
@@ -352,16 +375,33 @@ const controlRenderer = defineComponent({
...
       ...control,
+      openedPanels,
       currentlyExpanded,
       expansionPanelsProps,
       suggestToDelete,
       icons,
     };
   },
   computed: {
+    isSelected(): boolean {
```
sdirix, line 390: `isSelected`, `expandIcon`, `collapseIcon`, `togglePanel`, and `openedPanels` are duplicated verbatim in `ArrayControlRenderer.vue`.

packages/vue-vuetify/src/controls/OneOfRadioGroupControlRenderer.vue
```diff
@@ -18,6 +18,7 @@
       :persistent-hint="persistentHint()"
       :required="control.required"
       :error-messages="control.errors"
+      :inline="inline"
```
sdirix, line 21: Is this change intentional? It flips the rendering from vertical to horizontal which may look very bad for larger options. Same for the regular radio group.

packages/vue-vuetify/src/layouts/GroupRenderer.vue
```diff
@@ -1,61 +1,87 @@
...
+    VExpansionPanelText,
+    VExpansionPanelTitle,
+    VExpansionPanels,
   },
   props: {
     ...rendererProps<Layout>(),
   },
   setup(props: RendererProps<Layout>) {
-    return useVuetifyLayout(useJsonFormsLayout(props));
+    const openedPanels = ref<number | number[]>(0);
```
KiaraGrouwstra, line 77: fyi also allowing them to start closed would involve:
1. eager loading of collapsed fields, to mount and write their defaults (`v-expansion-panel-text` is lazy)
2. expanding fields with validation errors, so the user would know what to fix

packages/vue-vuetify/src/layouts/GroupRenderer.vue
```diff
@@ -1,61 +1,87 @@
...
+    v-bind="vuetifyProps('v-expansion-panels')"
   >
-    <v-card-title
-      v-if="layout.label"
-      :class="styles.group.label"
-      v-bind="vuetifyProps('v-card-title')"
-      >{{ layout.label }}</v-card-title
-    >
+    <v-expansion-panel>
+      <v-expansion-panel-title
```
KiaraGrouwstra, line 11: The array titles show a `validation-icon` when a child has an error. The group title does not. When a group is closed, the user cannot see errors in it. Can the group title show the same icon, using the errors under the group's path?

packages/vue-vuetify/src/layouts/ArrayLayoutRenderer.vue
```diff
@@ -1,271 +1,296 @@
...
-          </v-tooltip>
-        </slot>
-      </v-toolbar>
-    </v-card-title>
-    <v-card-text v-bind="vuetifyProps('v-card-text')">
-      <v-container
-        justify-space-around
-        align-content-center
-        :class="styles.arrayList.container"
+          <v-toolbar-title :class="styles.arrayList.label">{{
```
KiaraGrouwstra, line 19: When the panel is closed, the title does not show whether the array is empty. `dataLength` is already there. Can the title show it, for example `Items (3)`? The same applies to `ArrayControlRenderer.vue` and to the entry count in `AdditionalProperties.vue:9`.

packages/vue-vuetify/src/layouts/ArrayLayoutRenderer.vue
KiaraGrouwstra: +1. A shared composable can also hold the state of all panels in the form (for example through provide/inject). Then "expand all" and "collapse all" become possible, and a validation error can open its panel and the panels above it. If a panel mounts after a command, for example when the user adds an array item, it can read the last command when it registers.

KiaraGrouwstra:
We use [collapsible sections](https://git.fediversity.eu/fediversity/nix-module-form/pulls/24) in a form generated from a large, deeply nested schema. It has several hundred controls. Some lessons from that work may help the collapsible-panels part of this PR, possibly as the separate PR that was suggested:

- If a panel can start closed, a closed panel can hide a validation error. We open each panel that contains a failing field, and all panels above it.
- Keep the content of a closed panel mounted, and only hide it. Some renderers write a default value into the data when they mount. If closed content is not mounted, the data depends on which panels the user opened.
- A closed header is more useful when it shows what it contains: the number of items or entries, and whether there are errors.
- With many panels, users need "expand all" and "collapse all". That needs one shared state for all panels. It also removes the duplicate panel state (`openedPanels`, `togglePanel`, `isSelected`) in the array renderers.

Details are in the inline comments.

Assisted-by: `claude-code-opus-5-5`

PR #2535, opened 2026-01-21
Angular Material: Translate Autocomplete Control Renderer

daniel-shuy:
# Background

The company I work for ([SICPA Product Security](https://www.sicpa.com/)) is considering adopting JSON Forms for a project.

_Unfortunately_ the project is written in Angular, [and I noticed that the Angular Material renderer set is missing quite a lot of supported features compared to the React/Vue renderer sets](https://jsonforms.io/docs/renderer-sets).

As such, I have been testing the Angular Material renderer set quite thoroughly, and will be contributing some missing features, especially to the Angular Material renderer set.

# Description

The Angular Material renderer set renders [enums](https://jsonforms.io/docs/multiple-choice#enum) using the `AutocompleteControlRenderer`.

Unlike the other renderer sets, `AutocompleteControlRenderer` does not translate the enum values.

This PR updates `AutocompleteControlRenderer` to use `mapStateToEnumControlProps(JsonFormsState, OwnPropsOfControl & OwnPropsOfEnum)` like the other renderer sets, which translates the enum values as labels, while keeping the original values as the values:
https://github.com/eclipsesource/jsonforms/blob/e642b3d66851182cc6f98b2a08f3af534317c63d/packages/core/src/mappers/renderer.ts#L676-L707

For backwards compatibility, I've allowed the `options` [@Input](https://angular.dev/api/core/Input) to still take in `string[]` (by typing it as `EnumOption[] | string[]`), but it will not be translated if a `string[]` is passed.

### DIFF
--- a/packages/angular-material/src/library/controls/autocomplete.renderer.ts
+++ b/packages/angular-material/src/library/controls/autocomplete.renderer.ts
@@ -130,9 +130,9 @@ export class AutocompleteControlRenderer
 
   override onChange(ev: any) {
     const eventValue = this.getEventValue(ev);
-    const option = Array.from(this.valuesToTranslatedOptions?.values() ?? []).find(
-      (option) => option.label === eventValue
-    );
+    const option = Array.from(
+      this.valuesToTranslatedOptions?.values() ?? []
+    ).find((option) => option.label === eventValue);
     if (!option) {
       super.onChange(ev);
       return;


packages/angular-material/src/library/controls/autocomplete.renderer.ts
```diff
@@ -164,24 +182,44 @@ export class AutocompleteControlRenderer
     this.triggerValidation();
   }
 
-  filter(val: string): string[] {
-    return (this.options || this.scopedSchema.enum || []).filter(
+  displayFn(option: EnumOption): string {
+    return option.label;
+  }
```
lucas-koehler: This function signature suggests it always receives an `EnumOption`, but Angular Material's autocomplete will pass the raw value (string) when displaying the current selection in the input field.


```suggestion
  displayFn(option: EnumOption | string): string {
    if (!option) {
      return '';
    }
    return typeof option === 'string' ? option : option.label;
  }
```

packages/angular-material/test/autocomplete-control.spec.ts
lucas-koehler, line 1: Several `expect(inputElement.value)` assertions were removed from the tests. Was this intentional due to the `displayFn` changes affecting how values are displayed? If so, it would be helpful to add replacement assertions that verify the translated label is displayed correctly in the input field, rather than removing the coverage entirely.

packages/angular-material/src/library/controls/autocomplete.renderer.ts
daniel-shuy: I made a mistake, turns out `displayWith` is not even needed, I was overcomplicating it

packages/angular-material/test/autocomplete-control.spec.ts
daniel-shuy: They are working again without `displayWith`, I've added them back

daniel-shuy:
@lucas-koehler sorry, I had some leftover code that depended on `displayWith`. I've removed it, please try again

lucas-koehler:
Hi @daniel-shuy ,
I'd say it's acceptable to break the API here as long as it only affects this renderer resp. renderer set. We do not want to change the core behavior of JSON Forms :)

daniel-shuy:
@lucas-koehler Great! I've made the changes and updated the tests, it should work now

lucas-koehler (changes requested):
Thank you for the updates but something is still missing.
When I select an option in the enum i18n example at the preview https://deploy-preview-2535--jsonforms-examples.netlify.app/angular-material/ the control stays empty:

https://github.com/user-attachments/assets/6fd4af94-140e-4dff-b8ba-e71c60f06149

packages/angular-material/src/library/controls/autocomplete.renderer.ts
```diff
@@ -105,16 +110,40 @@ export class AutocompleteControlRenderer
...
+    state: JsonFormsState
+  ): StatePropsOfControl & OwnPropsOfEnum {
+    return mapStateToEnumControlProps(state, this.getOwnProps());
+  }
+
   getEventValue = (event: any) => event.target.value;
 
+  override onChange(ev: any) {
+    const eventValue = this.getEventValue(ev);
+    const option = Array.from(this.valuesToTranslatedOptions?.values()).find(
```
lucas-koehler: Array.from(undefined) throws an error, thus have a safe fallback:

```suggestion
    const option = Array.from(this.valuesToTranslatedOptions?.values() ?? []).find(
```

packages/angular-material/src/library/controls/autocomplete.renderer.ts
```diff
@@ -124,6 +153,12 @@ export class AutocompleteControlRenderer
     );
   }
 
+  override mapAdditionalProps(_props: StatePropsOfControl & OwnPropsOfEnum) {
+    this.valuesToTranslatedOptions = new Map(
+      _props.options.map((option) => [option.value, option])
```
lucas-koehler: Options may be undefined.

```suggestion
      (_props.options ?? []).map((option) => [option.value, option])
```

packages/angular-material/src/library/controls/autocomplete.renderer.ts
```diff
@@ -136,30 +171,71 @@ export class AutocompleteControlRenderer
...
-        option.toLowerCase().indexOf(val.toLowerCase()) === 0
+  // use arrow function to bind "this" reference
+  displayFn = (option?: string | EnumOption): string => {
+    if (!option) {
+      return '';
+    }
+
+    if (typeof option === 'string') {
+      if (!this.valuesToTranslatedOptions) {
+        return '';
```
lucas-koehler: We should rather show the raw, untranslated value than nothing as long as the translation is not available. Or do you see any issues with that?

```suggestion
        return option; // show raw value until translations are ready
```

packages/angular-material/src/library/controls/autocomplete.renderer.ts
```diff
@@ -136,30 +171,71 @@ export class AutocompleteControlRenderer
...
-      options: this.options,
+      options: this.stringOptionsToEnumOptions(this.options),
     };
   }
-}
 
-export const enumControlTester: RankedTester = rankWith(2, isEnumControl);
+  trackByEnumOptionId(_index: number, option: EnumOption): string {
+    return option.value;
+  }
```
lucas-koehler: This method seems to be unused now. Can it be removed?

packages/angular-material/src/library/controls/autocomplete.renderer.ts
daniel-shuy: ah, makes sense

packages/angular-material/src/library/controls/autocomplete.renderer.ts
daniel-shuy: Ah oops, good catch, I previously added the [trackBy](https://angular.dev/api/common/NgForOf#ngForTrackBy) expression to [ngFor](https://angular.dev/api/common/NgFor) directive, but then decided to refactor it to use the [@for](https://angular.dev/api/core/@for) directive with the [track](https://angular.dev/api/core/@for#track-and-objects-identity) expression instead, which as you can see is much more concise

PR #2536, opened 2026-01-21
feat: Pass oneOf subschema to Translator

daniel-shuy:
When translating [oneOf](https://jsonforms.io/docs/multiple-choice#one-of), pass the subschema to the `Translator` to be consistent with other `Translator` calls so that the `Translator` can access the subschema, e.g.
https://github.com/eclipsesource/jsonforms/blob/6bbfbb331df36343b0346d41ed80854456fff9ff/packages/core/src/i18n/i18nUtil.ts#L126-L131

### DIFF
--- a/packages/core/src/mappers/renderer.ts
+++ b/packages/core/src/mappers/renderer.ts
@@ -349,11 +349,11 @@ export const oneOfToEnumOptionMapper = (
   if (t) {
     // prefer schema keys as they can be more specialized
     if (e.i18n) {
-      label = t(e.i18n, label);
+      label = t(e.i18n, label, { schema: e });
     } else if (fallbackI18nKey) {
-      label = t(`${fallbackI18nKey}.${label}`, label);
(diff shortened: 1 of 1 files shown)


packages/core/src/mappers/renderer.ts
```diff
@@ -349,11 +349,11 @@ export const oneOfToEnumOptionMapper = (
   if (t) {
     // prefer schema keys as they can be more specialized
     if (e.i18n) {
-      label = t(e.i18n, label);
+      label = t(e.i18n, label, e);
```
lucas-koehler: Looking at the control mapper, the schema is passed in an object. I think this would be preferable for consistency reasons. I.e. `{ schema: e }`. What do you think?

daniel-shuy: Ah oops, that was a mistake, fixed it, thanks for reviewing!

lucas-koehler (approved):
Thanks for the update and the contribution. LGTM now 🚀

PR #2540, opened 2026-01-24
Filter error keywords before touch

kchobantonov:
A new configuration option, filterErrorKeywordsBeforeTouch, allows selective suppression of validation errors before a control is touched, based on AJV error keywords.

This option exists because in most real-world forms, only certain errors—most commonly required—should be hidden prior to user interaction. Other validation errors (such as minLength, pattern, or format) often provide useful guidance and should remain visible immediately.

Instead of hiding all validation errors before interaction, you can now explicitly choose which error types are suppressed and which are shown.

This option is only applied when enableFilterErrorsBeforeTouch is enabled.
If enableFilterErrorsBeforeTouch is false, filterErrorKeywordsBeforeTouch has no effect and all validation errors are shown immediately.

If enableFilterErrorsBeforeTouch is true and filterErrorKeywordsBeforeTouch is not provided, the default behavior applies: no validation errors are shown before the control is touched.

Once the control is touched, all validation errors are always displayed, regardless of these settings.

### DIFF
--- a/packages/vue-vuetify/src/util/composition.ts
+++ b/packages/vue-vuetify/src/util/composition.ts
@@ -10,8 +10,11 @@ import {
   defaultJsonFormsI18nState,
   getArrayTranslations,
   getCombinatorTranslations,
+  getCombinedErrorMessage,
   getControlPath,
+  getErrorTranslator,
   getFirstPrimitiveProp,
+  getTranslator,
   isDescriptionHidden,
   type ControlElement,
   type DispatchPropsOfControl,
@@ -132,6 +135,7 @@ export const useVuetifyLabel = <
  */
 export const useVuetifyControl = <
   T extends {
+    schema: JsonSchema;
     uischema: ControlElement;
     path: string;
     config: any;
@@ -198,13 +202,26 @@ export const useVuetifyControl = <
           (error) => input.control.value.path === getControlPath(error),
         ) ?? [];
 
-      const allErrorsFiltered =
-        errorsAtControl.length > 0 &&
-        errorsAtControl.every(
-          (error) => error.keyword && filterKeywords.includes(error.keyword),
-        );
+      // Filter out errors that match the filterKeywords, keep the rest
+      const errorsToShow = errorsAtControl.filter(
+        (error) => !error.keyword || !filterKeywords.includes(error.keyword),
+      );
+      // If no errors were filtered out (all errors remain), return original errors string
+      if (errorsToShow.length === errorsAtControl.length) {
+        return input.control.value.errors;
+      }
 
-      return allErrorsFiltered ? '' : input.control.value.errors;
+      const t = getTranslator()({ jsonforms });
+      const te = getErrorTranslator()({ jsonforms });
+
+      return getCombinedErrorMessage(
+        errorsToShow,
+        te,
+        t,
+        input.control.value.schema,
+        input.control.value.uischema,
+        input.control.value.path,
+      );
     }
 
     // default, all errors are filtered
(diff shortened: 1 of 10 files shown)


packages/vue-vuetify/src/util/composition.ts
```diff
@@ -176,10 +177,38 @@ export const useVuetifyControl = <
...
+      !appliedOptions.value.enableFilterErrorsBeforeTouch
+    ) {
+      return input.control.value.errors;
+    }
+
+    const filterKeywords = appliedOptions.value.filterErrorKeywordsBeforeTouch;
+
+    // Filtering is enabled - check if specific keywords are configured
+    if (Array.isArray(filterKeywords) && filterKeywords.length > 0) {
+      // Granular filtering: only hide specific error keywords
```
sdirix, line 199: I interpreted the comments to filter specific keywords, also the option is named `enableFilterErrorsBeforeTouch`. However the implementation does not filter the specific keyword from the errors but only hides the error if it only contains errors of that type.

So if "Error A" is filtered but the control has both "Error A" and "Error B", then the control will show both "Error A" and "Error B".

We either need to adapt the functionality if this was not intended or rename the option and/or adapt the comments.

packages/vue-vuetify/src/util/composition.ts
```diff
@@ -176,10 +177,38 @@ export const useVuetifyControl = <
     isFocused.value = false;
   };
 
+  const jsonforms = useJsonForms();
```
sdirix, line 184: Nitpick: We are already within JSON Forms state so you could just use `const jsonforms = inject<JsonFormsSubStates>('jsonforms')` directly.

packages/vue-vuetify/src/util/composition.ts
```diff
@@ -176,10 +177,38 @@ export const useVuetifyControl = <
...
-      ? input.control.value.errors
-      : '';
+    // Always show errors if touched, no errors exist, or filtering is not enabled
+    if (
+      touched.value ||
+      !input.control.value.errors ||
+      !appliedOptions.value.enableFilterErrorsBeforeTouch
+    ) {
+      return input.control.value.errors;
+    }
```
sdirix, line 193: Should these changes also be done for array children errors? Currently they continue to show their errors

sdirix (changes requested):
@kchobantonov Thanks for your PR ❤️ 

For the future I would prefer if you could open more granular PRs. One reason some of your PRs take longer to be merged is because they often span multiple topics. For example if this PR would just contain the `location` changes, it could have been merged already.

We prefer smaller PRs, separated by topic, if possible.

packages/vue-vuetify/src/util/composition.ts
kchobantonov: will update the children also - just overlooked that

packages/vue-vuetify/src/util/composition.ts
kchobantonov: I just do not want that inject<JsonFormsSubStates>('jsonforms') scattered around - e.g. the name of the injection key - jsonforms - also it would be good if that was a Symbol to not collide with anything else but since it is already there .... - I have other PRs where I use the useJsonForms instead on the inject because of this -and also moved the useJsonForms in the vue project rather than in the vue-vuetify but that is in another PR

packages/vue-vuetify/src/util/composition.ts
kchobantonov: fixed

packages/vue-vuetify/src/util/composition.ts
kchobantonov: fixed with remapping the error string in case of multiple values where if we can't filter out errors then we reuse already mapped errors string

sdirix (approved):
Works for me, thanks for the changes.

packages/vue-vuetify/src/util/composition.ts
```diff
@@ -176,10 +181,51 @@ export const useVuetifyControl = <
...
+      const te = getErrorTranslator()({ jsonforms });
+
+      return getCombinedErrorMessage(
+        errorsToShow,
+        te,
+        t,
+        input.control.value.schema,
+        input.control.value.uischema,
+        input.control.value.path,
+      );
```
sdirix, line 224: We can't change it in this PR but this is a topic for a potential JSON Forms 4.0. We basically repeat what the core mappers do because we need to customize their behavior. The mappers are too feature rich and should be more modular.

PR #2550, opened 2026-03-02
Angular Material: Remove EnumOption[] type from AutocompleteControlRenderer.options

daniel-shuy:
In #2535, I changed `AutocompleteControlRenderer.options` from `string[]` to `EnumOption[] | string[]`, then mapped the `EnumOption[] | string[]` to `EnumOption[]`, which was the type of the internal `options` prop, using the `stringOptionsToEnumOptions` function.

This allowed users of the `AutocompleteControlRenderer` to pass `EnumOption`s with differing `label` and `value`, which should not be allowed.

### DIFF
--- a/packages/angular-material/src/library/controls/autocomplete.renderer.ts
+++ b/packages/angular-material/src/library/controls/autocomplete.renderer.ts
@@ -110,7 +110,7 @@ export class AutocompleteControlRenderer
   extends JsonFormsControl
   implements OnInit
 {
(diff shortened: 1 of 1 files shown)


daniel-shuy:
Sorry, this should be part of #2551, as users should use `OneOfEnumControlRenderer` instead of `EnumControlRenderer`/`AutocompleteControlRenderer` if they want to have non-matching label/value pairs. I'll close this PR.

PR #2551, opened 2026-03-02
Angular Material: OneOf Enum Control Renderer

daniel-shuy:
Renamed `AutocompleteControlRenderer` to `EnumControlRenderer` to be consistent with other renderer sets
- For backwards compatibility, I've retained the old [component selector](https://angular.dev/guide/components/selectors) and exported the component with the old class name as an alias

The `OneOfEnumControlRenderer` has the same implementation as the `EnumControlRenderer`, the only difference being that it takes in `EnumOption[]` instead of `string[]`, so that a different label can be displayed instead of the value:
https://github.com/eclipsesource/jsonforms/blob/3a9bc7449e96414026b9e7f2572b4d0d54000a89/packages/core/src/mappers/renderer.ts#L320-L323
- `oneOf.const` is mapped to `EnumOption.value`
- `oneOf.title` is mapped to `EnumOption.label`


packages/angular-material/src/library/controls/one-of-enum.renderer.ts
```diff
@@ -0,0 +1,220 @@
...
+      option.label.toLowerCase().startsWith(label.toLowerCase())
+    );
+  }
+  protected getOwnProps(): OwnPropsOfControl & OwnPropsOfEnum {
+    return {
+      ...super.getOwnProps(),
+      options: this.options,
+    };
+  }
+}
```
sdirix: This file is ~95% identical to `EnumControlRenderer`. The only real differences are `mapToProps` and `getOwnProps`.

`OneOfEnumControlRenderer` should extend `EnumControlRenderer` instead of `JsonFormsControl`. Override `mapToProps` and `getOwnProps`, done. The `@Component` decorator still needs its own template, but extract the template string into a shared `const` in `enum.renderer.ts` so both decorators reference the same thing. That eliminates ~100 lines of duplicated class logic and keeps the templates in sync.

sdirix (changes requested):
Thank you for your contribution! ❤️

packages/angular-material/src/library/controls/one-of-enum.renderer.ts
daniel-shuy: I did think of that previously, but decided not to in case their implementation would diverge in the future. But thinking about it again, that is very unlikely to happen.

I've done what you suggested, but changed `EnumControlRenderer` to extend `OneOfEnumControlRenderer` instead, since `EnumControlRenderer` has to map the `string[]` options to `EnumOption[]` either way. I've also extracted the template string and style string into separate `.html` and `.scss` files instead of shared `const`s, so that IDEs/editors can provide proper syntax highlighting, code completion, etc. on them.

packages/angular-material/src/library/module.ts
```diff
@@ -45,9 +45,10 @@ import { MatTabsModule } from '@angular/material/tabs';
 import { MatToolbarModule } from '@angular/material/toolbar';
 import { CUSTOM_ELEMENTS_SCHEMA, NgModule } from '@angular/core';
 import { JsonFormsModule } from '@jsonforms/angular';
-import { AutocompleteControlRenderer } from './controls/autocomplete.renderer';
 import { BooleanControlRenderer } from './controls/boolean.renderer';
 import { DateControlRenderer } from './controls/date.renderer';
+import { EnumControlRenderer } from './controls/enum.renderer';
+import { OneOfEnumControlRenderer } from './controls/enum.renderer';
```
sdirix: Same here, merge into one import. The linter flags this as `import/no-duplicates`.

packages/angular-material/.eslintrc.js
sdirix, line 1: Is there a specific reason you adjust the ESLint rules?

daniel-shuy: Yes, I moved the shared template string of `enum.renderer.ts` to [enum.renderer.html](https://github.com/eclipsesource/jsonforms/pull/2551/changes#diff-98476d446064bbe6bd312cdbcf572b01f46c53a664ef462a01c7e856c5ef1a83) (so that IDEs/editors can provide proper syntax highlighting, code completion, etc. on it). The previous ESLint configuration does not lint Angular templates in `.html` files

daniel-shuy: The change simply scopes the existing ESLint configuration to `.ts` files, and extends `plugin:@angular-eslint/template/recommended` for `.html` files

packages/angular-material/src/library/controls/enum.renderer.ts
```diff
@@ -53,50 +55,9 @@ import { MatInputModule } from '@angular/material/input';
...
-        flex-direction: row;
-      }
-      mat-form-field {
-        flex: 1 1 auto;
-      }
-    `,
-  ],
+  selector: 'OneOfEnumControlRenderer',
+  templateUrl: './enum.renderer.html',
+  styleUrls: ['./enum.renderer.scss'],
```
sdirix: Every other renderer in this package uses inline `template:` and `styles:`. I'd prefer a shared `const` for the template and styles strings that both `@Component` decorators reference, which is what I originally suggested. That keeps the convention consistent and removes the need for the `.eslintrc.js` restructuring.

daniel-shuy: Ok, I've made the changes as you suggested and reverted the changes to `.eslintrc.js`, but note that the shared `const`s will be treated as literal strings by IDEs/editors (no syntax highlighting, code completion, etc)

packages/angular-material/src/library/controls/enum.renderer.ts
```diff
@@ -209,29 +213,51 @@ export class AutocompleteControlRenderer
...
+  // eslint-disable-next-line @angular-eslint/no-input-rename
+  @Input('options')
+  set stringOptions(strOptions: string[]) {
+    this.options = strOptions.map((str) => {
+      return {
+        label: str,
+        value: str,
+      };
+    });
+  }
```
sdirix, line 249: The old `AutocompleteControlRenderer` accepted `EnumOption[] | string[]` via `@Input`. This setter only handles `string[]`. If someone has `<AutocompleteControlRenderer [options]="enumOptionArray">`, the setter will silently produce garbage.

We should remove the old component completely so they get an obvious error or make it backward compatible. If we remove it, we should mention it in the migration guide

sdirix (changes requested):
Looks good, but I saw a minor issue.

packages/angular-material/src/library/controls/enum.renderer.ts
daniel-shuy: Actually, the `@Input` was previously changed by me in a previous PR ([#2535](https://github.com/eclipsesource/jsonforms/pull/2535/changes#diff-cf1c2cafaf106519583c7738e87c99c65bc7595a614a0d49e03a0e93e1d82f84R113)) from `string[]` to `EnumOption[] | string[]` so that users can pass in options with different labels from their values to the `AutocompleteControlRenderer`. I reverted that change since the new `OneOfEnumControlRenderer` should be used for that purpose instead.

Since #2535 is not part of any release, the `@Input` type can be safely reverted back to `string[]`.

sdirix (approved):

PR #2562, opened 2026-03-20
core: improve Translator type for strict TypeScript compatibility

### LINKED ISSUE #2528: Can't implement Translator correctly with TSConfig strict
### Describe the bug

When [strictFunctionTypes](https://www.typescriptlang.org/tsconfig/#strictFunctionTypes) and [strictNullChecks](https://www.typescriptlang.org/tsconfig/#strictNullChecks) are enabled in TSConfig (quite often as nowadays most TS projects have [strict](https://www.typescriptlang.org/tsconfig/#strict) enabled), the [Translator](https://github.com/eclipsesource/jsonforms/blob/6bbfbb331df36343b0346d41ed80854456fff9ff/packages/core/src/store/i18nTypes.ts#L4-L8) interface cannot be implemented correctly.

[strictFunctionTypes](https://www.typescriptlang.org/tsconfig/#strictFunctionTypes) requires function parameters to be contravariant and function return types to be covariant.

[strictNullChecks](https://www.typescriptlang.org/tsconfig/#strictNullChecks) causes `string` and `string | undefined` to be treated as different types.

Because the function implementation must match all 3 overloads:
https://github.com/eclipsesource/jsonforms/blob/6bbfbb331df36343b0346d41ed80854456fff9ff/packages/core/src/store/i18nTypes.ts#L4-L8
- For the `defaultMessage` parameter, the only type that is contravariant with `string`, `undefined` and `string | undefined` is `string | undefined`
- For the return type, the only type that is covariant with `string` and `string | undefined` is `string`

The only implementation that meets both is:
```ts
(id: string, defaultMessage?: string, values?: any) => string;
```
which means that the `defaultMessage` can be `undefined`, but the return value cannot be `undefined`.

This makes it impossible to implement the note in https://jsonforms.io/docs/i18n/#defaultmessage:
> If the `defaultMessage` is `undefined`, you should also return `undefined` if there is no translation for the given key. Returning an empty string (or something similar) instead may result in undesired behavior. JSON Forms will use `undefined` when the message could be skipped or another more generic key could be tried.

### Expected behavior

Able to implement the [Translator](https://github.com/eclipsesource/jsonforms/blob/6bbfbb331df36343b0346d41ed80854456fff9ff/packages/core/src/store/i18nTypes.ts#L4-L8) interface as:
```ts
(id: string, defaultMessage?: string, values?: any) => string | undefined;
```

1. Go to [https://github.com/eclipsesource/jsonforms/blob/v3.7.0/packages/examples/tsconfig.json](packages/examples/tsconfig.json)
2. Set [strictFunctionTypes](https://www.typescriptlang.org/tsconfig/#strictFunctionTypes) and [strictNullChecks](https://www.typescriptlang.org/tsconfig/#strictNullChecks) to `true` in `compilerOptions`
3. Build the project
4. See that [packages/examples/src/examples/i18n.ts](https://github.com/eclipsesource/jsonforms/blob/6bbfbb331df36343b0346d41ed80854456fff9ff/packages/examples/src/examples/i18n.ts#L102-L104) fails to compile
https://github.com/eclipsesource/jsonforms/blob/6bbfbb331df36343b0346d41ed80854456fff9ff/packages/examples/src/examples/i18n.ts#L102-L104

### Screenshots

_No response_

3.7.0

### Package

Core

### Additional context

_No response_

lucas-koehler:
Fix #2528

- Replace overloaded Translator type with a generic conditional type
- The generic type allows inferring the return type from the handed over default message. I.e. a `undefined` defaultMessage results in `string | undefined` while a string default message results in `string`.
- Add createTranslator helper to allow implementation without type assertions under strictFunctionTypes and strictNullChecks.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,45 @@
 # Migration guide
 
+## Migrating to JSON Forms 3.8
+
+### `Translator` type changed from overloaded signatures to a generic conditional type
+
+The `Translator` type was changed to improve compatibility with TypeScript's `strictFunctionTypes` and `strictNullChecks` compiler options (see [#2528](https://github.com/eclipsesource/jsonforms/issues/2528)).
+
+If you were previously assigning a function directly to the `Translator` type, this may no longer compile:
+
+```ts
+// No longer compiles
+const t: Translator = (id, defaultMessage) => defaultMessage ?? id;
+```
+
+Use the new `createTranslator` helper instead:
+
+```ts
+import { createTranslator } from '@jsonforms/core';
+
+const t = createTranslator((id, defaultMessage) => defaultMessage ?? id);
+```
+
+This also replaces the `as Translator` workaround that was previously needed under strict TypeScript settings.
+
+#### Vue: `Translator` return type in Options API
+
+If you have custom Vue renderers that access a `Translator` via `this` (Options API), the return type may no longer narrow to `string` even when a `defaultMessage` is provided.
+This is because Vue's ref unwrapping loses the generic parameter of the new conditional type.
+
+To fix this, use `as string` when you know a default message is always provided:
+
+```ts
+// Before (may now return string | undefined)
+return this.t(label, label);
+
+// After
+return this.t(label, label) as string;
+```
+
+This does not affect the Composition API where `Translator` is accessed directly from a `ComputedRef`.
+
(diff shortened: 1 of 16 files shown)


sdirix (dismissed):
LGTM Thanks

sdirix (approved):

PR #2565, opened 2026-04-07
dev: Update to Node 24 and pnpm 10.33.0 and update GH actions

lucas-koehler:
- Update node to ^24
- Update pnpm corepack version to 10.33.0
- Update @types/node dependency to 24 and update lockfile
- Update GitHub Actions to actions/checkout@v6, actions/setup-node@v6, and pnpm/action-setup v5.0.0
- Remove now obsolete npm update step from publish workflow

This should fix the publish workflow which is currently broken due to https://github.com/nodejs/node/issues/62425


sdirix (approved):
Works for me. Thanks!

PR #2566, opened 2026-04-14
angular: remove deprecated use of ComponentFactoryResolver

lucas-koehler:
- Replaced by using ViewContainerRef.createComponent directly with a component.
- Needed for compatibility with Angular 22 which removes API access to ComponentFactoryResolver


sdirix:
Works for me!

sdirix (approved):

PR #2568, opened 2026-04-16
vue: add jest testing hints on imports

### LINKED ISSUE #2250: Jest tests depend on order of imports for defineComponent
### Describe the bug

We are using CustomRenderers on top of vuetify and test them with jest, vue-test-utils and testing-library. We had very weird errors during test execution:
```
console.warn
    [Vue warn]: Property "controlWrapper" was accessed during render but is not defined on instance. 
...
Cannot read properties of undefined (reading 'control')
```
It turns out that the error happens because of the import order of `defineComponent`. In our custom renderer, we had the following:

```
<script lang="ts">
import { ControlElement } from "@jsonforms/core";
import { rendererProps, RendererProps, useJsonFormsEnumControl } from "@jsonforms/vue";
import { useVuetifyControl, ControlWrapper, DisabledIconFocus } from "@jsonforms/vue-vuetify";
import { defineComponent } from "vue";
```

if this is switched to
```
<script lang="ts">
import { ControlElement } from "@jsonforms/core";
import { defineComponent } from "vue";
import { rendererProps, RendererProps, useJsonFormsEnumControl } from "@jsonforms/vue";
import { useVuetifyControl, ControlWrapper, DisabledIconFocus } from "@jsonforms/vue-vuetify";
```

the tests run properly. Note that the display is completely fine.

### Expected behavior

Either the import order shouldn't matter, or at least that should be written extremely prominently in the documentation.

**Reproduction:**

1. Go to 'https://github.com/Martin-Idel/jsonformsproblem'
2. Clone the repository and execute `yarn install` and then `yarn test`
3. See error
4. Go to the file `SelectFieldRenderer` and switch the order of `import { defineComponent } from "vue";` and `import {
  rendererProps,
  RendererProps,
  useJsonFormsEnumControl,
} from "@jsonforms/vue";`
5. See no error

You can also run `yarn dev` to see that everything seems nice in the browser.

**Most relevant code (see reproduction repo)**

```
<template>
  <control-wrapper
    v-bind="controlWrapper"
    :styles="styles"
    :isFocused="isFocused"
    :appliedOptions="appliedOptions"
  >
    <v-row class="ml-1 mr-1" style="flex-wrap: nowrap">
      <v-select
        :id="control.path.replaceAll('.', '-')"
        :class="styles.control.input"
        :disabled="false"
        :autofocus="appliedOptions.focus"
        :density="'compact'"
        :variant="'underlined'"
        :label="control.label"
        :required="control.required"
        :model-value="control.data"
        :items="useItems"
        :item-title="(item) => item['label']"
        validate-on-blur
        item-value="value"
        v-bind="vuetifyProps('v-select')"
        @update:model-value="onChange"
        @focus="isFocused = true"
        @blur="isFocused = false"
      />
    </v-row>
  </control-wrapper>
</template>

<script lang="ts">
import { ControlElement } from "@jsonforms/core";
import {
  rendererProps,
  RendererProps,
  useJsonFormsEnumControl,
} from "@jsonforms/vue";
import { defineComponent } from "vue";
import { ControlWrapper, DisabledIconFocus } from "@jsonforms/vue-vuetify";
import { useVuetifyControl } from "@jsonforms/vue-vuetify";

const selectFieldRenderer = defineComponent({
  name: "SelectFieldRenderer",
  components: {
    ControlWrapper,
  },
  directives: {
    DisabledIconFocus,
  },
  props: {
  },
    return {
      ...useVuetifyControl(useJsonFormsEnumControl(props), (value) =>
        value !== null ? value : undefined
      ),
    };
  },
  computed: {
    useItems(): unknown[] {
      return (this.control.schema as { enum: string[] }).enum;
    },
  },
});

export default selectFieldRenderer;
</script>
```

and here is the test:
```
import "@testing-library/jest-dom";
import TestSelectFieldRenderer from "../testcomponents/TestSelectFieldRenderer.vue";
import { createVuetify } from "vuetify";
import * as components from "vuetify/components";
import * as directives from "vuetify/directives";
import { render } from "@testing-library/vue";

describe("The SelectFieldRenderer component", () => {
  it("displays selected value", async () => {
    const { getByLabelText, getByText } = await setupComponent("First");

    expect(getByLabelText("This is a title")).toBeInTheDocument();
    expect(getByText("First")).toBeInTheDocument();
  });

  const setupComponent = async (value: "First" | "Second" | undefined) => {
    const vuetify = createVuetify({ components, directives });

    return {
      ...render(TestSelectFieldRenderer, {
        global: {
          plugins: [vuetify],
        },
        props: {
          data: { select: value },
          schema:
            // eslint-disable-next-line max-len
            `{"properties":{"select":{"title":"This is a title","type":"string","enum":["First","Second"]}},"type":"object"}`,
          uiSchema:
            '{"type":"VerticalLayout","elements":[{"type":"SelectField","scope":"#/properties/select"}]}',
        },
      }),
    };
  };
});
```

and the test component:
```
<template>
  <json-forms
    :data="props.data"
    :schema="JSON.parse(schema)"
    :uischema="JSON.parse(uiSchema)"
    :renderers="renderers"
    :validationMode="'ValidateAndShow'"
  />
</template>

<script lang="ts" setup>
import { rankWith, uiTypeIs } from "@jsonforms/core";
import { JsonForms } from "@jsonforms/vue";
import { vuetifyRenderers } from "@jsonforms/vue-vuetify";
import selectFieldRenderer from "@/components/SelectFieldRenderer.vue";

const props = defineProps({
  data: { type: Object, required: true },
  schema: { type: String, required: true },
  uiSchema: { type: String, required: true },
});

const renderers = Object.freeze([
  ...vuetifyRenderers,
  {
    tester: rankWith(15, uiTypeIs("SelectField")),
    renderer: selectFieldRenderer,
  },
]);
</script>

<style scoped />
```

### Screenshots

_No response_

v3.1.0

### Framework

Vue

### RendererSet

### Additional context

We use jsonforms/vue-vuetify renderers in v3.1.0.preview - however I believe the problem resides with jsonforms/vue as that import messes up the test.

lucas-koehler:
- Add documentation on import order for jest (or other) unit tests using CJS transformations
- ~Add explicit `exports` property to vue's package.json to facilitate tools finding the correct bundle~

Closes #2250

### DIFF
--- a/packages/vue/README.md
+++ b/packages/vue/README.md
@@ -289,6 +289,27 @@ const myComponent = defineComponent({
 The injected `jsonforms` object is not meant to be modified directly.
 Instead it should be modified via the provided `dispatch` and by changing the props of the `json-forms` component.
 
+### Testing with Jest / Vitest
+
+When testing custom renderers with Jest or Vitest using CJS transforms, `vue` must be imported **before** `@jsonforms/vue` in your renderer files.
+This is due to the CJS bundle eagerly evaluating Vue component definitions at `require()` time, which can cause issues when Jest's module resolution processes imports sequentially.
+
+```ts
+// Correct - import vue before @jsonforms/vue
+import { defineComponent } from 'vue';
+import { rendererProps, useJsonFormsControl } from '@jsonforms/vue';
+```
+
+```ts
+// May cause errors in tests:
+//   "Property 'controlWrapper' was accessed during render but is not defined on instance"
+import { rendererProps, useJsonFormsControl } from '@jsonforms/vue';
+import { defineComponent } from 'vue';
+```
+
+This only affects test environments using CJS module transforms.
+Browser builds using Webpack, Vite, or other ESM-aware bundlers are not affected.
+
 ## License
 
 The JSON Forms project is licensed under the MIT License. See the [LICENSE file](https://github.com/eclipsesource/jsonforms/blob/master/LICENSE) for more information.
--- a/packages/vue/README.md
+++ b/packages/vue/README.md
@@ -291,8 +291,7 @@ Instead it should be modified via the provided `dispatch` and by changing the pr
 
 ### Testing with Jest / Vitest
 
-When testing custom renderers with Jest or Vitest using CJS transforms, `vue` must be imported **before** `@jsonforms/vue` in your renderer files.
-This is due to the CJS bundle eagerly evaluating Vue component definitions at `require()` time, which can cause issues when Jest's module resolution processes imports sequentially.
+When testing custom renderers in a CJS-transformed test environment (Jest, or Vitest configured with CJS transforms), `vue` must be imported **before** `@jsonforms/vue` in your renderer files:
 
 ```ts
 // Correct - import vue before @jsonforms/vue
@@ -301,14 +300,15 @@ import { rendererProps, useJsonFormsControl } from '@jsonforms/vue';
 ```
 
 ```ts
-// May cause errors in tests:
-//   "Property 'controlWrapper' was accessed during render but is not defined on instance"
+// May produce errors such as:
+//   "Property '<name>' was accessed during render but is not defined on instance"
 import { rendererProps, useJsonFormsControl } from '@jsonforms/vue';
 import { defineComponent } from 'vue';
 ```
 
-This only affects test environments using CJS module transforms.
-Browser builds using Webpack, Vite, or other ESM-aware bundlers are not affected.
+The reason is that several components shipped by `@jsonforms/vue` (e.g. `JsonForms`, `DispatchRenderer`) call `defineComponent` at module load. When the package is consumed via `require()` and the test runner's CJS transform doesn't hoist imports, `vue` must already be evaluated at that point. Thus, importing `vue` first in renderer and test files is the safest default.
+
+Browser/dev builds using Webpack, Vite, or other ESM-aware bundlers are not affected.
 
 ## License
 
--- a/packages/vue-vanilla/README.md
+++ b/packages/vue-vanilla/README.md
@@ -194,6 +194,10 @@ Attributes not specified here fall back to either the `defaultStyles` or provide
 }
 ```
 
+### Testing in CJS-transformed environments
+
+When writing tests for custom renderers in a Jest/Vitest CJS-transformed environment, `vue` should be imported first in your custom renderers. See [Testing with Jest / Vitest](../vue/README.md#testing-with-jest--vitest) in the `@jsonforms/vue` README.
+
 ## License
 
 The JSONForms project is licensed under the MIT License. See the [LICENSE file](https://github.com/eclipsesource/jsonforms/blob/master/LICENSE) for more information.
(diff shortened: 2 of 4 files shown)


lucas-koehler:
The "exports" seems to have caused test issues on windows and mac :sweat_smile: 

**Reason:**
The exports field is strict and relies on accurate realpath() resolution to determine package identity and resolve relative entry points. In a pnpm workspace monorepo, where packages are symlinked (or junction-linked on Windows), this resolution has subtle cross-platform differences that can cause the exports map to either fail entirely or resolve to the wrong bundle format — while the more lenient main/module fallback works reliably on all platforms.                                                                                            
                                                                             
This is a known category of issue in the Node.js ecosystem — exports + workspace symlinks + cross-platform is a notorious combination. Removing the exports field was the pragmatic fix.

packages/vue/README.md
```diff
@@ -289,6 +289,27 @@ const myComponent = defineComponent({
...
+
+```ts
+// Correct - import vue before @jsonforms/vue
+import { defineComponent } from 'vue';
+import { rendererProps, useJsonFormsControl } from '@jsonforms/vue';
+```
+
+```ts
+// May cause errors in tests:
+//   "Property 'controlWrapper' was accessed during render but is not defined on instance"
```
sdirix: The quoted error references `controlWrapper`, which comes from the Vuetify bindings. The generic `Vue` README does not seem like the correct place for this.

packages/vue/README.md
```diff
@@ -289,6 +289,27 @@ const myComponent = defineComponent({
 The injected `jsonforms` object is not meant to be modified directly.
 Instead it should be modified via the provided `dispatch` and by changing the props of the `json-forms` component.
 
+### Testing with Jest / Vitest
+
+When testing custom renderers with Jest or Vitest using CJS transforms, `vue` must be imported **before** `@jsonforms/vue` in your renderer files.
+This is due to the CJS bundle eagerly evaluating Vue component definitions at `require()` time, which can cause issues when Jest's module resolution processes imports sequentially.
```
sdirix: This is a bit generic. The cause seems to be that we call `defineComponent` at module load so `vue` needs to be initialized first. Do other component libraries have the same issue?

packages/vue/README.md
```diff
@@ -289,6 +289,27 @@ const myComponent = defineComponent({
 The injected `jsonforms` object is not meant to be modified directly.
 Instead it should be modified via the provided `dispatch` and by changing the props of the `json-forms` component.
 
+### Testing with Jest / Vitest
```
sdirix: The PR description says `Closes #2250`, but this only documents a workaround; the underlying bug (import order being load-bearing) still exists. The original commit message used "Mitigates" which seems more accurate. Either keep the issue open after merge with a doc-note, or change the PR body.

sdirix (changes requested):

lucas-koehler:
@sdirix I made the wording a bit more precise and also added a note to the vue-vuetify and vue-vanilla readmes referencing the vue readme.
IMO this should be sufficient for now and we can add more information later if necessary.

sdirix (approved):
Thank you!

PR #2573, opened 2026-05-18
angular: build against v20, allow v22 RC, inject() instead of constructor injection

lucas-koehler:
Further changes:
- use `inject()` instead of constructor injection: constructor injections are discouraged by angular's lint rules starting with Angular 20
- Upgrade to typescript ~5.8.3 to be compatible with Angular 20 including a small  fix in the vue-vuetify package to keep it compatible with 5.8.3
- Set minimum dependency age of 72h to reduce likelyhood of supply chain attacks when updating dependencies

I tested this locally with the seed building against 22 RC.0 and it worked fine :) I  needed one small code change resulting from an unrelated update of the Angular enum renderer. In addition, constructor injection of custom renderers needed to be replaced as described in the migration doc.
The examples app worked fine for me, too which builds against Angular 20 now, too


MIGRATION.md
```diff
@@ -40,6 +40,41 @@ return this.t(label, label) as string;
 
 This does not affect the Composition API where `Translator` is accessed directly from a `ComputedRef`.
 
+### Angular renderers use `inject()` instead of constructor injection
```
sdirix, line 49: Works for me. We should just mention that we also migrated the Angular version to 20

sdirix: @lucas-koehler I added a notice myself

lucas-koehler: Oh yea good catch and thanks :)

sdirix (dismissed):
Works for me! Thanks!

sdirix (approved):

PR #2581, opened 2026-05-21
react: remove usage of React.ReducerAction

### LINKED ISSUE #2572: React 19 has no exported member 'ReducerAction'
### Describe the bug

In @jsonforms/react/src/JsonFormsContext.tsx, there is import `ReducerAction`, it works in React 18.3, but does not work in 19, because there is no such type (@types/react/index.d.ts)

### Expected behavior

Project build

1. Set up project with react 19
2. Add
`"@jsonforms/core": "^3.7.0",
    "@jsonforms/react": "^3.7.0",`
3. Try to build the project
4. See error TS2305: Module '"react"' has no exported member 'ReducerAction'.

### Screenshots

Image

3.5, 3.7

### Package

React Bindings

### Additional context

_No response_

sdirix:
React 19 removed the ReducerAction type and therefore errors. We now use CoreActions directly.

Fixes #2572

Note that when using VS Code Typescript will complain in the touched files because VS Code already uses TS v6+. You need to configure VS Code to use the workspace version of Typescript to get rid of the errors.


lucas-koehler (approved):
LGTM! Thanks for the fix 🚀

PR #2583, opened 2026-05-22
angular: add option to toggle off errors without user interaction

### LINKED ISSUE #2458: Angular - Formcontrol is always touched
### Describe the bug

Having any field inside json forms even if it's from angular material renderers or custom component renderers they are always touched, even if the user did not do any action. This can be seen from the: `form.touched` property that comes from the extended JsonFormsControl class 

### Expected behavior

The touched property should initially be false when the user did not do any action on the component and turn to true when the user does any action on it

1. Checkout and start the angular seed: https://github.com/eclipsesource/jsonforms-angular-seed
2. On the template of: src/app/custom.autocomplete.ts add the: {{ form.touched }} to verify

### Screenshots

Image

Image

v3.5.1

### Package

Angular Bindings

### Additional context

_No response_

pedroestabruxelles:
By default the library is marking all fields as touched, to avoid creating regressions I added an option to toggle this behavior off.

Added as a component option or as a global configuration. 
Since everything is being treated as "any" which is not really a good convention for this component I added an interface with the new option, where for now will be only available on Angular  (feel free to move to the core if someone wants to implement on the other frameworks).

Fixes #2458


packages/angular/src/library/abstract-control.ts
```diff
@@ -176,7 +177,11 @@ export abstract class JsonFormsAbstractControl<
   protected triggerValidation() {
     // these cause the correct update of the error underline, seems to be
     // related to ionic-team/ionic#11640
-    this.form.markAsTouched();
+    const config = this.jsonFormsService.getConfig();
+    const appliedUiSchemaOptions = merge({}, config, this.uischema?.options);
+    if (appliedUiSchemaOptions.showErrorsImmediately) {
```
sdirix: This breaks several angular-material tests (`Number/Date/Enum/OneOfEnum/Text control should display errors`). Those tests construct `JsonFormsAngularService` directly (e.g. [date-control.spec.ts](https://github.com/eclipsesource/jsonforms/blob/53918bc8f9536bed3ded81f803b89e9be4137ae1/packages/angular-material/test/date-control.spec.ts#L287-L308)), so the merge that injects `angularConfigDefault` in `JsonForms` never runs and `showErrorsImmediately` ends up undefined here. The default needs to be applied where `setConfig` is dispatched in the [service init](https://github.com/eclipsesource/jsonforms/blob/53918bc8f9536bed3ded81f803b89e9be4137ae1/packages/angular/src/library/jsonforms.service.ts#L77-L80).

By swapping the default we don't need the initial config at all.

packages/angular/src/library/index.ts
```diff
@@ -22,12 +22,16 @@
...
+export * from './array-control';
 export * from './base.renderer';
 export * from './control';
-export * from './array-control';
+export * from './jsonforms-root.component';
 export * from './jsonforms.component';
+export {
+  JsonFormsAngularConfig,
+  provideJsonFormsConfig,
+} from './jsonforms.config';
```
sdirix: We usually use `export * from './jsonforms.config';`

sdirix:
Thanks for the PR. However we can only accept it in case you sign the CLA

pedroestabruxelles:
Already did not sure it's still saying I haven't. I've give another try.

packages/angular/src/library/jsonforms.config.ts
```diff
@@ -0,0 +1,62 @@
...
+  THE SOFTWARE.
+*/
+import { InjectionToken, Provider } from '@angular/core';
+
+export interface JsonFormsAngularConfig {
+  /**
+   * When true (default), validation error indicators are shown immediately without
+   * requiring the user to focus or interact with the field first
+   */
+  showErrorsImmediately?: boolean;
```
sdirix: I think it would be better to name this differently and swap the default, i.e. `showErrorsOnTouch` as an opt-in feature and then have the default be `false`. As `undefined` is also falsy we don't need a specific default config containing a `true`.

pedroestabruxelles:
"You have agreed to the CLA for multiple repositories or organizations"
The form is blocked so not sure what am I supposed to do additionally?
@sdirix

sdirix:
Looks like the commit author points to a different Github account than the one you signed the CLA with, i.e. `pedro-fernandes-bxl` vs `pedroestabruxelles`

sdirix (changes requested):
Thanks for the contribution. I had a look over the code, please see my comments.

pedroestabruxelles:
> Looks like the commit author points to a different Github account than the one you signed the CLA with, i.e. `pedro-fernandes-bxl` vs `pedroestabruxelles`

Oh snap, forgot about that. I guess I will have to redo the whole thing on my personal laptop.

sdirix:
You could amend the existing commit and force push to switch the author

packages/angular/src/library/index.ts
pedroestabruxelles: this was done on purpose to make public only the parts that are needed to be public

packages/angular/src/library/jsonforms.config.ts
pedroestabruxelles: If it's not true by default, it would cause a regression. 
If you want me to change the default behavior to not trigger the touch event on initialization let me know.

sdirix: `showErrorsOnTouch` is formulated differently so it would be `false` by default.

packages/angular/src/library/index.ts
sdirix: We usually export everything because it's hard to foresee what adopters want to access

packages/angular/src/library/abstract-control.ts
pedroestabruxelles: I'm not sure if I understand. I've ran the tests and all passed.
And the components seem to work with or without configuration.

packages/angular/src/library/jsonforms.config.ts
```diff
@@ -0,0 +1,62 @@
...
+  THE SOFTWARE.
+*/
+import { InjectionToken, Provider } from '@angular/core';
+
+export interface JsonFormsAngularConfig {
+  /**
+   * When true, validation error indicators are shown immediately without
+   * requiring the user to focus or interact with the field first
+   */
+  showErrorsOnTouch?: boolean;
```
sdirix, line 31: The doc comment is inverted. When `showErrorsOnTouch` is `true`, `markAsTouched()` is skipped, so errors only appear after the user actually interacts with the field. The comment describes the opposite behavior.

packages/angular/src/library/jsonforms.config.ts
```diff
@@ -0,0 +1,62 @@
...
+  [key: string]: unknown;
+}
+
+export const angularConfigDefault: JsonFormsAngularConfig = {
+  showErrorsOnTouch: false,
+};
```
sdirix: `angularConfigDefault` is redundant: `showErrorsOnTouch: false` is equivalent to `undefined` in the `!appliedUiSchemaOptions.showErrorsOnTouch` check. Dropping the const (and the merges referencing it in `jsonforms-root.component.ts`) would simplify the wiring and we don't need a framework-specific default at all.

packages/angular-material/example/app/app.component.ts
```diff
@@ -121,4 +127,11 @@ export class AppComponent {
   toggleReadonly() {
     this.readonly = !this.readonly;
   }
+
+  toggleShowErrorsImmediately() {
```
sdirix: Leftover naming from the previous iteration. Rename to `toggleShowErrorsOnTouch` to match the config flag.

sdirix (changes requested):
Thanks for the changes. Please have a look at my comments.

packages/examples/src/examples/show-errors-on-touch.ts
```diff
@@ -0,0 +1,125 @@
...
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+import { registerExamples } from '../register';
+
+export const schema = {
```
sdirix: Is this example needed? There is nothing "show-errors-on-touch" related in here? So we can just skip all the examples changes?

packages/angular/src/library/jsonforms.config.ts
pedroestabruxelles: removed

packages/angular-material/example/app/app.component.ts
pedroestabruxelles: done

packages/examples/src/examples/show-errors-on-touch.ts
pedroestabruxelles: Removed, it was to test with all component as required.

packages/angular/src/library/jsonforms.config.ts
```diff
@@ -0,0 +1,57 @@
...
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+import { InjectionToken, Provider } from '@angular/core';
+
+export interface JsonFormsAngularConfig {
+  /**
+   * When true, markAsTouched() is skipped, so errors only appear after the user actually interacts with the field.
```
sdirix:
```suggestion
   * When true, validation errors only appear after the user has interacted with the field.
```

sdirix (dismissed):

sdirix (approved):

sdirix:

PR #2584, opened 2026-05-28
core: improve id generation

sdirix:
- Makes id generation overridable
- Improves the worst case in current id generation by keeping tab on the next open count
- Keeps current visible behavior the same as before

Related to https://github.com/eclipsesource/jsonforms/pull/2383

### DIFF
--- a/packages/angular/src/library/abstract-control.ts
+++ b/packages/angular/src/library/abstract-control.ts
@@ -33,10 +33,10 @@ import {
   Actions,
   computeLabel,
   ControlElement,
+  Id,
   JsonFormsState,
   JsonSchema,
   OwnPropsOfControl,
-  removeId,
   StatePropsOfControl,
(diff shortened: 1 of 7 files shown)


lucas-koehler (approved):
LGTM thanks!

There might be an edge case where freed ids are not reused:
When an id is freed that had collided with a different prefix's id, the freed low suffix may not be reused, so a later id for that prefix can get a higher number. This can't happen with the JSON-pointer scopes JSON Forms generates ids from.

AS this should not happen and does not violate the expectation of unique ids, there is no need to complicate the current code.

lucas-koehler:
@sdirix For transparency: I rebased the commits onto master and changed the wording very slightly replacing `chore` with `core`: IMO these changes are changes to the core package and not a simple chore. I'll update the PR title accordingly

PR #2585, opened 2026-05-28
fix: refactor update data to avoid lodash issues

### LINKED ISSUE #2102: Field with brackets in property name
### Describe the bug

When you add a property with brackets in the property name ("test[0]") in your schema and UI schema, you can't type or enter any value in the corresponding input field. However, when you pass some value for this field as data, you can type a value in the input field.

### Expected behavior

I can enter a value in an input field for a property that has brackets in its name.

1. Clone the jsonforms angular seed from https://github.com/eclipsesource/jsonforms-angular-seed
2. Replace contents of  "src\assets\schema.json" with
```
{
  "type": "object",
  "properties": {
    "test[0]": { "type": "string" }
  }
}
```
3. Replace contents of "src\assets\uischema.json" with
```
{
  "type": "Control",
  "scope": "#/properties/test[0]"
}
```
4. Run the jsonforms angular seed
5. Go to localhost:4200
6. Try to type something in the field "Test 0", this does not work
7. Replace the contents of "src\app\data.ts" with
```
export default {
  "test[0]": "TEST",
};
```
8. Go to localhost:4200
9. The field "Test 0" will now contain the value "TEST" and you can now change this value

### Screenshots

_No response_

Google Chrome 110.0.5481.178

v3.0.0

### Framework

Angular

### RendererSet

Material

### Additional context

_No response_

### LINKED ISSUE #2397: Update data bug on nested property with numeric key
### Describe the bug

Updating data for nested property with numeric key is working kinda funny 😅 
For that kind of schema:
```
{
  "type": "object",
  "properties": {
    "group-key": {
      "type": "object",
      "properties": {
        "15": {
          "type": "string"
        }
      }
    }
  },
}
```

updating data for property with "15" key will result in assigning to "group-key" an array of 16 elements where first 15 are `null` and at index 15 you will have value that you entered.
Something like that:

```
{
  "group-key": [
    null, null, null, null, null,
    null, null, null, null, null,
    null, null, null, null, null,
    "something"
  ]
}
```

### Expected behavior

Expected `data` would be:

```
{
  "group-key": {
    "15": "something"
  }
}
```

1. Create schema with nested numeric key: 
```
{
  "type": "object",
  "properties": {
    "group-key": {
      "type": "object",
      "properties": {
        "15": {
          "type": "string"
        }
      }
    }
  },
}
```
2. Create UiSchema:
```
{
  "type": "VerticalLayout",
  "elements": [
    {
      "type": "Group",
      "label": "Some Group",
      "elements": [
        {
          "type": "Control",
          "label": "Label",
          "scope": "#/properties/group-key/properties/15"
        }
      ]
    }
  ]
}
```
3. Enter some text in nested input
4. See `data` of form: 
```
{
  "group": [
    null, null, null, null, null,
    null, null, null, null, null,
    null, null, null, null, null,
    "something"
  ]
}
```

### Screenshots

_No response_

v3.2.1

### Package

React Material Renderers

### Additional context

```
{
  "type": "object",
  "properties": {
    "group": {
      "type": "object",
      "properties": {
        "15": {
          "type": "string"
        },
        "non-numeric-key": {
          "type": "string"
        }
      }
    }
  },
  "required": []
}
```

if there are two nested field and firstly you will enter value for "non-numeric-key" then everything will work fine.
It will look like this:
```
{
  "group-key": {
    "non-numeric-key": "something"
  }
}
```
then numeric key will be added correctly:
```
{
  "group-key": {
    "15": "hello there",
    "non-numeric-key": "something"
  }
}
```

But the other way will result in array with `null` elements

sdirix:
Replace the lodash/fp/set call in UPDATE_DATA with a dedicated helper. This fixes issues with numeric segments (e.g. "group-key.15") and bracket characters in property names (e.g. "test[0]").

Fixes #2102
Fixes #2397

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -40,6 +40,16 @@ return this.t(label, label) as string;
 
 This does not affect the Composition API where `Translator` is accessed directly from a `ComputedRef`.
 
+### Data update paths treat all segments literally
+
+Data updates (e.g. dispatched `update` actions) previously wrote to the form data via lodash's `set`/`unset`, which interpret bracket notation and array indices in paths.
+This corrupted data for property names that look like lodash path syntax, for example numeric property names like `"15"` were turned into array indices and names containing brackets like `"prop[0]"` were split up (see [#2397](https://github.com/eclipsesource/jsonforms/issues/2397) and [#2102](https://github.com/eclipsesource/jsonforms/issues/2102)).
+
+Updates now use the new `setDataAt`/`unsetDataAt` utilities of `@jsonforms/core`, which split paths on `.` and treat every segment as a literal property name, matching how JSON Forms resolves values for display.
+When a missing intermediate container is created, the JSON Schema decides whether it becomes an array or an object; without schema type information, a numeric follow-up segment creates an array, as before.
+
+If you dispatch update actions yourself, make sure to use dot-separated paths (e.g. `update('list.0.name', ...)`) instead of lodash bracket syntax (e.g. `update('list[0].name', ...)`), which is no longer interpreted.
+
 ### Angular support now targets Angular 20 to 22
 
 When using JSON Forms 3.8, your Angular application now needs to target Angular 20, 21 or 22.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,17 @@
 # Migration guide
 
+## Migration to JSON Forms 3.9
+
+### Data update paths treat all segments literally
+
+Data updates (e.g. dispatched `update` actions) previously wrote to the form data via lodash's `set`/`unset`, which interpret bracket notation and array indices in paths.
+This corrupted data for property names that look like lodash path syntax, for example numeric property names like `"15"` were turned into array indices and names containing brackets like `"prop[0]"` were split up (see [#2397](https://github.com/eclipsesource/jsonforms/issues/2397) and [#2102](https://github.com/eclipsesource/jsonforms/issues/2102)).
+
+Updates now use the new `setDataAt`/`unsetDataAt` utilities of `@jsonforms/core`, which split paths on `.` and treat every segment as a literal property name, matching how JSON Forms resolves values for display.
+When a missing intermediate container is created, the JSON Schema decides whether it becomes an array or an object; without schema type information, a numeric follow-up segment creates an array, as before.
+
+If you dispatch update actions yourself, make sure to use dot-separated paths (e.g. `update('list.0.name', ...)`) instead of lodash bracket syntax (e.g. `update('list[0].name', ...)`), which is no longer interpreted.
+
 ## Migrating to JSON Forms 3.8
 
 ### `Translator` type changed from overloaded signatures to a generic conditional type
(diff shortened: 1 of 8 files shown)


packages/core/src/util/setData.ts
```diff
@@ -0,0 +1,187 @@
...
+export const setDataAt = (
+  data: any,
+  path: string,
+  value: any,
+  rootSchema?: JsonSchema
+): any => {
+  const segments = splitPath(path);
+  if (segments.length === 0) {
+    return value;
+  }
```
lucas-koehler: `''.split('.')` returns `['']`, never `[]`, so this guard is unreachable. An empty path falls through to `assign(container, '', value)`, which writes to key `''`.That happens to match the old lodash behavior (`setFp('', v, {b:2}) → {b:2, '': v}`), so it's not a regression, but the guard is misleading.

Should we drop it, or if the intent was to treat an empty path as "replace root", handle that explicitly?

packages/core/src/util/setData.ts
```diff
@@ -0,0 +1,187 @@
...
+ * Immutably unsets the value at the dotted `path` within `data`.
+ *
+ * Numeric path segments and bracket notation in segments are treated as
+ * plain object property names, mirroring the semantics of {@link setDataAt}.
+ */
+export const unsetDataAt = (data: any, path: string): any => {
+  const segments = splitPath(path);
+  if (segments.length === 0) {
+    return data;
+  }
```
lucas-koehler: Same as https://github.com/eclipsesource/jsonforms/pull/2585/changes#r3334312514

lucas-koehler (commented):
Looks pretty good to me already 🚀 
I have one minor concern inline. please have a look :)

packages/core/src/util/setData.ts
```diff
@@ -0,0 +1,224 @@
...
+  return {};
+};
+
+/**
+ * Looks up the own property `segment` of `data`, ignoring inherited
+ * properties like `__proto__`, mirroring the semantics of `resolveData`.
+ */
+const ownPropertyValue = (data: any, segment: string): any =>
+  data != null && Object.prototype.hasOwnProperty.call(data, segment)
```
lucas-koehler: Is it on purpose to use `!= null` here? It seems "weird" to use the unstrict operator instead of either being explicit with multiple `!==` cases or using `!data` to simply check its truthy

MIGRATION.md
```diff
@@ -40,6 +40,16 @@ return this.t(label, label) as string;
...
+
+Data updates (e.g. dispatched `update` actions) previously wrote to the form data via lodash's `set`/`unset`, which interpret bracket notation and array indices in paths.
+This corrupted data for property names that look like lodash path syntax, for example numeric property names like `"15"` were turned into array indices and names containing brackets like `"prop[0]"` were split up (see [#2397](https://github.com/eclipsesource/jsonforms/issues/2397) and [#2102](https://github.com/eclipsesource/jsonforms/issues/2102)).
+
+Updates now use the new `setDataAt`/`unsetDataAt` utilities of `@jsonforms/core`, which split paths on `.` and treat every segment as a literal property name, matching how JSON Forms resolves values for display.
+When a missing intermediate container is created, the JSON Schema decides whether it becomes an array or an object; without schema type information, a numeric follow-up segment creates an array, as before.
+
+If you dispatch update actions yourself, make sure to use dot-separated paths (e.g. `update('list.0.name', ...)`) instead of lodash bracket syntax (e.g. `update('list[0].name', ...)`), which is no longer interpreted.
+
```
lucas-koehler: As 3.8 was released by now, this needs to be moved to a new section `## Migrating to JSON Forms 3.9`

lucas-koehler (changes requested):
Thanks for the update and advanced edge case handling 🚀 
The changes already looks pretty good to me. I only have one question and one needed updated for the migration guide. Please have a look :)

lucas-koehler (approved):

PR #2586, opened 2026-05-28
core: handle nullable object/array in generateUISchema

### LINKED ISSUE #2411: Adding ‘null’ and ‘object’ type to property values for a schema
### Describe the bug

Hi,

I am trying to set a ‘type: [‘null’, ‘object’]’ for a schema item with the below code as an example:

```
schema={{
     type: 'object',
     properties: {
          name: {
               type: 'string',
               description: 'Name example'
          },
          geometry: {
               type: ['null', 'object'],
               properties: {
                    shape: {
                         type: 'string'
                         description: 'Example geometry shape property'
                    },
                    etc........
               }
          }
     }
}}
```

When I put this schema into my code it kills my app. Adding in type: 'object' it works fine.

Is it allowed to have type: ['null', 'object'] as I have done in the geometry section above?

### Expected behavior

both types null and object will be applied to a properties property.

add 

```
{
     type: 'object',
     properties: {
          name: {
               type: 'string',
               description: 'Name example'
          },
          geometry: {
               type: ['null', 'object'],
               properties: {
                    shape: {
                         type: 'string'
                         description: 'Example geometry shape property'
                    },
                    etc........
               }
          }
     }
}
```

into the schema option of JSON forms

### Screenshots

_No response_

3.4.1

### Package

_No response_

### Additional context

_No response_

sdirix:
Treat ['null', T] type arrays as nullable T for UI generation so that nested nullable objects traverse their properties instead of producing a bare Control. This avoids infinite recursion in object renderers that regenerate the UI schema for nested nullable objects.

Fixes #2411

### DIFF
--- a/packages/core/src/generators/uischema.ts
+++ b/packages/core/src/generators/uischema.ts
@@ -141,13 +141,21 @@ const generateUISchema = (
     return null;
   }
 
-  if (types.length > 1) {
+  // If the schema declares multiple types alongside 'null' (e.g. ['null', 'object']),
+  // treat it as a nullable instance of the non-null type for UI generation purposes.
+  // Without this, a nested nullable object would only produce a single Control,
+  // which can lead to infinite recursion when an object renderer regenerates the
+  // UI schema for that nested schema (see issue #2411).
+  const nonNullTypes = types.filter((t) => t !== 'null');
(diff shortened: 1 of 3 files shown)


lucas-koehler (approved):
LGTM, thanks!

PR #2588, opened 2026-06-01
Upgrade mui packages to v9 (material-renderers)

AirBorne04:
I have upgraded the @mui/* packages to the latest ^9 version and adjusted for changes of component api and tests to properly pass.

There are two components where I would require feedback:
1. changes to components where the api changed more than just prop names: 
a. packages/material-renderers/src/additional/ListWithDetailMasterItem.tsx
b. packages/material-renderers/src/layouts/ExpandPanelRenderer.tsx

Please let me know, if there is anything else I should adjust,
Daniel


lucas-koehler:
Hi @AirBorne04 , thanks for the contribution :heart: 
There are some lint errors reported by this build: https://github.com/eclipsesource/jsonforms/actions/runs/26755183304/job/80061183345?pr=2588

Regarding the ExpandPanelRenderer: The `Grid` changes look reasonable. I just wonder whether we really need the `component='div'` for the IconButtons. Is this recommended when migrating to MUI 9 and/or is there another reason we need to specify this explicitly?

Regarding the:
The master item itself containing the Delete button looks good in the examples. Thus, I think the change is fine to remove the deprecated `ListItemSecondaryAction`.
However, the detail itself seems to be broken now (e.g. the Customer group should be below the 2. The layout does not work as expected. I assume this is caused by another component though
image

Compare to how it looks on master:
image

AirBorne04:
@lucas-koehler thank you for taking a look at the PR, i have fixed the lint errors.

regarding the `component='div'`, the reason is that React was complaining about `Button` elements nested into `Button` elements and that is illegal html nesting. I was not sure though which higher level element actually was the `Button` and did not yet investigate that.

i will take a look at the layout in the grid, as soon as i get a chance.

lucas-koehler:
@AirBorne04 For our planning, do you plan to continue working on this?

AirBorne04:
@lucas-koehler i am currently busy with other things, so i am not sure when i would get to it again. I have seen the other PR ongoing though.

lucas-koehler:
Hi @AirBorne04 , no worries. Thanks for contributing anyway :heart: 
I'll close this PR now because the other one is close to merging :)

PR #2596, opened 2026-06-16
feat(renderers): add mixed and additional property editors

kchobantonov:
Add MixedRenderer and AdditionalProperties support across Material renderers.

- add React Material mixed/additional property renderers with tree navigation
- add Angular Material mixed/additional property renderers with Material tree support
- update Vue Vuetify mixed/additional property UX with v-treeview and Vuetify-styled split panes
- support nested object/array navigation, rename/delete actions, and dynamic property editing
- add shared Vuetify splitpane components and renderer icon mappings

### DIFF
--- a/packages/angular-material/src/library/other/additional-properties.renderer.ts
+++ b/packages/angular-material/src/library/other/additional-properties.renderer.ts
@@ -1,0 +1,698 @@
+/*
+  The MIT License
+
+  Copyright (c) 2017-2026 EclipseSource Munich
+  https://github.com/eclipsesource/jsonforms
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
+  furnished to do so, subject to the following conditions:
+
+  The above copyright notice and this permission notice shall be included in
+  all copies or substantial portions of the Software.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+*/
+import { CommonModule } from '@angular/common';
+import {
+  ChangeDetectionStrategy,
+  Component,
+  ViewEncapsulation,
+  Input,
+  OnChanges,
+  SimpleChanges,
+  inject,
+} from '@angular/core';
+import { FormsModule } from '@angular/forms';
+import { MatButtonModule } from '@angular/material/button';
+import { MatFormFieldModule } from '@angular/material/form-field';
+import { MatIconModule } from '@angular/material/icon';
+import { MatInputModule } from '@angular/material/input';
+import { MatMenuModule } from '@angular/material/menu';
+import { MatTooltipModule } from '@angular/material/tooltip';
+import { JsonFormsModule, JsonFormsAngularService } from '@jsonforms/angular';
+import {
+  Actions,
+  composePaths,
+  ControlElement,
+  createControlElement,
+  createDefaultValue,
+  Generate,
+  GroupLayout,
+  JsonFormsCellRendererRegistryEntry,
+  JsonFormsRendererRegistryEntry,
+  JsonSchema,
+  JsonSchema7,
+  Resolve,
+  UISchemaElement,
+} from '@jsonforms/core';
+import startCase from 'lodash/startCase';
+
+interface AdditionalPropertyItem {
+  propertyName: string;
+  path: string;
+  schema: JsonSchema;
+  uischema: UISchemaElement;
+}
+
+const ANY_TYPE: JsonSchema7['type'] = [
+  'array',
+  'boolean',
+  'integer',
+  'null',
+  'number',
+  'object',
+  'string',
+];
+
+const toObjectSchema = (schema: JsonSchema): JsonSchema7 =>
+  typeof schema === 'object' ? (schema as JsonSchema7) : {};
+
+const hasAdditionalProperties = (schema: JsonSchema): boolean => {
+  const objectSchema = toObjectSchema(schema);
+  return (
+    Boolean(
+      objectSchema.patternProperties &&
+        Object.keys(objectSchema.patternProperties).length > 0
+    ) ||
+    typeof objectSchema.additionalProperties === 'object' ||
+    objectSchema.additionalProperties === true
+  );
+};
+
+const getMatchingAdditionalPropertySchema = (
+  propName: string,
+  parentSchema: JsonSchema,
+  rootSchema: JsonSchema
+): JsonSchema => {
+  const objectSchema = toObjectSchema(parentSchema);
+  let propSchema: JsonSchema | undefined;
+
+  if (objectSchema.patternProperties) {
+    const matchedPattern = Object.keys(objectSchema.patternProperties).find(
+      (pattern) => new RegExp(pattern).test(propName)
+    );
+    if (matchedPattern) {
+      propSchema = objectSchema.patternProperties[matchedPattern];
+    }
+  }
+
+  if (
+    (!propSchema && typeof objectSchema.additionalProperties === 'object') ||
+    objectSchema.additionalProperties === true
+  ) {
+    propSchema =
+      objectSchema.additionalProperties === true
+        ? { additionalProperties: true }
+        : objectSchema.additionalProperties;
+  }
+
+  if (typeof propSchema === 'object' && typeof propSchema.$ref === 'string') {
+    propSchema = Resolve.schema(rootSchema, propSchema.$ref, rootSchema);
+  }
+
+  propSchema = propSchema ?? {};
+
+  if (typeof propSchema === 'object' && propSchema.type === undefined) {
+    propSchema = {
+      ...propSchema,
+      type: ANY_TYPE,
+    };
+  }
+
+  return propSchema;
+};
+
+const toAdditionalPropertyItem = (
+  propName: string,
+  parentPath: string,
+  parentSchema: JsonSchema,
+  rootSchema: JsonSchema
+): AdditionalPropertyItem => {
+  let propSchema = getMatchingAdditionalPropertySchema(
+    propName,
+    parentSchema,
+    rootSchema
+  );
+  let propUiSchema: UISchemaElement = createControlElement('#');
+
+  if (typeof propSchema === 'object' && propSchema.type === 'array') {
+    propUiSchema = Generate.uiSchema(
+      propSchema,
+      'Group',
+      undefined,
+      rootSchema
+    );
+    (propUiSchema as GroupLayout).label =
+      propSchema.title ?? startCase(propName);
+  }
+
+  if (typeof propSchema === 'object') {
+    propSchema = {
+      ...propSchema,
+      title: propName,
+    };
+    if (propSchema.type === 'object') {
+      propSchema.additionalProperties =
+        propSchema.additionalProperties !== false
+          ? propSchema.additionalProperties ?? true
+          : false;
+    } else if (propSchema.type === 'array') {
+      propSchema.items = propSchema.items ?? {};
+    }
+  }
+
+  return {
+    propertyName: propName,
+    path: composePaths(parentPath, propName),
+    schema: propSchema,
+    uischema: propUiSchema,
+  };
+};
+
+const getPropertyNamePattern = (schema: JsonSchema): string | undefined => {
+  const objectSchema = toObjectSchema(schema);
+  const propertyNames = objectSchema.propertyNames as JsonSchema7 | undefined;
+  if (typeof propertyNames === 'object' && propertyNames.pattern) {
+    return propertyNames.pattern;
+  }
+
+  if (
+    objectSchema.additionalProperties === false &&
+    objectSchema.patternProperties
+  ) {
+    const patterns = Object.keys(objectSchema.patternProperties);
+    return patterns.length > 0 ? patterns.join('|') : undefined;
+  }
+
+  return undefined;
+};
+
+@Component({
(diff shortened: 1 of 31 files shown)


kchobantonov:
@sdirix  please review

packages/material-renderers/src/complex/MaterialMixedRenderer.tsx
```diff
@@ -0,0 +1,1810 @@
...
+
+      const uischema = findUISchema(
+        uischemas,
+        cleanedSchema,
+        control.scope,
+        path,
+        () => createControlElement(control.scope ?? '#'),
+        schemaControl,
+        rootSchema
+      );
```
sdirix, line 366: This dispatches the already resolved schema and already composed path together with a uischema whose scope is still the full `control.scope`, so the nested control resolves scope and path a second time and binds to e.g. `mixedSimple.mixedSimple`. Repro in the "Mixed Object" example: `mixedSimple` has data `"String"` but the field renders empty, and typing one character flips the control into tree mode showing char-indexed pseudo properties while the data stays untouched. The grafting in lines 339-356 papers over this for object/array types only. The Angular port has the same bug (and lacks the grafting entirely), Vue Vuetify does not.

packages/material-renderers/src/complex/MaterialMixedRenderer.tsx
```diff
@@ -0,0 +1,1810 @@
...
+};
+
+const createMixedRenderInfos = (
+  parentSchema: JsonSchema,
+  schema: JsonSchema,
+  rootSchema: JsonSchema,
+  control: ControlElement,
+  path: string,
+  uischemas: JsonFormsUISchemaRegistryEntry[]
+): SchemaRenderInfo[] => {
```
sdirix, line 263: `createMixedRenderInfos`, `cleanSchema`, `getSchemaTypesAsArray`, `getJsonDataType`, `buildTreeFromData` and the tester are framework independent and now exist three times: here, in `angular-material/src/library/other/mixed.renderer.ts` and in vue-vuetify. The copies already drifted (this one grafts the parent schema in lines 339-356, the Angular one does not). Core is the established home for this kind of logic, see `createCombinatorRenderInfos`. The shared model logic and testers should move to `@jsonforms/core` so only the view layer stays per framework.

packages/material-renderers/src/complex/MaterialMixedRenderer.tsx
```diff
@@ -0,0 +1,1810 @@
...
+        ) {
+          (resolvedSchema.items as JsonSchema7).type = [
+            'array',
+            'boolean',
+            'integer',
+            'null',
+            'number',
+            'object',
+            'string',
+          ];
```
sdirix, line 322: `resolvedSchema.items` is still the object from the input schema (the spread in line 273 is shallow) or from the rootSchema after `$ref` resolution, so writing `type` here permanently mutates the schema in the store. The `cloneDeep` only happens afterwards in line 326; clone before this block. Same issue in the Angular copy.

packages/material-renderers/src/complex/MaterialMixedRenderer.tsx
```diff
@@ -0,0 +1,1810 @@
...
+      ) {
+        cancelRename();
+        return;
+      }
+
+      if (trimmed in parentData) {
+        setRenameError(`Property "${trimmed}" already exists`);
+        return;
+      }
+
```
sdirix, line 1292: The add/rename flow in `MaterialAdditionalProperties` rejects names containing `.`, `[` or `]`, but the tree rename does not, so renaming a property to `a.b` here produces paths that `compose`/`get` split apart. Apply the same name validation (also in the Angular copy).

packages/material-renderers/src/complex/MaterialMixedRenderer.tsx
```diff
@@ -0,0 +1,1810 @@
...
+  schema: JsonSchema,
+  context: TesterContext
+) =>
+  isMixedSchema(uischema as UISchemaElement & Scopable, schema, context) &&
+  (isControl(uischema) || isDefaultGenUiSchema(uischema));
+
+export const materialMixedControlTester: RankedTester = rankWith(
+  20,
+  isMixedControl
+);
```
sdirix, line 1808: Rank 20 makes this claim every schema with a `type` array and every `schema: true` in all React Material apps, so a plain nullable `{"type": ["string", "null"]}` now renders the type selector UI where it previously fell through. This is a large breaking change for many applications.

packages/material-renderers/src/complex/MaterialMixedRenderer.tsx
sdirix, line 1: There is not a single new unit test for the ~5700 added lines across the three renderer sets (coverage drops 3.7% per coveralls). The pure helpers here (`createMixedRenderInfos`, `buildTreeFromData`, `cleanSchema`, the property name validation) are trivially unit-testable, and every other complex renderer in this package has a suite.

packages/material-renderers/src/complex/MaterialMixedRenderer.tsx
```diff
@@ -0,0 +1,1810 @@
...
+  if (typeof schema !== 'object') {
+    return [
+      'array',
+      'boolean',
+      'integer',
+      'null',
+      'number',
+      'object',
+      'string',
+    ];
```
sdirix, line 233: This 7-element type list literal is repeated 7 times in this file.

packages/material-renderers/src/complex/MaterialAdditionalProperties.tsx
```diff
@@ -0,0 +1,609 @@
...
+
+  if (
+    (!propSchema && typeof objectSchema.additionalProperties === 'object') ||
+    objectSchema.additionalProperties === true
+  ) {
+    propSchema =
+      objectSchema.additionalProperties === true
+        ? { additionalProperties: true }
+        : objectSchema.additionalProperties;
+  }
```
sdirix, line 130: Operator precedence bug: when a `patternProperties` entry matched, `additionalProperties === true` still overwrites `propSchema` because the `!propSchema` guard only covers the first operand. Should be `!propSchema && (typeof ... === 'object' || ... === true)`. Same code in `angular-material/src/library/other/additional-properties.renderer.ts` lines 109-117.

packages/material-renderers/src/complex/MaterialAdditionalProperties.tsx
```diff
@@ -0,0 +1,609 @@
...
+        {additionalPropertiesTitle ? (
+          <Typography sx={{ pt: 1 }} variant='body2'>
+            {additionalPropertiesTitle}
+          </Typography>
+        ) : null}
+        <TextField
+          fullWidth
+          error={Boolean(propertyNameError)}
+          helperText={propertyNameError}
+          label='Property Name'
```
sdirix, line 418: All user facing strings in the new React and Angular renderers are hardcoded English. They need to be translated.

packages/material-renderers/src/complex/MaterialObjectRenderer.tsx
```diff
@@ -24,28 +24,55 @@
...
+  if (uischema.type !== 'Group') {
+    return uischema;
+  }
+
+  const { label: _label, ...layout } = uischema as GroupLayout;
+  return {
+    ...layout,
+    type: 'VerticalLayout',
+  };
+};
```
sdirix, line 57: This also rewrites Group uischemas coming from the uischemas registry, so a user who explicitly registered a Group for a nested object silently gets a VerticalLayout inside your Card instead. The Card frame also duplicates what the Group renderer already draws.

Besides that, this is a massive breaking change as this will be a visual change for every existing app

packages/material-renderers/package.json
```diff
@@ -98,6 +98,7 @@
     "@mui/icons-material": "^7.0.0",
     "@mui/material": "^7.0.0",
     "@mui/x-date-pickers": "^8.0.0",
+    "@mui/x-tree-view": "^8.0.0",
```
sdirix, line 101: `@mui/x-tree-view` as a new peer dependency is a breaking change: every consumer has to install it even if they never render a mixed control.

packages/material-renderers/src/complex/DynamicPropertyDispatch.tsx
```diff
@@ -0,0 +1,111 @@
...
+              ? createDefaultValue(schema, rootSchema)
+              : value;
+          },
+        });
+        return;
+      }
+
+      jsonforms.dispatch?.(action);
+    };
+  }, [jsonforms, path, rootSchema, schema]);
```
sdirix, line 88: Each framework now solves "clearing must not drop the dynamic property" differently: React wraps the context dispatch here, Angular adds a `preserveUndefinedAsDefault` input to the binding, and Vue uses an injection flag. This concern should get one shared, core-level design instead of three mechanisms.

packages/angular/src/library/abstract-control.ts
```diff
@@ -56,6 +57,7 @@ export abstract class JsonFormsAbstractControl<
   @Input() id: string;
   @Input() disabled: boolean;
   @Input() visible: boolean;
+  @Input() preserveUndefinedAsDefault = false;
```
sdirix, line 60: This adds public API to the `@jsonforms/angular` binding for a renderer-set concern, without tests or docs, and diverges from how React (dispatch wrapper) and Vue (injection flag) solve the same problem in this PR.

packages/angular/src/library/jsonforms.component.ts
```diff
@@ -85,6 +89,21 @@ export class JsonFormsOutlet
...
+    if (
+      changes.schema ||
+      changes.uischema ||
+      changes.path ||
+      changes.preserveUndefinedAsDefault
+    ) {
+      this.update(this.jsonformsService.getState());
+    }
+  }
+
```
sdirix, line 106: This changes the update behavior of every `JsonFormsOutlet` usage, not just the mixed renderer (inputs previously only took effect via the `renderProps` setter and the state subscription). That is a behavior change to the framework binding and needs its own test coverage.

packages/angular-material/src/library/other/mixed.renderer.ts
```diff
@@ -0,0 +1,1678 @@
...
+    event.preventDefault();
+    this.draggingSplitter = true;
+  }
+
+  @HostListener('document:mousemove', ['$event'])
+  onDocumentMouseMove(event: MouseEvent): void {
+    if (!this.draggingSplitter) {
+      return;
+    }
+    this.treeWidth = Math.min(640, Math.max(220, event.clientX - 32));
```
sdirix, line 1519: Hardcoded `event.clientX - 32` assumes the tree pane starts 32px from the viewport edge. This needs to be more dynamic like the  React version already does.

packages/angular-material/src/library/other/additional-properties.renderer.ts
```diff
@@ -0,0 +1,698 @@
...
+      }
+      .additional-property-row:hover .additional-property-actions {
+        opacity: 1;
+      }
+      .additional-property-action-button.mat-mdc-icon-button {
+        --mdc-icon-button-state-layer-size: 24px;
+        height: 24px;
+        padding: 2px;
+        width: 24px;
+      }
```
sdirix, line 390: The buttons are shrunk to 24x24 but the MDC `.mat-mdc-button-touch-target` stays 48x48, so the delete button's touch target overlays the rename button, so clicking "rename" deletes the property.

packages/vue-vuetify/package.json
```diff
@@ -68,6 +68,7 @@
     "dayjs": "^1.10.6",
     "lodash": "^4.17.21",
     "maska": "^2.1.11",
+    "splitpanes": "^3.2.0",
```
sdirix, line 71: `splitpanes` moves from devDependencies to runtime dependencies and `VSplitpanes`/`VPane` become public exports of the package. React and Angular hand rolled the same functionality.

packages/vue-vuetify/src/icons/icons.ts
```diff
@@ -10,12 +10,27 @@ export type IconValue =
 
 export interface IconAliases {
   itemAdd: IconValue;
+  itemCancel: IconValue;
```
sdirix, line 13: `IconAliases` gains 17 new required members, which is a compile break for TS consumers that implement the interface for a custom icon set. Since `useIcons` merges overrides over the defaults anyway, these can be optional.

sdirix (changes requested):
Thanks for the effort that went into this. 

I tested all three renderer sets in the browser and the happy paths work.

However, the PR in its current state is far from mergeable, and I want to be transparent about why the bar here is different from previous `vue-vuetify` work: React Material is our most used renderer set. Changes there affect many applications, so default behavior changes, new peer dependencies, and untested code paths need a level of care that this PR does not yet have. Please see the inline comments for details; the summary of what it takes to get this mergeable:

**1. Move the shared logic to `@jsonforms/core`.** The utils, the tree building and the testers are framework independent and are now copy-pasted three times. A renderer set specific util is fine as long as only that renderer set uses the logic. As soon as we have the same logic in multiple renderer sets, the logic should be shared. Additionally the copies have already diverged, and that divergence is exactly why nested primitive mixed controls are broken in React and Angular but work in Vue.

**2. Fix the  bugs.** (a) Non-root mixed controls with primitive data are mis-bound in React and Angular: the field renders empty and typing flips the control into tree mode without ever updating the data. (b) In Angular, clicking the Rename button can delete the property because the MDC touch targets overlap. (c) The Angular splitter uses a hardcoded viewport offset. (d) `createMixedRenderInfos` mutates the schema in the store. (e) The `patternProperties`/`additionalProperties: true` precedence bug.

**3. No silent behavior changes for existing React Material and Angular Material users.** The object renderer rework (Card frames, rewriting registered Group uischemas), the rank-20 tester taking over every `type: [...]` and `schema: true`, and the new `@mui/x-tree-view` peer dependency all change or break existing applications. Each of these needs to be either removed, made unnoticeable, made opt-in, or explicitly agreed on with a changelog and migration note. Rewriting uischemas that users registered themselves is not acceptable in any case.

**4. One design for the "clearing must not drop the property" concern.** The PR ships three different mechanisms: a context dispatch wrapper in React, a new public `preserveUndefinedAsDefault` input on the `@jsonforms/angular` binding, and the existing injection flag in Vue. Framework binding API additions for a renderer-set feature need their own discussion, tests and docs. Let's design this once, at the core level.

**5. i18n and tests.** All user facing strings in the React and Angular renderers are hardcoded English, while vue-vuetify already translates exactly these. And there is not a single new unit test for ~5700 added lines (coverage drops 3.7%). The pure helpers are trivially testable once they live in core.

My suggestion for making this manageable: split the PR. The vue-vuetify UX improvements (which also currently conflict with master) can move forward quickly on their own. For React and Angular, start with a (draft) PR that extracts the shared logic into core with tests, then build the two renderers on top of it. I'm happy to review the core extraction design before you invest in the implementation.

kchobantonov:
@sdirix Thanks for the detailed review. I agree that the current scope is too large and that the shared logic needs a core-level design. I’ll supersede this PR with a Vue Vuetify-only PR first, including focused tests and the Vue-specific compatibility fixes. I’ll then open a draft core PR for the shared mixed-schema/tree/property logic and dynamic-property preservation design. React Material and Angular Material will follow as separate opt-in renderer PRs after that design is agreed.

kchobantonov:
@sdirix the vuetify work now in separate PR https://github.com/eclipsesource/jsonforms/pull/2610

kchobantonov:
closing this PR since it need separate work for react and angular

PR #2597, opened 2026-06-16
Pass explicit undefined defaults to optional jsonforms

kchobantonov:
Pass explicit undefined defaults to optional jsonforms and dispatch injections so Vue does not warn when helper components are rendered outside a JsonForms provider.


kchobantonov:
@sdirix please review

lucas-koehler (approved):
Thanks for the contribution ❤️  The change makes sense to me and looks good :)

PR #2601, opened 2026-07-16
fix(vue-vuetify): decouple NumberControl precision from step (fixes #2523)

### LINKED ISSUE #2523: [vue-vuetify] Wrong default precision for NumberControl
### Describe the bug

The default precision for number fields in a form is equial to the step and by default is one decimal after the point, which leads to rounding of values and unability for user to input more decimals places.
It seems from the code that precision in the NumberControRenderer depends on the step:
```
precision(): number | undefined {
      if (!this.step || Number.isInteger(this.step)) return undefined;
      // Handle scientific notation and float imprecision
      const stepStr = this.step.toString();
      if (stepStr.indexOf('e-') > -1) {
        // Handle cases like 1e-3
        return parseInt(stepStr.split('e-')[1], 10);
      }
      const fraction = stepStr.split('.')[1];
      return fraction ? fraction.length : undefined;
    }
```
which shouldn't be. It should be allowed to input a number of any precision by default.

Also to mention, that `precision` function must return `null`, not `undefined` according to the [vuetify docs](https://vuetifyjs.com/en/api/v-number-input/#props-precision)

### Expected behavior

It should be allowed to enter a number of any precision by default, not depending on the step. The precision should be limited only by an explicit setting.

1. Create any form with a number field, do not set a step parameter;
2. Try to input a fractual with more than one decimal after the point;
3. The form doesn't allow it;

### Screenshots

_No response_

v3.7.0

### Package

Vue Vuetify Renderers

### Additional context

_No response_

darius-lesch:
### Description
This PR resolves [#2523](https://github.com/eclipsesource/jsonforms/issues/2523) regarding the default precision limitation in the Vue-Vuetify `NumberControlRenderer`.

**The Problem:**
Previously, the `precision` computed property artificially restricted decimal places by tying them to the length of the `step` property (which defaults to `0.1`). Furthermore, when `step` was an integer, it returned `undefined`, which did not lift the restriction in Vuetify (Vuetify's `VNumberInput` requires `null` to allow infinite precision). This prevented users from typing standard decimal values.

**The Solution:**
* Decoupled `precision` from `step`.
* The control now returns `null` by default, allowing the native Vuetify input to accept unlimited decimal places.
* If a developer explicitly provides an `options.precision` property via the UI Schema, that specific configuration is now respected and passed to the input.

### Motivation and Context
Allows the `NumberControl` to behave naturally like a standard numeric input without abruptly truncating user data, while still allowing developers to restrict precision if they explicitly configure it in the UI Schema. 

### How Has This Been Tested?
* Added unit test asserting that `precision` resolves to `null` by default.
* Added unit test asserting that `options.precision` is correctly passed through to the `VNumberInput`.
* Updated the existing snapshot (the default value `1.0` correctly reverted to `1` as artificial precision formatting was removed).
* Tested manually in a downstream Vue 3 + Vite application.

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/packages/vue-vuetify/src/controls/NumberControlRenderer.vue
+++ b/packages/vue-vuetify/src/controls/NumberControlRenderer.vue
@@ -78,7 +78,6 @@ const controlRenderer = defineComponent({
     precision(): number | null {
       const options: any = this.appliedOptions;
 
-      // If precision is explicitly configured in the UI Schema options, respect it
       if (options.precision !== undefined && options.precision !== null) {
         return Number(options.precision);
       }
--- a/packages/vue-vuetify/src/controls/NumberControlRenderer.vue
+++ b/packages/vue-vuetify/src/controls/NumberControlRenderer.vue
@@ -75,16 +75,16 @@ const controlRenderer = defineComponent({
       const options: any = this.appliedOptions;
       return options.step ?? 0.1;
     },
-    precision(): number | undefined {
-      if (!this.step || Number.isInteger(this.step)) return undefined;
-      // Handle scientific notation and float imprecision
-      const stepStr = this.step.toString();
-      if (stepStr.indexOf('e-') > -1) {
-        // Handle cases like 1e-3
-        return parseInt(stepStr.split('e-')[1], 10);
+    precision(): number | null {
+      const options: any = this.appliedOptions;
+
+      // If precision is explicitly configured in the UI Schema options, respect it
+      if (options.precision !== undefined && options.precision !== null) {
+        return Number(options.precision);
       }
-      const fraction = stepStr.split('.')[1];
-      return fraction ? fraction.length : undefined;
+
+      // Return null (not undefined) to allow Vuetify to accept any precision by default
+      return null;
     },
     value(): number | null | undefined {
       if (
--- a/packages/vue-vuetify/tests/unit/controls/NumberControlRenderer.spec.ts
+++ b/packages/vue-vuetify/tests/unit/controls/NumberControlRenderer.spec.ts
@@ -4,6 +4,7 @@ import NumberControlRenderer from '../../../src/controls/NumberControlRenderer.v
 import { entry as numberControlRendererEntry } from '../../../src/controls/NumberControlRenderer.entry';
 import { wait } from '../../../tests';
 import { mountJsonForms } from '../util';
+import { VNumberInput } from 'vuetify/components';
 
 describe('NumberControlRenderer.vue', () => {
   const renderers = [numberControlRendererEntry];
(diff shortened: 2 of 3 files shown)


packages/vue-vuetify/tests/unit/controls/NumberControlRenderer.spec.ts
```diff
@@ -57,3 +57,35 @@ describe('NumberControlRenderer.vue', () => {
     expect(wrapper.html()).toMatchSnapshot();
   });
 });
+
+import { VNumberInput } from 'vuetify/components';
```
sdirix: Please move this up with the other imports at the top of the file.

packages/vue-vuetify/src/controls/NumberControlRenderer.vue
```diff
@@ -75,16 +75,16 @@ const controlRenderer = defineComponent({
...
-      if (!this.step || Number.isInteger(this.step)) return undefined;
-      // Handle scientific notation and float imprecision
-      const stepStr = this.step.toString();
-      if (stepStr.indexOf('e-') > -1) {
-        // Handle cases like 1e-3
-        return parseInt(stepStr.split('e-')[1], 10);
+    precision(): number | null {
+      const options: any = this.appliedOptions;
+
+      // If precision is explicitly configured in the UI Schema options, respect it
```
sdirix: This comment just restates the `if` below it, please remove it

packages/vue-vuetify/src/controls/NumberControlRenderer.vue
```diff
@@ -75,16 +75,16 @@ const controlRenderer = defineComponent({
...
+      if (options.precision !== undefined && options.precision !== null) {
+        return Number(options.precision);
       }
-      const fraction = stepStr.split('.')[1];
-      return fraction ? fraction.length : undefined;
+
+      // Return null (not undefined) to allow Vuetify to accept any precision by default
```
sdirix, line 85: This comment is fine as it describes why `null` is returned

sdirix (changes requested):
Thanks for the contribution!

Please have a look at my comments.

packages/vue-vuetify/tests/unit/controls/NumberControlRenderer.spec.ts
darius-lesch: Thanks for the feedback. Change has been implemented.

packages/vue-vuetify/src/controls/NumberControlRenderer.vue
darius-lesch: Comment has been removed.

packages/vue-vuetify/src/controls/NumberControlRenderer.vue
darius-lesch: Got it. Thanks for the feedback @sdirix!

sdirix (approved):

sdirix:
Thank you for the contribution :heart:

PR #2602, opened 2026-07-19
fix(@vue-vuetify): reuse the parent AJV

KiaraGrouwstra:
Reuse the same AJV instance in @vue-vuetify.
Ensures that user configuration of AJV is retained.

Disclaimer: I used a coding agent in the creation of this patch.

### DIFF
--- a/packages/vue-vuetify/tests/unit/additional/AdditionalPropertiesAjv.spec.ts
+++ b/packages/vue-vuetify/tests/unit/additional/AdditionalPropertiesAjv.spec.ts
@@ -1,0 +1,43 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { clearAllIds, createAjv } from '@jsonforms/core';
+import { markRaw } from 'vue';
+import { extendedVuetifyRenderers } from '../../../src';
+import { mountJsonForms } from '../util/util';
+
+// Check that `additionalProperties` reuses the parent form's AJV
+describe('AdditionalProperties nested AJV', () => {
+  // A map whose key `pattern` is legal only without the `u` flag
+  const schema = {
+    type: 'object' as const,
+    properties: {
+      secretFiles: {
+        type: 'object' as const,
+        additionalProperties: { type: 'string' as const },
+        propertyNames: {
+          pattern: '^"([^"$\\\\]|\\$(?!{)|\\\\.)*"$',
+        },
+      },
+    },
+  };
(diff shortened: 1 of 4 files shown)


lucas-koehler (commented):
Hi @KiaraGrouwstra ,
thank you very much for contributing to JSON Forms ❤️ 
I had a look at the change and I don't think it makes a difference:
`useJsonForms()` of the vue package already injects the JSON Forms core state. Its `ajv` property should be the same ajv as injected by `useAjv()`. See `useAjv`s code here:
https://github.com/eclipsesource/jsonforms/blob/6db5bc87e763922c6853d76b166885978e280c74/packages/vue/src/jsonFormsCompositions.ts#L608-L620

Am I missing something that  makes this change necessary?

KiaraGrouwstra:
Hi @lucas-koehler, thanks for your response!
I added a test demonstrating the underlying issue now.

packages/vue-vuetify/tests/unit/additional/AdditionalPropertiesAjv.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+import { mount } from '@vue/test-utils';
+import { markRaw, defineComponent, h } from 'vue';
+import { createVuetify } from 'vuetify';
+import * as components from 'vuetify/components';
+import * as directives from 'vuetify/directives';
+import { VApp } from 'vuetify/components';
+import { extendedVuetifyRenderers } from '../../../src';
+
+global.ResizeObserver = require('resize-observer-polyfill');
+
```
sdirix: This duplicates the vuetify + ResizeObserver bootstrap from [`tests/unit/util/util.ts`](https://github.com/eclipsesource/jsonforms/blob/6db5bc87e763922c6853d76b166885978e280c74/packages/vue-vuetify/tests/unit/util/util.ts#L14-L19). `mountJsonForms` doesn't take an `ajv` yet, so extending it with an optional param would let this test reuse the shared helper instead of re-rolling the setup.

lucas-koehler:
@KiaraGrouwstra Sorry for the confusion, I mislooked what excatly `useJsonForms` returns vs. how ajv is attempted to be extracted from that. Using `useAjv` instead of course makes sense :+1:

lucas-koehler (approved):
LGTM now! Thanks for the contribution ❤️

PR #2603, opened 2026-07-19
fix(vue-vuetify): resolve a `$ref` on propertyNames in the additional-properties nested form

KiaraGrouwstra:
Resolve a `$ref` on @vue-vuetify `additionalProperties`' `propertyNames` against the root schema (`control.value.rootSchema`, via `Resolve.schema`) before spreading it into the nested form's schema.
Further add a test verifying these are resolved with this change.

The `additional-properties` (map) renderer builds the nested new-key form's schema by spreading `control.value.schema.propertyNames`. When `propertyNames` is a `$ref` into the root's `$defs` (the shape a schema that shares a named key type across many maps emits), the nested `<json-forms>` mounts it as its _own_ root, where `#/$defs/...` does not exist. AJV then throws `can't resolve reference #/$defs/... from id #` and the map fails to render.

Disclaimer: I used a coding agent in the creation of this patch.

### DIFF
--- a/packages/vue-vuetify/tests/unit/additional/AdditionalPropertiesRef.spec.ts
+++ b/packages/vue-vuetify/tests/unit/additional/AdditionalPropertiesRef.spec.ts
@@ -1,0 +1,35 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { clearAllIds } from '@jsonforms/core';
+import { extendedVuetifyRenderers } from '../../../src';
+import { mountJsonForms } from '../util';
+
+// Test use of `$ref` in additional-properties `propertyNames`
+describe('AdditionalProperties nested $ref propertyNames', () => {
+  const schema = {
+    type: 'object' as const,
+    $defs: {
+      attrName: {
+        type: 'string' as const,
+        pattern: '^[A-Za-z_][A-Za-z0-9_]*$',
+      },
+    },
+    properties: {
+      secretFiles: {
+        type: 'object' as const,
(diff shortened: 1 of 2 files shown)


packages/vue-vuetify/tests/unit/additional/AdditionalPropertiesRef.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+          h(JsonForms, {
+            data: { secretFiles: {} },
+            schema,
+            uischema,
+            renderers: markRaw(extendedVuetifyRenderers),
+            onChange: () => undefined,
+          }),
+        );
+    },
+  });
```
sdirix: This reimplements the vuetify setup (createVuetify, the ResizeObserver polyfill, and a `VApp` wrapper) that the shared `mountJsonForms` helper and its `TestComponent` already provide: https://github.com/eclipsesource/jsonforms/blob/6db5bc87e763922c6853d76b166885978e280c74/packages/vue-vuetify/tests/unit/util/util.ts#L21-L43

The whole block could collapse to `mountJsonForms({ secretFiles: {} }, schema, extendedVuetifyRenderers, uischema)`, matching the other specs in this directory.

sdirix (changes requested):
Thank you for the contribution! Looks good! But the test could be simplified. Please have a look.

sdirix (approved):
Tank you for the contribution ❤️

PR #2607, opened 2026-07-24
fix (vue-vanilla): preserve strict data types in vue-vanilla enum controls (fixes #2450)

### LINKED ISSUE #2450: (in vue-vanilla) EnumControlRenderer.vue (and EnumOneOf...) always sets String vALUE in Data Json
### Describe the bug

I have this field in my schema:
“land”: { “title”: “Land”, “type”: “integer”,
“oneOf”:[ {“const”: 19, “title”: “Deutschland” },
{“const”: 20, “title”: “Frankreich” } ]
},
but when i select one entry of the drop down list, i get a String of the integer in my json-object:
“land”: “19”,

so an error always occurs: “must be integer must match exactly one schema in oneOf”

this happens in  EnumOneOfControlRenderer and is the same in EnumControlRenderer:
target.selectedIndex === 0 ? undefined : target.value  //  << this is always stringified

### Expected behavior

an integer value should be set in the json-objekt
“land”: 19,

open a jsonform with an enum of type integer in the jsonschema and select a value.

### Screenshots

_No response_

v3.5.1

### Package

Vue Vanilla Renderers

### Additional context

@sirix wrote in the forum: ""
if you get a string in the data, then this means the renderer dispatched the wrong data. It should use the const of the schema it is handed over, instead it’s likely using the value from the input element it is rendering which seems to stringify them.
[Here the actual options should be used, instead we use the value of the DOM element](https://github.com/eclipsesource/jsonforms/blob/52182632589658b61394c870df1cefacff7d8775/packages/vue-vanilla/src/controls/EnumOneOfControlRenderer.vue#L56). Feel free to open an issue and contributing a fix if you’d like
""
i helped me wirh this peace of code:
target.selectedIndex === 0 ? undefined : ( Number.isNaN(target.value) ? target.value :Number(target.value) )

darius-lesch:
Fixes #2450 
**Description**
This PR fixes an issue in the `vue-vanilla` package (#2450) where selecting a non-string value (like an integer) from an enum dropdown implicitly coerces the value to a string in the JSONForms state, causing schema validation failures.

**Main Changes**

* **Updated `EnumControlRenderer.vue` & `EnumOneOfControlRenderer.vue**`: Modified the `useVanillaControl` parsing callback to extract the selected value directly from the `control.value.options` state array rather than reading `target.value` from the DOM.
* **Index Mapping**: Applied a `target.selectedIndex - 1` lookup to accurately map the DOM selection back to the state array, accounting for the hardcoded empty `<option>` at index `0`.
* **Test Coverage**: Added new unit tests for both renderers using `integer` schemas to explicitly verify that strict number types are preserved in the data state.

**Motivation & Justification**
Native HTML `<select>` elements inherently serialize all bound data into strings. The previous implementation relied on reading the DOM's `target.value`, which destroyed the original type information. Using the DOM's `selectedIndex` as a pointer to fetch the original value from the JSONForms state array bypasses this serialization completely. This provides a robust, universally type-safe solution for all schema types (numbers, booleans, etc.) without relying on brittle manual type-casting logic.

### DIFF
--- a/packages/vue-vanilla/src/controls/EnumControlRenderer.vue
+++ b/packages/vue-vanilla/src/controls/EnumControlRenderer.vue
@@ -52,8 +52,11 @@ const controlRenderer = defineComponent({
     ...rendererProps<ControlElement>(),
   },
   setup(props: RendererProps<ControlElement>) {
-    return useVanillaControl(useJsonFormsEnumControl(props), (target) =>
-      target.selectedIndex === 0 ? undefined : target.value
+    const input = useJsonFormsEnumControl(props);
+    return useVanillaControl(input, (target) =>
+      target.selectedIndex === 0
+        ? undefined
+        : input.control.value.options[target.selectedIndex - 1].value
     );
   },
 });
--- a/packages/vue-vanilla/src/controls/EnumOneOfControlRenderer.vue
+++ b/packages/vue-vanilla/src/controls/EnumOneOfControlRenderer.vue
@@ -52,8 +52,11 @@ const controlRenderer = defineComponent({
     ...rendererProps<ControlElement>(),
   },
   setup(props: RendererProps<ControlElement>) {
-    return useVanillaControl(useJsonFormsOneOfEnumControl(props), (target) =>
-      target.selectedIndex === 0 ? undefined : target.value
+    const input = useJsonFormsOneOfEnumControl(props);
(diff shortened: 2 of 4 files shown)


lucas-koehler (approved):
Hi @darius-lesch , thanks for the contribution ❤️ 
The changes LGTM and are a nice improvement 🚀

PR #2609, opened 2026-07-30
Resolve PR #2585 Change Requests

darius-lesch:
The PR resolves two comments in #2585:

**1. `packages/core/src/util/setData.ts`**

The use of the loose `!= null` was originally intentional, as it is a standard JavaScript shorthand to check for both `null` and `undefined` simultaneously. We can't use `!data` in this context because it would be unsafe; it would incorrectly reject valid falsy data values that we want to keep, such as `0`, `false`, or an empty string `""`.

However, I completely agree that loose inequality can look like a typo or cause confusion at first glance. To make the intent strictly clear and avoid any "weirdness", I've updated the logic to use the explicit multiple cases as you suggested: `data !== null && data !== undefined`.

**2. `MIGRATION.md`**

I have also moved the data update path documentation into a new `## Migrating to JSON Forms 3.9` section in `MIGRATION.md`.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,17 @@
 # Migration guide
 
+## Migration to JSON Forms 3.9
+
+### Data update paths treat all segments literally
+
+Data updates (e.g. dispatched `update` actions) previously wrote to the form data via lodash's `set`/`unset`, which interpret bracket notation and array indices in paths.
+This corrupted data for property names that look like lodash path syntax, for example numeric property names like `"15"` were turned into array indices and names containing brackets like `"prop[0]"` were split up (see [#2397](https://github.com/eclipsesource/jsonforms/issues/2397) and [#2102](https://github.com/eclipsesource/jsonforms/issues/2102)).
+
+Updates now use the new `setDataAt`/`unsetDataAt` utilities of `@jsonforms/core`, which split paths on `.` and treat every segment as a literal property name, matching how JSON Forms resolves values for display.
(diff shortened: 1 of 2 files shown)


darius-lesch:
@sdirix : Hope this helps resolve the change requests in PR #2585, get milestone 3.9 complete. and Let me know if there's anything else needed to get v3.9 shipped!

sdirix:
Hi @darius-lesch, do you need JSON Forms 3.9 urgently shipped for your use cases? Would a pre-release of the current state be sufficient?

darius-lesch:
> Hi @darius-lesch, do you need JSON Forms 3.9 urgently shipped for your use cases? Would a pre-release of the current state be sufficient?

Hi @sdirix , no, there is no immediate urgency for my use cases. A pre-release of the current state is more than sufficient.

lucas-koehler (changes requested):
Hi @darius-lesch , thanks for the fixes ❤️ 
There is a small lint error due to expected formatting  in the changed code line.

darius-lesch:
> Hi @darius-lesch , thanks for the fixes ❤️ There is a small lint error due to expected formatting in the changed code line.

Hi @lucas-koehler , linting issue has been resolved now.

lucas-koehler (approved):
LGTM now, thanks for the update!

lucas-koehler:
@darius-lesch a pre-release `3.9.0-alpha.0` should now be available containing all latest commits :)

PR #2610, opened 2026-08-02
feat(vue-vuetify): improve mixed and additional property editors

kchobantonov:
## Summary

This PR improves the Vue Vuetify handling of mixed-type and dynamic additional properties.

It adds a tree-based editor for object and array values while keeping primitive values editable through the existing mixed renderer. The implementation is scoped to the Vue Vuetify package.

## Changes

- Add a split-pane tree/detail view for mixed object and array values.
- Support navigating, renaming, and deleting dynamic additional properties from the tree.
- Use the same property-name validation for renaming through:
  - the Additional Properties component
  - the mixed-renderer tree
- Validate renamed properties against:
  - existing property names
  - `propertyNames`
  - applicable `patternProperties`
- Preserve the property value and its type when renaming.
- Correctly update mixed renderers when an additional property’s value type changes.
- Avoid mutating the provided JSON Schema while preparing renderer schemas.
- Support property names containing brackets while rejecting dots, which are JSON Forms path separators.
- Add translated labels, tooltips, accessibility labels, generated array-item labels, and validation errors.
- Add `splitpanes` to the Vue Vuetify peer and development dependencies.

## Validation

- Added tests for dynamic property-name validation and translation handling.
- Added coverage for additional-property value type changes.


kchobantonov:
@sdirix please review

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -435,31 +1108,350 @@ const controlRenderer = defineComponent({
...
+      const result = [toTreeNodeId(input.control.value.path)];
+      let currentPath = input.control.value.path;
+
+      segments.slice(0, -1).forEach((segment) => {
+        currentPath = composePropertyPath(currentPath, segment);
+        result.push(toTreeNodeId(currentPath));
+      });
+
+      return result;
+    };
```
EclipseSourceAI, line 1170: This treats `path` as relative to `input.control.value.path`, but every caller passes an absolute path (`selectCurrentPath`, `commitRename`, `deleteNode`). For control path `dynamic` and target `dynamic.a.b` it produces `dynamic`, `dynamic.dynamic`, `dynamic.dynamic.a`, so the actual ancestor `dynamic.a` is never opened and the bogus ids get dropped again by the `treeNodes` watcher.

I reproduced it in the example app: with `additionalProperties` data `{"dynamic":{"a":{"b":{"c":1}}}}`, clicking the eye button on `b` switches the detail pane but leaves `a` collapsed and nothing highlighted in the tree. Depth 1 appears to work only because the root id is always included, and a root-scoped mixed control works because the prefix is empty. Running `getRelativePath(path)` first fixes it.

packages/vue-vuetify/src/complex/components/AdditionalProperties.vue
```diff
@@ -505,45 +543,120 @@ export default defineComponent({
...
+      } else if (schemaType === 'object') {
+        classes.push('additional-property-row--object');
+      } else if (schemaType === 'array') {
+        classes.push('additional-property-row--array');
+      } else {
+        classes.push('additional-property-row--primitive');
+      }
+
+      return classes;
+    },
```
EclipseSourceAI, line 582: The `--mixed`/`--object`/`--array`/`--primitive` modifiers aren't referenced by the new scoped styles or anywhere else in the package, so this computes classes nobody uses. Either style them or drop the method and put `class="additional-property-row"` on the row directly.

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -18,86 +18,277 @@
...
+                          hide-details
+                          autofocus
+                          :error="Boolean(renameError)"
+                          :title="renameError ?? undefined"
+                          v-bind="vuetifyProps('v-text-field')"
+                          @click.stop
+                          @keydown.stop.enter="commitRename(item)"
+                          @keydown.stop.esc="cancelRename"
+                          @blur="commitRename(item)"
+                        />
```
EclipseSourceAI, line 93: The rename error is only reachable via the native `title` tooltip here (the field is `hide-details`), and `renameError` isn't recomputed while typing, so it stays red after you've fixed the name. The rename menu in `AdditionalProperties.vue` does this properly with `:error-messages` plus `@update:model-value="updateRenameError(...)"`; worth matching. In the browser the "already defined" message is invisible unless you hover the field.

packages/vue-vuetify/package.json
```diff
@@ -68,6 +68,7 @@
     "dayjs": "^1.10.6",
     "lodash": "^4.17.21",
     "maska": "^2.1.11",
+    "splitpanes": "^3.1.5",
```
EclipseSourceAI, line 71: This makes `splitpanes` a hard install requirement for every consumer of the package, including those that never render a mixed type, and there's no `peerDependenciesMeta` marking it optional. Combined with `VSplitpanes.sass` re-implementing splitpanes' own base layout CSS (the `splitpanes/dist/splitpanes.css` import is gone), the package takes on 218 lines that have to track upstream. A maintainer should make that call explicitly, and the README quick start probably needs the new peer added.

packages/vue-vuetify/src/util/dynamicProperties.ts
```diff
@@ -0,0 +1,174 @@
...
+  data: unknown;
+  propertyNameSchema?: JsonSchema7;
+  ajv?: Ajv;
+}
+
+export const composePropertyPath = (
+  parentPath: string,
+  propertyName: string,
+): string => (parentPath ? `${parentPath}.${propertyName}` : propertyName);
+
```
EclipseSourceAI, line 22: Always dot-joining is right for core's current path handling, since `resolveData` and `setDataAt` split purely on `.` ([setData.ts#L31](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/core/src/util/setData.ts#L31)), whereas core's `compose` drops the separator when the segment starts with `[` ([path.ts#L29](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/core/src/util/path.ts#L29)). The downside is that a `[foo]` property created here won't resolve in material/vanilla/angular, which all still use `composePaths`. Fixing `compose` in core would make bracket support work everywhere instead of just this renderer set.

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -346,18 +601,394 @@ export function getJsonDataType(value: any): string | null {
...
+    if (node.nodeId === targetPath) {
+      return node;
+    }
+    const child = findNodeByPath(node.children ?? [], targetPath);
+    if (child) {
+      return child;
+    }
+  }
+  return undefined;
+}
```
EclipseSourceAI, line 971: These are all pure, Vue-free functions and they're the part of this PR most likely to regress. Moving them next to the new `src/util/dynamicProperties.ts` would let them be unit tested the same way the validation helpers now are, and would cut this SFC back to something reviewable. Small thing while you're in there: `findNodeByPath` compares `node.nodeId`, so `targetPath` is a misleading parameter name.

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -346,18 +601,394 @@ export function getJsonDataType(value: any): string | null {
...
+  return {
+    id: path,
+    schema,
+    uischema: createControlElement('#'),
+    path,
+    label,
+    required: false,
+    enabled,
+    readonly,
+  };
```
EclipseSourceAI, line 754: `id`, `required` and `label` are set for every node but never read (the detail pane only uses `schema`/`uischema`/`path`/`enabled`/`readonly`, and the template reads `item.label`). Dropping them from `TreeNodeControl` makes it obvious what the detail pane actually needs.

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -18,86 +18,277 @@
...
+        <v-tooltip location="top">
+          <template #activator="{ props }">
+            <v-btn
+              v-bind="props"
+              class="mixed-navigate-button"
+              :icon="icons.current.value.visibilityOn"
+              :aria-label="mixedTranslations.viewAriaLabel(computedLabel)"
+              variant="text"
+              color="primary"
+              :disabled="!navigationContext"
```
EclipseSourceAI, line 230: This can never be true: the branch only renders when `isNestedComplexType` is set, which requires `!isRoot`, which requires `navigationContext` to exist.

packages/vue-vuetify/src/complex/MixedRenderer.vue
```diff
@@ -435,31 +1108,350 @@ const controlRenderer = defineComponent({
...
+            input.control.value.rootSchema,
+            input.control.value.path,
+            vuetifyControl.computedLabel.value,
+            input.control.value.enabled,
+            input.control.value.readonly,
+            showPrimitivesInTree.value,
+            mixedTranslations.itemLabel,
+          )
+        : [],
+    );
```
EclipseSourceAI, line 1143: `treeNodes` depends on `control.data`, so every keystroke in the detail pane rebuilds the whole tree and each node runs `cloneDeep` + `cleanSchema` (twice for arrays, via `prepareArraySchema`). Fine for the example data, but it could get noticeable on larger objects; caching per node path or keying the rebuild on the data shape rather than the value would help.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
> This review was done by an AI agent and therefore may contain mistakes. Feel free to ignore any comment you disagree with. Noting *why* helps, since replies are read and taken into account in follow-up reviews.
>
>
>

Reworks mixed-type and dynamic `additionalProperties` editing in `@jsonforms/vue-vuetify`: object/array mixed values now get a splitpanes tree + detail editor, nested mixed controls navigate into that tree instead of nesting renderers, and property renaming is added both in the tree and in the Additional Properties editor.

The extraction of `src/util/dynamicProperties.ts` is the strongest part: `findPropertySchema` and `getPropertyNameSchema` come straight out of `AdditionalProperties.vue` and are now shared by both rename paths, with unit tests. The non-mutating schema preparation in `createMixedRenderInfos` and replacing the `JSON.stringify` descriptor-rebuild check with `haveAdditionalPropertyNamesChanged` are both real fixes.

Where to focus:

- `getPathAncestorNodeIds` in `MixedRenderer.vue` mixes up absolute and relative paths, so "reveal in tree" never expands ancestors when the mixed control's own path is non-empty (the normal `additionalProperties` case). Reproduced in the running example app on both the Additional Properties and Mixed Object examples.
- Making `splitpanes` a required peer dependency plus re-implementing its base CSS in `VSplitpanes.sass` is a packaging decision that deserves an explicit call.
- `MixedRenderer.vue` is now ~1600 lines; the pure tree-building layer would be better off in `src/util` where it can be tested.

`pnpm test`, `pnpm lint`, `pnpm type-check` and the full `pnpm build` all pass on the merged branch, and the CSS output includes the splitpanes rules.

packages/vue-vuetify/src/util/dynamicProperties.ts
kchobantonov: This needs separate PR fix since that could impact other code

kchobantonov:
@sdirix @lucas-koehler ready for review - also side question there is vuetify4 already - are we planning to upgrade ?

PR #2614, opened 2026-08-05
MUI v9 upgrade (material-renderers)

### LINKED ISSUE #2569: react-material: Support MUI v9
The react material renderers should support latest MUI and MUI X [version 9](https://mui.com/blog/introducing-mui-v9/).

### Package

React Material Renderers

jgteske:


packages/material-renderers/package.json
```diff
@@ -95,19 +95,19 @@
     "@emotion/styled": "^11.3.0",
     "@jsonforms/core": "3.9.0-alpha.0",
     "@jsonforms/react": "3.9.0-alpha.0",
-    "@mui/icons-material": "^7.0.0",
-    "@mui/material": "^7.0.0",
-    "@mui/x-date-pickers": "^8.0.0",
+    "@mui/icons-material": "^7.3.0 || ^9.0.0",
+    "@mui/material": "^7.3.0 || ^9.0.0",
```
lucas-koehler: Before a full review and you invest anymore effort into supporting both of these, we do not need to continue supporting MUI 7.
If we cannot guarantee it still works I'd rather remove the possibility for MUI 7 than consumers running into issues using JSON Forms with that. As this can easily happen when changing the renderer set later based on MUI 9, please remove support for MUI 7 :)

When talking about trying to avoid API breakage in the issue, this did not include the MUI version :D Sorry if that was not clear

jgteske: Technically it should work, due to the changes on MUIs side are mainly removing of deprecated function. So they exist in both Versions.

But I totally understand that removing the support is the better approach and if this is fine, I will remove the support. :)

jgteske: Support for MUI 7 is removed :)

packages/material-renderers/src/mui-controls/MuiAutocomplete.tsx
```diff
@@ -41,6 +42,16 @@ import {
...
+// Material UI 7.x delivers the input wiring of renderInput params as legacy
+// top-level props (InputProps/inputProps/InputLabelProps), while v9 nests it
+// under slotProps. Both majors are supported as peers, so both shapes must be
+// normalized to slotProps before rendering the TextField.
+type LegacyRenderInputParams = {
+  InputProps?: Record<string, unknown>;
+  inputProps?: Record<string, unknown>;
+  InputLabelProps?: Record<string, unknown>;
+};
+
```
lucas-koehler: This and its usage below should be obsolete now that we decided to remove MUI 7 support. Thus, it should be removed again.

packages/material-renderers/src/complex/ValidationIcon.tsx
```diff
@@ -40,7 +40,7 @@ const ValidationIcon: React.FC<ValidationProps> = ({ errorMessages, id }) => {
   return (
     <Tooltip id={id} title={errorMessages}>
       <StyledBadge badgeContent={errorMessages.split('\n').length}>
-        <ErrorOutline color='inherit' />
+        <ErrorOutlined color='inherit' />
```
lucas-koehler: Icon `ErrorOutlineOutlined` is the equivalent to v7's `ErrorOutline`. Please adapt this.

lucas-koehler (changes requested):
Hi @jgteske , thank you very much for this contribution ❤️ 
The changes already look pretty good to me and worked well using the example app!

Please have a look at the two inline comments.
Besides that, please add a note to the MIGRATION.md regarding the MUI update. See https://github.com/eclipsesource/jsonforms/blob/master/MIGRATION.md#update-to-material-ui-v5-in-react-material for a previous Material UI update note there.

Note: I opened a follow up issue https://github.com/eclipsesource/jsonforms/issues/2615 that was discovered here but the same behavior was already present before the upgrade in this PR. Thus, we do not need to fix this here. Even if you want to contribute this (which of course is always welcome :heart:  ), I think a separate PR is preferable so we can merge this PR ASAP and create a pre-release.

packages/material-renderers/src/util/layout.tsx
```diff
@@ -78,7 +78,7 @@ const MaterialLayoutRendererComponent = ({
     return (
       <Grid
         container
-        direction={direction}
+        sx={{ flexDirection: direction }}
```
EclipseSourceAI: Why is `direction` moved into `sx` here while the same PR keeps it as a Grid prop in `MaterialTableControl.tsx` (line 312) and `ExpandPanelRenderer.tsx` (line 166), and `MaterialListWithDetailRenderer` keeps it untouched ([link](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/material-renderers/src/additional/MaterialListWithDetailRenderer.tsx#L137-L137))? If those still compile, `direction={direction}` works in v9 and this change is unnecessary churn; if it does not, the other three need the same fix.

packages/material-renderers/src/mui-controls/MuiAutocomplete.tsx
```diff
@@ -137,7 +156,18 @@ export const MuiAutocomplete = (
...
+                htmlInput: {
+                  ...legacyHtmlInputProps,
+                  ...paramsSlotProps?.htmlInput,
+                },
+                inputLabel: {
+                  ...legacyInputLabelProps,
+                  ...paramsSlotProps?.inputLabel,
+                  ...(data ? { shrink: true } : undefined),
+                },
+              }}
```
EclipseSourceAI: When the v7 shim goes away, spread `params.slotProps` here instead of enumerating three slots. This `slotProps` replaces the one that `{...params}` already passes, so any slot the Autocomplete render params carry beyond `input`/`htmlInput`/`inputLabel` is silently dropped.

packages/material-renderers/src/controls/MaterialNativeControl.tsx
```diff
@@ -93,7 +93,7 @@ export const MaterialNativeControl = (props: ControlProps) => {
       onFocus={onFocus}
       onBlur={onBlur}
       helperText={!isValid ? errors : showDescription ? description : null}
-      InputLabelProps={{ shrink: true }}
+      slotProps={{ inputLabel: { shrink: true } }}
```
EclipseSourceAI, line 96: The TextField based controls are migrated, but the InputBase based ones still pass `inputProps` straight through ([MuiInputText](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/material-renderers/src/mui-controls/MuiInputText.tsx#L124-L124), same in `MuiInputNumber`, `MuiInputInteger`, `MuiInputNumberFormat` and `MaterialAnyOfStringOrEnumControl`), while `MuiCheckbox` already uses the slot form ([link](https://github.com/eclipsesource/jsonforms/blob/141f28838414f4c117d7b024bff9d8b09d800b82/packages/material-renderers/src/mui-controls/MuiCheckbox.tsx#L62-L62)). Was `inputProps` checked against v9 for those, or is it deliberately left as is?

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Upgrades `@jsonforms/material-renderers` from MUI 7 to MUI 9 (plus x-date-pickers 8 to 9) and migrates the props that v9 dropped: `inputProps`/`InputLabelProps` become `slotProps.htmlInput`/`slotProps.inputLabel`, Grid's `justifyContent`/`alignItems` move into `sx`, `Stack width` becomes `sx`, and one icon is renamed. The lockfile diff is clean (only MUI plus a transitive `@babel/runtime` bump) and there are no unrelated drive-by changes.

Points worth attention:
- `MuiAutocomplete` still carries the v7 compatibility shim, which contradicts the dropped v7 peer support (lucas-koehler's open thread). Beyond removing it, the hand-rolled three-slot merge there is the most fragile piece of the PR.
- `direction` is handled two different ways within the same PR: moved into `sx` in `util/layout.tsx` but kept as a Grid prop in `MaterialTableControl`/`ExpandPanelRenderer`.
- The `inputProps` usages on the InputBase based cells were not touched, so it is unclear whether they were checked against v9.
- Still open from the previous review: the `ErrorOutlineOutlined` icon and the MIGRATION.md note.

I could not verify anything in the UI (no network access in this run, so neither the deploy preview nor a local build was reachable). Since the sx migration changes how Grid alignment is applied, a maintainer should eyeball the array table rows, accordion headers and horizontal layouts.

packages/material-renderers/src/mui-controls/MuiAutocomplete.tsx
jgteske: This is totally correct and I will revert this changes :)

packages/material-renderers/src/controls/MaterialNativeControl.tsx
jgteske: The [MuiInputBase API](https://mui.com/material-ui/api/input-base/) still specifies `inputProps` as the way to go and not as deprecated. That why it is not changed for `MuiInputText`, ...

For [Checkbox](https://mui.com/material-ui/api/checkbox/) and [TextField](https://mui.com/material-ui/api/text-field/) the props were removed.

packages/material-renderers/src/util/layout.tsx
jgteske: The difference is the value being passed, not inconsistency.

The Prop direction still exists ([GridDirection](https://mui.com/material-ui/api/grid/)), but the support for `'column'`-types  was removed.

- this component (layout.tsx) directly sets the type to either `'row'` or `'column'`. With this changes the core functionality does not change.
- the other here mentioned components directly set the direction to `'row'`. We could remove them there as the `direction='row'` is default for this components.

jgteske:
> Hi @jgteske , thank you very much for this contribution ❤️ The changes already look pretty good to me and worked well using the example app!
> 
> Please have a look at the two inline comments. Besides that, please add a note to the MIGRATION.md regarding the MUI update. See https://github.com/eclipsesource/jsonforms/blob/master/MIGRATION.md#update-to-material-ui-v5-in-react-material for a previous Material UI update note there.
> 
> Note: I opened a follow up issue #2615 that was discovered here but the same behavior was already present before the upgrade in this PR. Thus, we do not need to fix this here. Even if you want to contribute this (which of course is always welcome ❤️ ), I think a separate PR is preferable so we can merge this PR ASAP and create a pre-release.

Thanks for your code review and your requested changes. 
I applied the changes and answered the other code-reviews :)

packages/material-renderers/src/controls/MaterialNativeControl.tsx
lucas-koehler: Thanks for the explanation. This makes sense to me as it keeps the amount of changed code lower 👍

packages/material-renderers/src/util/layout.tsx
EclipseSourceAI: Makes sense, thanks. `direction` here is an actual `'row' | 'column'` variable so it needs the `sx` workaround, while `MaterialTableControl`/`ExpandPanelRenderer` just hardcode the `'row'` default as a literal prop. Not an inconsistency, resolving.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up on the two commits pushed since my last review (`f9e2e43b`, `87a79fdd`).

Both fixes landed cleanly:
- `MuiAutocomplete.tsx`: the v7 compatibility shim is gone and `renderInput` now spreads `params.slotProps` directly, only overriding `inputLabel` for the `shrink` behavior. This addresses both my slot-merging concern and lucas-koehler's request to drop v7 support.
- `ValidationIcon.tsx`: icon renamed to `ErrorOutlineOutlined` as requested.

The `direction` question on `layout.tsx` is resolved too: the author's explanation checks out (dynamic `'row'|'column'` there vs. hardcoded `'row'` literals elsewhere), not an inconsistency, so I resolved that thread.

Nothing new to flag in the diff since last time. Still outstanding from the earlier round: the MIGRATION.md note lucas-koehler asked for hasn't been added yet, a maintainer should confirm that's still expected before merge.

packages/material-renderers/src/util/layout.tsx
lucas-koehler: As the support for column was removed, we should probably use a Stack instead of Grid if `direction` is `column`. This is recommended by the MUI API for vertical layouts. WDYT?

>the other here mentioned components directly set the direction to 'row'. We could remove them there as the direction='row' is default for this components.

I think this definitely makes sense. Could you do this, please?

lucas-koehler (changes requested):
Hi @jgteske , thanks for the updates! It almost LGTM. Please have a look at the one remaining inline comment :)
Please also add the migration entry to MIGRATION.md

packages/material-renderers/src/util/layout.tsx
jgteske: The Stack approach sounds definitely like the best practice here.

Both topics are implemented :)

lucas-koehler (approved):
@jgteske The changes look good to me now 🚀 ! Thanks for the contribution and all the updates ❤️ 
I added a small additional note to the MIGRATION.md.

PR #2619, opened 2026-08-28
fix(vue): support elementLabelProp in array controls

kchobantonov:
Fixes the “Array with custom element label” Vue demo example by adding support for elementLabelProp in the Vue Vanilla and Vue Vuetify array controls.
elementLabelProp now takes precedence over the legacy childLabelProp, while preserving backward compatibility.


kchobantonov:
@sdirix please review

lucas-koehler (approved):
Hi @kchobantonov , thanks for the fix ❤️  The changes look good to me 🚀

PR #2625, opened 2026-09-15
fix(react): include props.options in ctxToMultiEnumControlProps's options memo

### LINKED ISSUE #2624: ctxToMultiEnumControlProps omits props.options from its useMemo dependency array, unlike its enum/oneOf-enum siblings: multi enum control never picks up externally-updated options
### Describe the bug

`ctxToMultiEnumControlProps` in `packages/react/src/JsonFormsContext.tsx` (the multi enum variant used internally by `withJsonFormsMultiEnumProps`) memoizes its `options` result with a `useMemo` call that's missing `props.options` from its dependency array. Its two closest siblings, `ctxToEnumControlProps` (plain enum control) and `ctxToOneOfEnumControlProps` (one-of enum control), both correctly include it:

```typescript
// ctxToEnumControlProps — correct
const options = useMemo(
  () => enumProps.options,
  [props.options, enumProps.schema, ctx.i18n?.translate]
);

// ctxToOneOfEnumControlProps — correct
const options = useMemo(
  () => enumProps.options,
  [props.options, enumProps.schema, ctx.i18n?.translate]
);

// ctxToMultiEnumControlProps — missing props.options
const options = useMemo(
  () => enumProps.options,
  [enumProps.schema, ctx.i18n?.translate]
);
```

The two "cell" equivalents (`ctxToEnumCellProps`, `ctxToOneOfEnumCellProps`) also both correctly include their own-props options dependency, so this looks like an isolated omission on just this one function rather than a deliberate difference.

Impact: any consumer that passes its own `options` as an own prop into a multi enum control (most notably a custom renderer implementing dynamically-fetched options, composed outside `withJsonFormsMultiEnumProps` so it can inject `options` as a prop after an async fetch resolves) never sees that value take effect. The memo computed on the very first render (before the async options arrive) is permanently stuck, because none of the other dependencies (`enumProps.schema`, `ctx.i18n?.translate`) change when the options later arrive. The exact same pattern for a single-value enum/one-of-enum control works correctly, since `props.options` is a dependency there.

Concretely: any multi-select control whose available options are computed or fetched asynchronously and handed to it via `options` as an own prop will render an empty list forever, until something unrelated happens to also change `schema` (e.g. editing the field definition, which is why this can look intermittent).

Suggested fix:

```typescript
 export const ctxToMultiEnumControlProps = (
   ctx: JsonFormsStateContext,
   props: OwnPropsOfControl
 ) => {
   const enumProps = mapStateToMultiEnumControlProps(
     { jsonforms: { ...ctx } },
     props
   );
   const options = useMemo(
     () => enumProps.options,
-    [enumProps.schema, ctx.i18n?.translate]
+    [props.options, enumProps.schema, ctx.i18n?.translate]
   );
   return { ...enumProps, options };
 };
```

### Expected behavior

A multi enum control should pick up an externally-updated `options` own prop the same way a plain enum control or a one-of-enum control already does, since all three follow the identical pattern and only the multi-enum variant's memo omits the dependency that makes that work.

1. Render a control using `withJsonFormsMultiEnumProps`, wrapped by a custom HOC that supplies `options` as a prop asynchronously (e.g. `setTimeout(() => setOptions([...]), 0)` before passing `options={options}` down).
2. Observe that the wrapped component's `options` prop stays at its initial (empty) value even after the async update, while the identical pattern using `withJsonFormsOneOfEnumProps` correctly reflects the update.

### Screenshots

_No response_

@jsonforms/react@3.9.0-alpha.1

### Package

_No response_

### Additional context

_No response_

johannesss:
`ctxToMultiEnumControlProps` memoized its `options` result with a `useMemo` call missing `props.options` from its dependency array, unlike its two siblings, `ctxToEnumControlProps` and `ctxToOneOfEnumControlProps`, which both correctly include it. `mapStateToMultiEnumControlProps` already reads `ownProps.options` as its own preferred source over deriving options from the schema, so this is a supported way to supply options externally - a HOC composed outside `withJsonFormsMultiEnumProps` that injects options asynchronously (e.g. after a remote fetch resolves) never saw that update take effect, since none of the memo's other dependencies change when the options later arrive.

Also widens the `props` parameter type from `OwnPropsOfControl` to `OwnPropsOfControl & OwnPropsOfEnum`, matching `ctxToOneOfEnumControlProps` - the narrower type was masking the missing dependency, since `props.options` wasn't even a recognized property on it before.

Added a regression test mounting a control through `withJsonFormsMultiEnumProps`, wrapped in a small stateful provider supplying `options` as an own prop, asserting the rendered control picks up an update to that prop after mount.

Fixes eclipsesource/jsonforms#2624


sdirix:
@EclipseSourceAI

packages/react/test/JsonFormsContext.test.tsx
```diff
@@ -148,6 +150,74 @@ test('withJsonFormsEnumProps - enum: should supply control and enum props', () =
...
+      uischema={uischema}
+      renderers={renderers}
+    />
+  );
+
+  expect(wrapper.find(MockMultiEnumControlUnwrapped).props().options).toEqual([
+    { value: 'red', label: 'Red' },
+  ]);
+
+  setExternalOptions([
```
EclipseSourceAI: This `setState` runs outside `act()`, so React 17 logs an act warning and the assertion relies on legacy-mode synchronous rendering. Wrap it in `act()` from `react-dom/test-utils` (which also makes `wrapper.update()` unnecessary).

packages/react/test/JsonFormsContext.test.tsx
```diff
@@ -148,6 +150,74 @@ test('withJsonFormsEnumProps - enum: should supply control and enum props', () =
...
+    MockMultiEnumControlUnwrapped
+  );
+
+  const ExternalOptionsProvider = (ownProps: any) => {
+    const [options, setOptions] = React.useState<OwnPropsOfEnum['options']>([
+      { value: 'red', label: 'Red' },
+    ]);
+    setExternalOptions = setOptions;
+    return <MockMultiEnumControl {...ownProps} options={options} />;
+  };
```
EclipseSourceAI, line 175: Assigning `setExternalOptions` from the render body is a render side effect. Capturing it in a `useEffect` keeps the test from depending on when React decides to re-run the render.

packages/react/test/JsonFormsContext.test.tsx
```diff
@@ -148,6 +150,74 @@ test('withJsonFormsEnumProps - enum: should supply control and enum props', () =
...
+
+  const schema = {
+    type: 'object',
+    properties: {
+      colors: {
+        type: 'array',
+        items: { type: 'string' },
```
EclipseSourceAI: `items: { type: 'string' }` has neither `enum` nor `oneOf`, so both schema-derived fallbacks in `mapStateToMultiEnumControlProps` are `undefined` here ([master](https://github.com/eclipsesource/jsonforms/blob/41b7991e7753b2c4d851cbbf73bb7408cf5486da/packages/core/src/mappers/renderer.ts#L768-L785)). Giving `items` a `oneOf` would also pin down that `ownProps.options` keeps precedence over the derived options.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

## What this PR does

Adds `props.options` to the `useMemo` dependency array in `ctxToMultiEnumControlProps` and widens its `props` parameter from `OwnPropsOfControl` to `OwnPropsOfControl & OwnPropsOfEnum`, plus a regression test.

## Assessment

The fix looks correct and the scope is right. `ctxToMultiEnumControlProps` was the only enum mapper in `JsonFormsContext.tsx` missing `props.options` from its memo deps (`ctxToEnumControlProps`, `ctxToOneOfEnumControlProps`, `ctxToEnumCellProps` and `ctxToOneOfEnumCellProps` all already have it), and `git log -L` confirms the omission dates back to the original "Update React enum options on i18n changes" commit rather than being intentional. Since `OwnPropsOfEnum.options` is optional, widening the parameter type is not breaking. The Vue binding uses a `computed` over props instead of a manual memo, so there is no equivalent gap there, and Angular does not use the multi-enum mapper at all, so keeping this React-only is right.

I traced the regression test against master: `enumProps.schema` and `ctx.i18n?.translate` both stay referentially stable across the re-render, so the memo does return the stale array and the new test genuinely fails without the fix.

## Points for a maintainer

- All three inline comments are on the new test, not the fix: the state update happens outside `act()`, the setter is captured from the render body, and the test schema has no `enum`/`oneOf` under `items` so it does not cover the `ownProps.options` precedence path.
- Behaviour worth being aware of: `props.options` is compared by reference, so consumers passing an inline array literal to a multi-enum renderer now get a fresh `options` identity on every render. That is the same tradeoff the sibling mappers already make, so it is consistent, but it does change re-render behaviour for those consumers.
- No UI verification was done. The only consumer of `withJsonFormsMultiEnumProps` is `MaterialEnumArrayRenderer`, and no example supplies `options` as an own prop, so the fixed path is not reachable from the examples app or the deploy preview. Tests could not be executed either (no `node_modules` in this environment), so the test assessment above is static analysis only.

sdirix:
@johannesss Thanks for the contribution :heart:. The fix looks good! Please have a look at the AI comments regarding the tests.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review after commit 0e5ab7a.

All three earlier comments on the regression test are properly addressed:
- The `setExternalOptions` call is now wrapped in `act()`.
- The setter is captured via `useEffect` instead of during render.
- The `items` schema now has a `oneOf`, so the test actually proves `ownProps.options` takes precedence over the schema-derived fallback instead of both being empty.

The `@types/react-dom` dependency added to `packages/react/package.json` matches the version already used in `packages/examples-react` and `packages/material-renderers`, and the `pnpm-lock.yaml` diff is just that addition plus an unrelated transitive `follow-redirects` bump from the lockfile resolution, nothing to flag.

No new issues found in this commit. Resolved the three threads from my earlier review.

lucas-koehler (approved):
LGTM, thanks for the contribution ❤️

PR #2626, opened 2026-09-15
fix(material-renderers): import from @mui/material and @mui/icons-material per-component paths

johannesss:
Barrel imports from `@mui/material`/`@mui/icons-material` pull in every export those packages have (both re-export a very large number of components/icons from a single entry module), which defeats tree-shaking for bundlers that don't aggressively analyze through the barrel. MUI's own docs recommend importing each component/icon from its own path instead (e.g. `@mui/material/Button` rather than `{ Button } from '@mui/material'`) for exactly this reason.

`styled`/`useTheme`/`useThemeProps`/`Theme` come from `@mui/material/styles` rather than a same-named subpath, and a few prop-type-only imports (`InputProps`, `InputBaseProps`, `TextFieldProps`, etc.) come from their respective component's own module alongside the default export where one is also imported in the same file.

Verified via `npm run build` (type-checks and bundles clean) and `npm test` (465 tests passing) in `packages/material-renderers`.


sdirix:
@EclipseSourceAI

packages/material-renderers/src/additional/ListWithDetailMasterItem.tsx
```diff
@@ -24,16 +24,14 @@
...
-  Avatar,
-  IconButton,
-  ListItemButton,
-  ListItemAvatar,
-  ListItemSecondaryAction,
-  ListItemText,
-  Tooltip,
-} from '@mui/material';
-import { Delete } from '@mui/icons-material';
+import Avatar from '@mui/material/Avatar';
```
EclipseSourceAI, line 27: `rollup.config.js` externalizes `@mui/material` as a plain string, and rollup matches array-of-string externals exactly, so none of the ~60 new `@mui/material/X` / `@mui/icons-material/X` ids are covered by it ([external list](https://github.com/eclipsesource/jsonforms/blob/41b7991e7753b2c4d851cbbf73bb7408cf5486da/packages/material-renderers/rollup.config.js#L9-L16), where `/^lodash\/.*/` and `/^dayjs\/.*/` are exactly the pattern needed here, and `/^@material-ui\/.*/` is a leftover from v4). It only holds together because the lib config has no node-resolve plugin and rollup falls back to treating unresolved ids as external, so the build log should now be full of UNRESOLVED_IMPORT warnings. Add `/^@mui\/.*/` to `external`.

packages/material-renderers/src/controls/MaterialDateControl.tsx
```diff
@@ -33,7 +33,7 @@ import {
   rankWith,
 } from '@jsonforms/core';
 import { withJsonFormsControlProps } from '@jsonforms/react';
-import { FormHelperText } from '@mui/material';
+import FormHelperText from '@mui/material/FormHelperText';
 import { DatePicker, LocalizationProvider } from '@mui/x-date-pickers';
```
EclipseSourceAI: `@mui/x-date-pickers` is the same kind of barrel (every desktop/mobile/static picker variant plus fields and calendars) and is left untouched here and in `MaterialDateTimeControl`/`MaterialTimeControl`. `@mui/x-date-pickers/DatePicker` and `/LocalizationProvider` are valid subpaths, so the stated goal only lands halfway for the picker controls.

packages/material-renderers/src/util/theme.ts
```diff
@@ -22,15 +22,12 @@
...
-import {
-  Theme,
-  FilledInput,
-  Input,
-  OutlinedInput,
-  TextFieldProps,
-  useThemeProps,
-  InputBaseProps,
-} from '@mui/material';
+import { Theme, useThemeProps } from '@mui/material/styles';
```
EclipseSourceAI, line 25: Adjacent to this PR, but `@jsonforms/material-renderers` declares no `"sideEffects": false` ([package.json](https://github.com/eclipsesource/jsonforms/blob/41b7991e7753b2c4d851cbbf73bb7408cf5486da/packages/material-renderers/package.json#L42-L44), and no other package in the repo has it either), so consumer bundlers keep the whole renderer bundle alive no matter how the MUI imports are shaped. There are no CSS or side-effect imports under `src`, so it looks safe to add and would likely buy more than the import shape does.

packages/material-renderers/src/mui-controls/MuiSelect.tsx
```diff
@@ -25,7 +25,8 @@
 import React, { useMemo } from 'react';
 import { EnumCellProps, WithClassname } from '@jsonforms/core';
 
-import { MenuItem, Select } from '@mui/material';
+import MenuItem from '@mui/material/MenuItem';
```
EclipseSourceAI, line 28: Nothing stops this from regressing on the next PR. A `no-restricted-imports` pattern for `@mui/material` and `@mui/icons-material` in the package eslintrc ([rules block](https://github.com/eclipsesource/jsonforms/blob/41b7991e7753b2c4d851cbbf73bb7408cf5486da/packages/material-renderers/.eslintrc.js#L21-L34)) would keep it enforced. `example/index.tsx` ([link](https://github.com/eclipsesource/jsonforms/blob/41b7991e7753b2c4d851cbbf73bb7408cf5486da/packages/material-renderers/example/index.tsx#L27-L38)) and the tests are linted too and still use the barrel, so they would need an override or the same treatment.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

## What this PR does

Mechanical refactor of `packages/material-renderers/src`: every barrel import from `@mui/material` and `@mui/icons-material` becomes a per-component deep import (`import Button from '@mui/material/Button'`). `styled`/`useTheme`/`useThemeProps`/`Theme` move to `@mui/material/styles`, and the prop types (`InputProps`, `InputBaseProps`, `TextFieldProps`, `InputBaseComponentProps`, `FilterOptionsState`, `AutocompleteRenderOptionState`) move to their owning component module. 34 files, imports only.

## Review

I diffed the imported identifier set per file against `origin/master` and it matches exactly everywhere, so no binding was dropped or renamed. No drive-by changes, no added comments, no LLM-ish filler. `src` is fully converted, only tests and `example/index.tsx` still use the barrel. The `@mui/material/useAutocomplete` subpath for `FilterOptionsState` is the one unusual import, and it is correct since `@mui/material/Autocomplete` does not re-export that type.

## What to focus on

1. **`rollup.config.js` `external`** is the one thing a maintainer should actually check. It lists `@mui/material` as a plain string, and rollup matches string externals exactly, so the ~60 new subpath ids are not covered. It should still build (no node-resolve plugin in the lib config, so rollup externalizes unresolved ids), but the build log is likely full of `UNRESOLVED_IMPORT` warnings and the externalization is now accidental rather than declared.
2. **Scope**: `@mui/x-date-pickers` barrel imports in the three picker controls are untouched, and the package still has no `"sideEffects": false`. Both limit how much this actually buys.
3. **Payoff vs churn** is a maintainer call. Since the MUI v9 upgrade in #2614 the ESM barrel is already tree-shakeable for modern bundlers, so the real wins here are the CJS output and dev-server cold start rather than final ESM bundle size.

I could not verify this in the UI. No dependencies are installed in this worktree, no build is configured, and the sandboxed browser has no network access, so the Netlify preview was unreachable. Given the refactor is import-only and the identifier sets are unchanged, a quick manual pass over the array/table toolbars and the picker controls (the files with the most icon imports) should be enough.

packages/material-renderers/src/util/theme.ts
sdirix: Good point but can be done in a different PR

packages/material-renderers/src/controls/MaterialDateControl.tsx
sdirix: @johannesss Is there a specific reason you skipped applying the same change for the x-date-pickers?

packages/material-renderers/src/mui-controls/MuiSelect.tsx
sdirix: I agree, we should add this as an ESlint rule, MUI also recommends this: https://mui.com/material-ui/guides/minimizing-bundle-size/

packages/material-renderers/src/additional/ListWithDetailMasterItem.tsx
sdirix: Makes sense

sdirix:
@johannesss Thanks for the contribution :heart: Please have a look at the comments

packages/material-renderers/src/controls/MaterialDateControl.tsx
johannesss: @sdirix no, just a miss from me. fixing it now

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up on commit 59afea32, which addresses all three open points from my first review:

- rollup `external` now uses `/^@mui\/.*/` instead of the stale exact-string entries, so the new per-component ids are actually externalized instead of falling out by accident.
- `MaterialDateControl`/`MaterialDateTimeControl`/`MaterialTimeControl` now import `DatePicker`/`DateTimePicker`/`TimePicker`/`LocalizationProvider` from their own `@mui/x-date-pickers/*` subpaths, matching the rest of the package.
- A `no-restricted-imports` override for `@mui/material`, `@mui/icons-material`, and `@mui/x-date-pickers` was added, scoped to `src/**`, which keeps the barrel out of future changes without touching `example/`/`test/`.

Resolved those three threads. Didn't find anything new to flag in the diff. The `sideEffects: false` point stays out of scope per the earlier "different PR" call. Couldn't install deps in this worktree, so no UI check this round either, same limitation as before.

